PR #6, opened 2015-05-31
Add big O notation to provide info about the maximum order

lbenet:
Following a discussion off-line, I have included an big-O notation to display information
on the maximum order for `Taylor1` and `TaylorN` objects.

### DIFF
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -51,7 +51,8 @@ include("utils_TaylorN.jl")
 function pretty_print{T<:Number}(a::Taylor1{T})
     z = zero(T)
     space = utf8(" ")
-    a == zero(a) && return string(space, z)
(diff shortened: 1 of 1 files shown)


lbenet:
@dpsanders I haven't changed the display in the docs, which this PR will induce. Shall I include those changes here, or do you do it, since you authored #5?

dpsanders:
I can do it if you would like, but it will take a while for me to get round to it. If you have time then go ahead and do it.

For this kind of situation, it would be nice to use something like [Dexy](http://www.dexy.it/) to automatically include the output from the given code, rather than redoing it by hand.

lbenet:
I just pushed another version, and forget again to use `[ci skip]`...

I can take care on the manual update of the docs, and will also push the new gh-pages; I will also check Dexy.

dpsanders:
Merging this for now; the documentation changes can go in a separate PR.

PR #7, opened 2015-06-03
RFC: A lighter interface for variables

dpsanders:
Added possibility to change the names used for displaying variables (default is currently x, y etc.; need to add an easy way to use x_1 etc.). Also add a variables() function so can do  x,y=variables() -- currently not exported


src/hashtables.jl
```diff
@@ -26,11 +27,13 @@ end
...
     Maximum order       = $(_params_taylorN.maxOrder)
-    Number of variables = $(_params_taylorN.numVars)""" )
+    Number of variables = $(_params_taylorN.numVars)
+    Variable names = $(_params_taylorN.variable_names)
+    """)
     nothing
 end
 
-global const _params_taylorN = ParamsTaylorN(6,2)
+global const _params_taylorN = ParamsTaylorN(6,2, ["x", "y", "z", "t"])
```
lbenet: `t` is used in the display of `Taylor1` series, so the last entry should be something different, "w", for instance.

Two more related question: How can one change those names? And, isn't it odd that we  we have two independent variables, but we are defining four?

src/hashtables.jl
```diff
@@ -123,6 +126,8 @@ set_maxOrder(n::Int) = set_params_TaylorN(n, _params_taylorN.numVars)
 get_numVars() = _params_taylorN.numVars
 set_numVars(n::Int) = set_params_TaylorN(_params_taylorN.maxOrder, n)
 
+set_variable_names(names::Vector{UTF8String}) = _params_taylorN.variable_names = names
+
 function set_params_TaylorN(order::Int, numVars::Int)
```
lbenet: Below (line 138) needs changes, since `ParamsTaylorN` now has three field names; I think tests are not passing because of this line. A way around is to simply add another method for `ParamsTaylorN`

dpsanders: Yes I see, thanks.

src/hashtables.jl
dpsanders: There is an unexported `set_variables` method to change the names. This all needs a lot more work but I think the idea is good. 

I see what you mean about the number of variables. This was really just trying out some ideas and I shouldn't really have made it a PR yet. But I think I have a good idea about what to do, along the lines of `SymPy`:

x, y = variables("x y")

should fix both the names *and* the number of variables. (I.e. it should automatically change the current number of variables.) Does this sound good to you?

lbenet: This indeed sounds really nice. 

How will you handle the maximum degree and, perhaps more important, the fact that we are constrained by a given number of variables? My concern is related to use it as
```julia
x, y, z = variables("x y z")
```
when `get_numvars()` returns 2. Maybe a warning or an error...

dpsanders:
There is now an exported function `set_variables`, so you can do

`x, y, z = set_variables("x y z")`

or

`x, y, z, w = set_variables("x", 7, numvars=4)`

(probably `order` should be a keyword here)

Note this changes the number of variables and the order. My idea is that this should be the public interface to change these things and `set_numVars` etc. should not be exported. How does that sound?

src/hashtables.jl
dpsanders: See comment below: My idea is that this should now be the public interface.

dpsanders:
Creo que deberíamos suprimir los warnings
`Warning: redefining constant _params_taylorN`

lbenet:
I think we are discussing two separated issues here: renaming `_params_taylorN`, `set_numVars`, etc, and the implementation of a lighter interface to change the display of the distinct `TaylorN` variables.

For the first one, I do agree with you that those names are not attractive (my speciality :-) ), so please go on and rename them. Among the options above (for `set_numVars` and `set_maxOrder` I prefer `set_numvars` and `set_maxorder`). Something similar has to be implemented for `get_numVars` and `get_maxOrder`. 

For the latter one, I really like the idea and the approach. It's indeed a nice feature that changing the names adapts accordingly the rest of the parameters. But I think we should be able to change the parameters (`maxOrder` and 'numVars`) separately, without having to change the variable names. This forces us to have a reasonable default for naming the variables, which is set by the number of variables. In this respect, I think a simplel default is the one that looks like `xᵢ`, with i=1,..., numVars, where the function `subscriptify` can be useful. This default is easy to implement, though I do like to have different letters, such as `x`, `y`, `z`, `w`, etc. (We must avoid `t`, because that's the letter used for the output of `Taylor1` objects.)

lbenet:
The situation I have in mind --to avoid specifying things "one-by-one"-- is the test that checks an identity proved by Euler. There are 8 variables, which we simply want to distinguish without explicitly naming them. Currently, we have:
```
julia> set_params_TaylorN(4,8)
(4,8)
julia> show_params_TaylorN()
INFO: `TaylorN` and `HomogeneousPolynomial` parameters:
Maximum order       = 4
Number of variables = 8
Variable names = UTF8String["x","y","z","t"]
```
so there are 4 variables which seem to remain unamed, and leads to a BoundsError because variable 5 is not named.

One possibility it just occurred to me is through `taylorN_variable`, where a keyword argument (UTF8String) has as default something like `xᵢ` with i the number of the variable set in `taylorN_variable`.

lbenet:
Note that the tests currently pass, but if you run the code in the REPL it yields the mentioned Error.

dpsanders:
Actually this is already possible in this branch:

```julia
julia> a1, a2, a3, a4, b1, b2, b3, b4 = set_variables("x", 4, numvars=8)
8-element Array{TaylorSeries.TaylorN{Float64},1}:
  1.0x₁ + 𝒪(‖x‖⁵)
  1.0x₂ + 𝒪(‖x‖⁵)
  1.0x₃ + 𝒪(‖x‖⁵)
  1.0x₄ + 𝒪(‖x‖⁵)
  1.0x₅ + 𝒪(‖x‖⁵)
  1.0x₆ + 𝒪(‖x‖⁵)
  1.0x₇ + 𝒪(‖x‖⁵)
  1.0x₈ + 𝒪(‖x‖⁵)

julia> a1
 1.0x₁ + 𝒪(‖x‖⁵)

julia> b1
 1.0x₅ + 𝒪(‖x‖⁵)
```

dpsanders:
Of course now you can make it nicer:
```
subscriptify = TaylorSeries.subscriptify

variable_names = [string("a", subscriptify(i)) for i in 1:4]
append!(variable_names, [string("b", subscriptify(i)) for i in 1:4])

a1, a2, a3, a4, b1, b2, b3, b4 = set_variables(variable_names, 4)

julia> expr_lhs1 = a1^2 + a2^2 + a3^2 + a4^2
 1.0a₁² + 1.0a₂² + 1.0a₃² + 1.0a₄² + 𝒪(‖x‖⁵)

julia>     expr_lhs2 = b1^2 + b2^2 + b3^2 + b4^2
 1.0b₁² + 1.0b₂² + 1.0b₃² + 1.0b₄² + 𝒪(‖x‖⁵)

julia>     lhs = expr_lhs1 * expr_lhs2
 1.0a₁²b₁² + 1.0a₂²b₁² + 1.0a₃²b₁² + 1.0a₄²b₁² + 1.0a₁²b₂² + 1.0a₂²b₂² + 1.0a₃²b₂² + 1.0a₄²b₂² + 1.0a₁²b₃² + 1.0a₂²b₃² + 1.0a₃²b₃² + 1.0a₄²b₃² + 1.0a₁²b₄² + 1.0a₂²b₄² + 1.0a₃²b₄² + 1.0a₄²b₄² + 𝒪(‖x‖⁵)
```

dpsanders:
I think I prefer that the user must explicitly specify the variable names they want and the variables they will use for the resulting objects. Don't you like that? I would even go so far as to initialize with 0 variables or something like that, so they are forced to do so. 

As far as `t` goes, I'm not sure why we shouldn't use it if it's clear that it's a multivariable expansion.

lbenet:
Sorry, I think I misunderstood the way that `variables("x", 4, numvars=8)` was working; my fault. And, it is indeed nice, specially the way it simplifies the tests (and documentation!).

Regarding the error I spotted earlier, here I reproduce it
```julia
julia> using TaylorSeries

julia> set_params_TaylorN(4,8)
Warning: redefining constant _params_taylorN
(4,8)

julia> show_params_TaylorN()
INFO: `TaylorN` and `HomogeneousPolynomial` parameters:
Maximum order       = 4
Number of variables = 8
Variable names = UTF8String["x","y","z","t"]

julia> for i=1:4
            ai = symbol(string("a",i))
            bi = symbol(string("b",i))
            @eval ($ai) = taylorN_variable(Int,$i,4)
            @eval ($bi) = taylorN_variable(Int,4+($i),4)
        end

julia> a2
Error showing value of type TaylorSeries.TaylorN{Int64}:
ERROR: BoundsError: attempt to access 4-element Array{UTF8String,1}:
 "x"
 "y"
 "z"
 "t"
  at index [5]
 in homogPol2str at /Users/benet/Fisica/6-IntervalArithmetics/TaylorSeries.jl/src/TaylorSeries.jl:106
 in pretty_print at /Users/benet/Fisica/6-IntervalArithmetics/TaylorSeries.jl/src/TaylorSeries.jl:89
 in show at /Users/benet/Fisica/6-IntervalArithmetics/TaylorSeries.jl/src/TaylorSeries.jl:200
 in anonymous at show.jl:1254
 in with_output_limit at ./show.jl:1231
 in showlimited at show.jl:1253
 in writemime at replutil.jl:4
 in display at REPL.jl:113
 in display at REPL.jl:116
 in display at multimedia.jl:151
 in print_response at REPL.jl:133
 in print_response at REPL.jl:120
 in anonymous at REPL.jl:599
 in run_interface at ./LineEdit.jl:1554
 in run_frontend at ./REPL.jl:842
 in run_repl at ./REPL.jl:165
 in _start at ./client.jl:454
```
Certainly, this can be changed to the construction you described, which is quite nice. Yet, we must make sure that changing the parameters through `set_params_TaylorN(4,8)` by default implies `variables("x", 4, numvars=8)`. Also, to make sure that the length of `variable_names` in `ParamsTaylorN` is the `numvars` would be nice.

A final comment: maybe I'm too used to *only* change `numvars` and `maxorder` and get the rest done automatically, and probably too lazy as well. But I get your point, and we indeed may push the user to set this by hand.

lbenet:
So, I guess we can decide in favor of `variables("x", 4, numvars=8)` to define the variables names, `maxorder` and actual number of variables, instead of the current one. But we should delete, or at least not export, the current usage. I also think it is important to be sure that the length on `names` is `numvars` once this is set in `ParamsTaylorN`.

dpsanders:
Yes, I think we should just not export `set_numvars` and `set_maxorder`. In this way the user has to specify exactly which variables to use or the names and number of the variables in the case that they are indexed, and this ensures that `numvars` is correct.

lbenet:
Agreed. But also `set_params_TaylorN` should not be exported, if it is of any value to keep. We should simply avoid that the user falls into a situation as I did (above).

Would you also include this in the documentation and tests?

dpsanders:
I have incorporated `set_variables` into the tests and documentation and removed the deprecated exports.

lbenet:
I'm catching up, but my local branch and the remote one have diverged? Did you rebased?

dpsanders:
Yes, I rebased to have the latest version  of the docs.

lbenet:
Ok. I'll remove my local branch, pull the remote, and review it.

lbenet:
I think it is a good idea that there is a default `variable_names` in `ParamsTaylorN`, which is compatible with `numVars`. Right now, `variable_names` is a zero length UTF8String vector. That way, we could start right from the beginning avoiding `x, y = set_variables("x", numvars=2, order=6)`.

Additionally, I think it is a good idea to define some inner constructor in `ParamsTaylorN`, to avoid that `numVars` and the length of `variable_names` are different.

src/hashtables.jl
```diff
@@ -19,18 +19,30 @@
 type ParamsTaylorN
     maxOrder :: Int
     numVars  :: Int
+    variable_names :: Array{UTF8String,1}
```
lbenet: I think a reasonable default should be fixed, e.g., as in line 171 below. Something like
```julia
     variable_names :: UTF8String[string("x", subscriptify(i)) for i in 1:numVars]
```

Incidentally, I recall that you proposed to change maxOrder -> maxorder, and numVars -numvars. Am I right?

src/TaylorSeries.jl
```diff
@@ -169,7 +171,8 @@ function numbr2str{T<:Real}(zz::Complex{T}, ifirst::Bool=false)
     return cadena
 end
 
-name_taylorNvar(n::Int) = string("⋅x", subscriptify(n))
+#name_taylorNvar(n::Int) = string("⋅x", subscriptify(n))
+name_taylorNvar(i::Int) = string(" ", _params_taylorN.variable_names[i])
```
lbenet: Since "⋅" (\cdot) is removed from the display of `TaylorN` variables, shouldn't we be consistent and do the same for `Taylor1` variables? (cf line 62)

src/hashtables.jl
```diff
@@ -116,28 +129,69 @@ const indicesTable, sizeTable, posTable = generateTables()
...
+
+    set_variables(variable_names, order=order)
+
+end
+
+
+
+
+
+# function set_params_TaylorN(order::Int, numVars::Int)
```
lbenet: We should erase all this unneeded things, instead of leaving them there commented.

lbenet:
I noticed that you rebased things included in different PR (#9 and #10). While I understand quite well the motivation to integrate things related to the documentation, #10 belongs to a quite different thing and should have been kept separated.

dpsanders:
Yes you're right, sorry. I've just updated the user guide notebook on the user_guide branch. In this branch, I updated the user guide with `set_variables` in the markdown. I'll delete the notebook from this branch.

dpsanders:
By the way, I am unable to use `mkdocs serve`: it gives the message

`IOError: [Errno 2] No such file or directory: 'docs/Home'`

lbenet:
I don't know what's going on with mkdocs. I just tried it in this branch and everything is fine. What version do you have: I get version 0.13.2

dpsanders:
I'm not sure how useful it is to have defaults for parameter names -- how do you know what the user will want by default?  I have changed it to

`global const _params_taylorN = ParamsTaylorN(6, 2, UTF8String["x₁", "x₂"])`

but I actually think that it would be better to force the user to specify what they wanted.

Indeed I had a too-old version of `mkdocs`, working fine now, thanks.

dpsanders:
As regards the display, I have tried with `\cdot`, with space, and without space. I'm not sure I particularly have a preference, though I quite liked without any space (but this is a problem if you want to have variables called `xx` and `yy`, for example, which is why I put the space back in). Do you have a preference? The dots seemed to add too much noise. Ideally this could also be user-configurable?

dpsanders:
If the user does not use the constructor of `ParamsTaylorN`, I don't see a particular need to make it more complicated with an inner constructor.

dpsanders:
I think we should leave the names `numVars` etc. for the moment and do a complete refactor later (after 0.1)

lbenet:
My point is simply to make it consistent (the length of `variable_names` does not correspond with `numVars`), and if the defaults are reasonable with respect to `numVars` and `maxOrder`, you may get to work without using `set_variables`

lbenet:
Tests pass. I can merge this PR, and then we get into the small details. Yet, I'm not quite sure whether we should merge first #9 and #10

dpsanders:
I think #10 should be merged first (maybe it needs a rebase first)? Then we can rebase this branch against that. #9 is really independent and can be done afterwards.

lbenet:
Ok. You take care of #10 and I about this PR.

lbenet:
Shall I squash this and then merge it? (I'm not too experienced doing this, so if you can I prefer you do it.) I know how to do it the dirty way, though...

dpsanders:
What's the dirty way? Let me try to do it.

dpsanders:
Hmmm, it looks a bit complicated so I won't be able to do it right now, and probably not today. Do you want to have a go? (Feel free to use the dirty way...)

lbenet:
You update the branch (locally in your computer), try to merge the branch to master, and then solve all conflicts. Eventually, you'll have everything fine, and simply push it into master.

lbenet:
I guess rebasing might help...

lbenet:
Ok, I'll take care of this...

lbenet:
@dpsanders Done!

PR #9, opened 2015-06-03
RFC: Convert user guide to a notebook

dpsanders:


dpsanders:
I converted `user_guide.md` [renamed in this branch] to the notebook `User guide.ipynb` using
[notedown](https://github.com/aaren/notedown) with

`notedown user_guide.md --to notebook > user_guide.ipynb`

There are some missing equation links etc. -- I don't think these are possible yet with the notebook.

dpsanders:
Oops, I see that all the `julia>` prompts are still there -- more work to do...

lbenet:
If it helps, I did put the code on a separate file.

Incidentally, I tried to use [Dexy](www.dexy.it/) as you suggested, but didn't quite succeed in getting
a nice output.

lbenet:
Wouldn't this "collide" with the mkdocs structure? 

What about having both, the notebook in the examples dir, and the user_guide file in the docs dir? This would allow to have a consistent documentation, and the notebook for the user to play with.

lbenet:
@dpsanders I'm not quite sure what to do about this. 

My understanding is that you renamed one file (ok), and added the same as an ipynb. I think the latter should be moved to the "examples" dir, and be, from time to time, be updated.

Incidentally, did you change also the mkdocs.yml file so it points to the new file?

I just want to know what is needed for this to be merged.

dpsanders:
Yes, you are probably right that the notebook should go in the `examples` directory.
However, perhaps the static markdown version should be produced from the notebook, and not the other way around, since the notebook is much easier to update when the code changes -- e.g. the addition of the `O(x^6)` stuff would just have needed to rerun the notebook and then produce the markdown from that (e.g. with `notedown`).

lbenet:
It's easier to go from the notebook to markdown. Do you know if the notebook accepts the tricks to make MathJax work? I know you can display equations, but can you number them? I think it looks nice, though we may get around this.

dpsanders:
There is now a good version of the user guide as a notebook as `User guide.ipynb`, still in the `docs` directory. Unfortunately, I have not been able to export it to a nice markdown version. There is a previous version of the markdown (that I started to correct) still at `usersguide.md`. However, it is *really* painful to do the corrections by hand when the output changes, and much easier just to do it in the notebook.

I don't see any problem in just making a static version (i.e. a fixed HTML version of the notebook) and including it instead of `usersguide.md`. What do you think @lbenet ?

dpsanders:
I updated the user guide notebook and moved it to examples.

lbenet:
I'll take care about the manual translation of this into the documentation.

lbenet:
While working on the manual translation for the documentation, I noticed the problem that #13 solves. After that PR is merged, I'll rebase and do the manual translation fr the documentation.

lbenet:
I'm merging and I'll update the documentation as well.

PR #10, opened 2015-06-03
WIP: improve evalTaylor for TaylorN

lbenet:
`evalTaylor` (as it is now) yields *not so accurate* results due to accumulation of
round-off errors when used with `TaylorN`. Below, I implement a different
algorithm, which essentially consists on applying Horner's rule for each variable.

So far, this has only be tested in 2-variable expansions; currently it returns
a `TaylorN` object

Functions are not exported

### DIFF
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -36,7 +36,7 @@ import Base: zero, one, zeros, ones,
 ## Exported types and methods
 export Taylor1, TaylorN, HomogeneousPolynomial
 export taylor1_variable, taylorN_variable, get_coeff,
-    diffTaylor, integTaylor, evalTaylor, deriv,
+    diffTaylor, integTaylor, evaluate, deriv,
     set_params_TaylorN, show_params_TaylorN,
     set_maxOrder, get_maxOrder, set_numVars, get_numVars,
     ∇, jacobian, hessian
--- a/src/utils_Taylor1.jl
+++ b/src/utils_Taylor1.jl
@@ -521,7 +521,7 @@ end
 integTaylor{T<:Number}(a::Taylor1{T}) = integTaylor(a, zero(T))
 
 ## Evaluates a Taylor1 polynomial on a given point using Horner's rule ##
-function evalTaylor{T<:Number,S<:Number}(a::Taylor1{T}, dx::S)
+function evaluate{T<:Number,S<:Number}(a::Taylor1{T}, dx::S)
     R = promote_type(T,S)
     @inbounds suma = convert(R, a.coeffs[end])
     @inbounds for k = a.order:-1:1
@@ -529,9 +529,9 @@ function evalTaylor{T<:Number,S<:Number}(a::Taylor1{T}, dx::S)
     end
     suma
 end
-evalTaylor{T<:Number}(a::Taylor1{T}) = evalTaylor(a, zero(T))
+evaluate{T<:Number}(a::Taylor1{T}) = a.coeffs[1]
 
-function evalTaylor{T<:Number,S<:Number}(a::Taylor1{T}, x::Taylor1{S})
+function evaluate{T<:Number,S<:Number}(a::Taylor1{T}, x::Taylor1{S})
     a, x = fixshape(a, x)
     @inbounds suma = a.coeffs[end]
     @inbounds for k = a.order:-1:1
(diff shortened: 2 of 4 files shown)


lbenet:
Something odd was happening with Travis, because tests didn't pass, though they did locally

lbenet:
Incidentally, something that still needs some work in `evaluateNew` is related to use it on a vector of `TaylorN`, for example, for substitutions. The following shows that this works for `evaluate` but not for `evaluateNew`:
```julia
julia> x = taylorN_variable(1)
 1.0⋅x₁ + 𝒪(‖x‖¹⁸)

julia> y = taylorN_variable(2)
 1.0⋅x₂ + 𝒪(‖x‖¹⁸)

julia> evaluate(exy,[x,x])
 1.0 + 2.0⋅x₁ + 2.0⋅x₁² + 1.3333333333333333⋅x₁³ + 0.6666666666666666⋅x₁⁴ + 0.2666666666666667⋅x₁⁵ + 0.08888888888888888⋅x₁⁶ + 0.02539682539682539⋅x₁⁷ + 0.006349206349206349⋅x₁⁸ + 0.0014109347442680773⋅x₁⁹ + 0.0002821869488536155⋅x₁¹⁰ + 5.1306717973384636e-5⋅x₁¹¹ + 8.551119662230772e-6⋅x₁¹² + 1.3155568711124264e-6⋅x₁¹³ + 1.8793669587320378e-7⋅x₁¹⁴ + 2.5058226116427168e-8⋅x₁¹⁵ + 3.1322782645533972e-9⋅x₁¹⁶ + 3.6850332524157597e-10⋅x₁¹⁷ + 𝒪(‖x‖¹⁸)

julia> TaylorSeries.evaluateNew(exy,[x,x])
Error showing value of type TaylorSeries.TaylorN{TaylorSeries.TaylorN{Float64}}:
ERROR: MethodError: `numbr2str` has no method matching numbr2str(::TaylorSeries.TaylorN{Float64}, ::Bool)
```

dpsanders:
I don't actually understand what `evaluate(exy,[x,x])` is *supposed* to do!

lbenet:
`evaluate(exy,[x,x])` is a substitution, so it should yield `exp(2x)` as a result. 

Yet, comparing it with `exp(2*taylor1_variable(17))` there are differences in the coefficients, due to round-off errors, which is what I'd like to minimize.

dpsanders:
I don't see how `evaluate(exy,[x,x])` makes sense mathematically? What is the mathematical operation behind this?

lbenet:
I think a fair candidate is $\lim_{y\to x} f(x,y) = f(x,x)$. The mathematical operation behind that is substitution.

dpsanders:
But exactly we *can't* take this limit: we can think of $x$ and $y$ as being coordinates along orthogonal directions close to the expansion point. An alternative viewpoint is that the Taylor series is just a formal power series. In any case $x$ and $y$ are different objects and it doesn't make sense to replace a $y$ by an $x$ in this context, as far as I can see.

The only thing that makes sense, as far as I can see, is to substitute numerical values for these coefficients (and this does currently work).

lbenet:
I'm not quite sure I get your point in all this since the limit I described above is like checking the behaviour along (one) specific line: $ lim_{y \to x} tan(x+y) = tan(2x) $. In any case, what I can certainly do is allow only the substitution by numerical values, so either `Real` or `Complex`. 

Also, I could rename `evaluateNew`->`evaluate`, so the new implementation becomes the one used, since it is marginally more accurate that the current one. I could keep the current one as `evaluateOld` (or something similar) or simply delete it.

dpsanders:
I think you should just rename to `evaluate` and delete the old version if the new one is better -- git still remembers that it's there in the previous versions (if you ever need it) ;)

Yes, I think just allowing substituting numerical values is a good solution.

lbenet:
I'll do this later and push the PR.

lbenet:
Done! Tests pass.

lbenet:
Can you be more [explicit](https://github.com/JuliaDiff/TaylorSeries.jl/pull/7#issuecomment-113789801) about rebasing thir PR?

dpsanders:
Squashed; waiting for the green light.

lbenet:
Thanks for squashing this. I think you can go on and merge; then I'll merge #7, and get to the documentation

lbenet:
I erased this branch to avoid branches cluttering.

PR #13, opened 2015-06-22
Avoid `evaluate` a TaylorN expression on a TaylorN vector

lbenet:
This PR restricts `evaluate` a TaylorN expression on a TaylorN vector, since it yields wrong results.
I noticed it while working on #9, testing and cleaning up.


lbenet:
Tests pass; I'm merging to get #9 done...

PR #16, opened 2015-06-28
RFC: Refactoring of tables

dpsanders:
- The generation code for tables has been refactored.
- Tuples are now used instead of arrays in `index_table`
- The `pos_table` now uses the tuple directly as the key;
hashing is no longer required. (This is done in any case by the
dictionary)
- Names have been de-camel-cased
  (numVars -> num_vars, indexTable -> index_table)
- The tables are currently not `const`, since their type now changes
when num_vars changes. We will have to think about how to solve this
- `@simd` and `@inbounds` removed for now.  (Sorry, this has caused noise in the diff.)
- EDIT: I also split up and renamed some files 

Refactoring of variable names

Split out definition of parameter type into new file TaylorN_parameters.jl

Split out printing into printing.jl

Minor refactoring

Factored out generation of homogeneous polynomial indices of given degree

Rename files without utils

Complete rewrite of table generation code

Remove parameter stuff from hashtables

Correct new file names in TaylorSeries.jl

Refactoring of printing of homogeneous polynomials

Rename TaylorN_parameters.jl -> parameters.jl

Corrected include in TaylorSeries

Added tuple addition

Refactor tests into 3 files

Allow redefining table sizes again

Almost all tests pass, except when changing the number of variables, which will require a major refactoring

Remove const for tables for now. Remove @simd and @inbounds


dpsanders:
Hmmm, this is 5 times slower than master for the Fateman test. Not sure what's going on... Could it really be the `@inbounds` and `@simd`?

dpsanders:
Thinking more about it, I understand why you chose to have only the hash in the table; I guess this will indeed be faster. 

However, what we definitely need is a more abstract interface for looking up things in the table to hide these details. I should be able to get to this in the next couple of days.

dpsanders:
Thinking more about it, I understand why you chose to have only the hash in the table; I guess this will indeed be faster. 

However, what we definitely need is a more abstract interface for looking up things in the table to hide these details. I should be able to get to this in the next couple of days.

lbenet:
Sorry for the long silence; I've been quite busy in other things... I will try to get into this as soon as I can, but this will be certainly at least tomorrow.

So, before catching up and getting into the details, I'm intrigued about the performance of *only* changing in `indicesTable` (or `index_table`) the arrays by tuples. Do you have any kind of comparison? 

I agree that a new abstraction of the data may be the way to go. Yet, in julia 0.3.x the garbage collector is not as efficient as it is in 0.4. My point is that the enourmous amounts of allocated space do not show up like that in version 0.4.

dpsanders:
Yes sorry, I had the commits more atomically and then did a premature rebase. 
I'll try and redo this PR as a series of PRs. I won't be able to get to this for a couple of weeks though.

lbenet:
Getting back slowly to this, sorry, a very busy week... 

Here there are some comments:

- Why did you remove `@simd` and `@inbounds`? I know that this affects *a lot* the performance! I was quite careful to code checking that there is no way to be out of bounds. So `@inbounds` is well justified and checked. The use of `@simd` is another thing, and I was just using it whenever I was sure that reordering the `for`-loop should not influence on the numerical value obtained. I really think this should be reverted.

- I really like much more your naming conventions :+1: 

- I guess it is good programming style to use functions instead of constants (e.g., `get_order()` instead of `_params_taylorN_.order`). Yet, internally, I think that using constants yields better performance simply because they have a fixed type, and the compiler knows about this.

- At some point I had some thing like the `inverse_hash_map` function. The implementation I finally ended up with yields good performance, since you transverse once a `for` loop, which may a priori be a long thing. Using `map`, as you do it, essentially imposes a `for` loop each time it is used. In a way, my approach is alike to de-vectorizing, which is recommended for good performance.

A final comment. On the plane I rewrote everything related to `Taylor1` but using internally `NTuples` instead of `Array{T,1}`; essentially the same tests are used, except for grammatical changes and some additions, and the tests pass. Probably there are better ways to do it instead of what I implemented, but the performance tests that I made (integrate the Kepler problem in the most straight forward way)  yielded a terrible slowdown, so I am unsure that this is the way to go. I have not done anything with `TaylorN`, which I guess will display even worst comparison...

dpsanders:
Closing this; will redo as a series of small PRs and check performance more carefully.

Issue #14, opened 2015-06-28
Write multiplication by constants explicitly


dpsanders:
There should be explicit functions for multiplying by constants, since this is very efficient and does not require conversion.

lbenet:
I agree with this, and also for addition. Promotion in this case kills performance.

lbenet:
Closing this, since it's merged.

PR #17, opened 2015-07-04
Solves issue #14

lbenet:
This PR solves #14, including all arithmetic operations for `Taylor1`, `HomogeneousPolynomial` and `TaylorN`


lbenet:
@dpsanders Tests pass !

PR #18, opened 2015-07-13
Add Type to set_variables and tests

lbenet:
This adds the possibility to define a `TaylorN` variable using `set_variables` of any
parametric type.


lbenet:
@dpsanders Tests do not pass in the release version, due to some oddity with travis, think. In my machine, tests pass. Could you check in yours?

Incidentally, yesterday I managed to bring back to life the coveralls badge; moving TaylorSeries to JuliaDiff was a bit subtle. This would be (again) the first commit with coverage.

lbenet:
Tests pass, so I'm merging.

dpsanders:
In cases where there seems to be a problem with Travis, you can restart the build with the icon in the top right corner, which is what I did. I like the PR!

lbenet:
Thanks for tip!

PR #19, opened 2015-07-15
Add new tests to increase coverage and some clean-up

lbenet:
Among the things that I have cleaned up, I commented some convert
methods which I think speeded-down any calculation. I also
commented the function `variables` which was never used
(utils_TaylorN.jl).

### DIFF
--- a/src/utils_Taylor1.jl
+++ b/src/utils_Taylor1.jl
@@ -72,8 +72,6 @@ end
 convert{T<:Number, S<:Number}(::Type{Taylor1{T}}, b::Array{S,1}) =
     Taylor1(convert(Array{T,1},b))
 convert{T<:Number}(::Type{Taylor1{T}}, b::Number) = Taylor1([convert(T,b)], 0)
-# convert{T<:Number}(::Type{Taylor1{T}}, a::Taylor1{T}) = a
-# convert{T<:Number,S<:Number}(::Type{Taylor1{T}}, b::Array{S,1}) = Taylor1(convert(T,b))
 convert{T<:Number}(::Type{Taylor1{T}}, b::T) = Taylor1([b], 0)
 
 promote_rule{T<:Number, S<:Number}(::Type{Taylor1{T}}, ::Type{Taylor1{S}}) =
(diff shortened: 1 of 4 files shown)


lbenet:
I think this is ready to be merged.

@dpsanders Do you know how to test `show` methods?

dpsanders:
Here's one possible way to get the output of `show` methods as a string:
```julia
julia> buffer = IOBuffer()
IOBuffer(data=Uint8[...], readable=true, writable=true, seekable=true, append=false, size=0, maxsize=Inf, ptr=1, mark=-1)

julia> show(buffer, 3)

julia> bytestring(buffer)
"3"
```
Now that I think of it, can't you just do something like the following, using `string`?
```julia
julia> using TaylorSeries
x, y =
julia> x, y = set_variables("x y")
2-element Array{TaylorN{Float64},1}:
  1.0 x + 𝒪(‖x‖⁷)
  1.0 y + 𝒪(‖x‖⁷)

julia> x
 1.0 x + 𝒪(‖x‖⁷)

julia> string(x)
" 1.0 x + 𝒪(‖x‖⁷)"
```

dpsanders:
Apparently `string` does in fact use `IOBuffer` internally. I think instead of `bytestring` the function should be `takebuf_string` (a truly awful name for a function!)

src/utils_Taylor1.jl
```diff
@@ -72,8 +72,8 @@ end
 convert{T<:Number, S<:Number}(::Type{Taylor1{T}}, b::Array{S,1}) =
     Taylor1(convert(Array{T,1},b))
 convert{T<:Number}(::Type{Taylor1{T}}, b::Number) = Taylor1([convert(T,b)], 0)
-convert{T<:Number}(::Type{Taylor1{T}}, a::Taylor1{T}) = a
-convert{T<:Number}(::Type{Taylor1{T}}, b::Array{T,1}) = Taylor1(b)
+# convert{T<:Number}(::Type{Taylor1{T}}, a::Taylor1{T}) = a
```
dpsanders: I would suggest just deleting all these methods, instead of commenting them out.

lbenet:
I've followed your first suggestion, using `IOBuffer()` and `bytestring`, which I find is quite honest.

I think there is still something to be done to get a better coverage, since some functions in TaylorSeries.jl and hashtables.jl still have functions that remain untested.

In any case, tests are passing.

lbenet:
My mistake. I'm leaving tomorrow so I wanted to get a couple of things done before...

I indeed agree that your second suggestion is much more readable, now that I understand it. I can submit a new commit to this branch, but this will be during this afternoon.

Incidentally, do you know why coveralls hasn't updated the bagde?

dpsanders:
Don't worry, I already did it in #20 ;)

PR #20, opened 2015-07-15
Replace IOBuffer with string in tests

dpsanders:


lbenet:
I was simply too slow to comment on this https://github.com/JuliaDiff/TaylorSeries.jl/pull/19#issuecomment-121694417
Thanks for doing it!

lbenet:
Merging... I lost the patience with coveralls....

PR #21, opened 2015-07-22
Rename get_maxOrder -> get_order and get_numVars -> get_numvars.

dpsanders:
Also replaces direct accesses of fields in `_params_TaylorN_`  with calls to `get_order` and `get_numvars`. 

Note that these calls are inlined and so do not affect performance (checked with the Fateman benchmarks, which I have separated out into a file `fateman.jl` in the new `perf` directory).

### DIFF
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -40,7 +40,7 @@ export Taylor1, TaylorN, HomogeneousPolynomial
 export taylor1_variable, taylorN_variable, get_coeff,
     diffTaylor, integTaylor, evaluate, deriv,
     show_params_TaylorN,
-    get_maxOrder, get_numVars,
+    get_order, get_numvars,
     set_variables,
     ∇, jacobian, hessian
 
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -115,7 +115,7 @@ function pretty_print{T<:Number}(a::TaylorN{T})
 end
 
 function homogPol2str{T<:Number}(a::HomogeneousPolynomial{T})
-    numVars = _params_TaylorN_.numVars
+    numVars = get_numvars()
     order = a.order
     z = zero(T)
     space = utf8(" ")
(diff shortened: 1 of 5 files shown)


src/TaylorSeries.jl
```diff
@@ -191,13 +191,13 @@ function numbr2str{T<:Real}(zz::Complex{T}, ifirst::Bool=false)
 end
 
 #name_taylorNvar(n::Int) = string("⋅x", subscriptify(n))
-name_taylorNvar(i::Int) = string(" ", _params_taylorN.variable_names[i])
+name_taylorNvar(i::Int) = string(" ", _params_TaylorN_.variable_names[i])
```
lbenet: To be fully consequent with this PR, shouldn't we add the following new function?
```julia
function get_variable_names(i::Int)
    @assert 1 ≦ i ≦ get_numvars()
    _params_TaylorN_.variable_names[i]
end
get_variable_names() = _params_TaylorN_.variable_names
``` 
A reason against this is that it would be used only once, as far as I can see...

src/hashtables.jl
```diff
@@ -11,40 +11,35 @@
...
+
+
 @doc """Display the current parameters for `TaylorN` and
 `HomogeneousPolynomial`""" ->
 function show_params_TaylorN()
     info( """Parameters for `TaylorN` and `HomogeneousPolynomial`:
-    Maximum order       = $(_params_taylorN.maxOrder)
-    Number of variables = $(_params_taylorN.numVars)
-    Variable names      = $(_params_taylorN.variable_names)
+    Maximum order       = $(_params_TaylorN_.order)
```
lbenet: Is it reasonable to use here `get_order()`, `get_numvars()` and `get_variable_names()`?

src/TaylorSeries.jl
lbenet: The second method is already defined in hashtable.jl. I just noticed...

lbenet:
Apart from the two minor things described above, I think this is ready to be merged.
I'll wait for your reply.

lbenet:
I'm merging this, to continue development which incorporates these and other changes.

PR #22, opened 2015-07-22
Refactoring of table names, and splitting out files

dpsanders:
Follows on from #21.

After the experience with #16 , I have left all the commits as atomic as possible.
I can rebase and squash later if desired.

PR #23, opened 2015-07-22
Refactor hash table generation.

dpsanders:
This is a drop-in replacement for the previous version except that the exact order
in which each set of indices is generated is changed. 

~~The performance is basically the same as in the previous version.~~
EDIT: The performance is now significantly better: 8.4s instead of 12s for the first Fateman test on my machine!

Follows on from #22

### DIFF
--- a/perf/fateman.jl
+++ b/perf/fateman.jl
@@ -1,6 +1,6 @@
 using TaylorSeries
 
-set_variables("x", numvars=4, order=40)
+x, y, z, w = set_variables(Int128, "x", numvars=4, order=40)
 
 function fateman1(degree::Int)
     T = Int128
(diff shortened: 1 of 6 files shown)


lbenet:
Can you also rebase this PR?

While I like a lot the new implementation of the tables, I'm checking if it can be optimized a bit more. In my machine, the Fatemann tests (julia 0.4) lasts 3.84 sec, versus Mathematica 3.17, so we are within 20%.

dpsanders:
In my next work I have gained another factor of 2.5 ;)  I'll put it up as a new PR, but I am having problems rebasing this one...

lbenet:
Tests are currently failing due to some recent change JuliaLang/julia#12285. 

Shall I try to fix this and incorporate the modifications in the tests?

dpsanders:
I have fixed the FactCheck syntax, that has changed from `=>` to `-->`. However, there is a mess with merge conflicts due to rebasing the previous PRs :/

lbenet:
Rebase and squash is done... Tests pass, so I'm merging.

PR #25, opened 2015-07-26
Clean up files

dpsanders:


lbenet:
Please correct me if I'm wrong, but I think the commints in this PR are already in #23. If so, I think we can simply close this PR.

dpsanders:
I think you're right; in any case they're rather trivial.

PR #26, opened 2015-07-30
Convert arrays of coefficients [3,4,5,6] into Int64s by using arithmetic base (order+1).

dpsanders:
E.g. [3,4,5,6] with order 9 becomes the integer 3456.

### DIFF
--- a/src/TaylorN.jl
+++ b/src/TaylorN.jl
@@ -192,7 +192,7 @@ taylorN_variable(nv::Int, order::Int=get_order()) =
 ## get_coeff
 function get_coeff(a::HomogeneousPolynomial, v::Array{Int,1})
     @assert length(v) == get_numvars()
-    kdic = hash(v)
+    kdic = in_base(get_order(),v)
     @inbounds n = pos_table[a.order+1][kdic]
     a.coeffs[n]
 end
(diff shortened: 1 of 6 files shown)


lbenet:
I think it is now a good point to rebase to the current master; I think the only commit that introduces something new for this PR is https://github.com/JuliaDiff/TaylorSeries.jl/commit/5e5962b. I guess the easiest is to implement those changes in a new branch from the current master, and force-push them to this branch.

In any case, I'll be playing

lbenet:
I cleaned up the PR, using the (only) important commit https://github.com/JuliaDiff/TaylorSeries.jl/commit/5e5962b

Tests are not passing. I'll check now this in detail.

dpsanders:
Oh sorry, I only changed * and ^; when I have a moment I'll try to get the tests to work.

lbenet:
Can you review the changes I implemented on top of your commit? Tests pass (locally, at least), and your changes indeed increased the performance! The beauty, in my opinion, lies in * and square, where the trick to pass the vectors to integers allows to simply sum the integers!

lbenet:
I also added a test involving two of the function in fateman.jl; while the whole series is obtained, I check one specific coefficient for both implementations, which happens to be larger than `typemax(Int)`.

Tests pass!

dpsanders:
LGTM -- merging!

dpsanders:
Note that there are very large numbers of allocations and a lot of memory allocated, even though the time required is half what it previously was. This suggests that more optimization should be possible...

PR #28, opened 2015-08-02
Small clean up

dpsanders:


lbenet:
I just noticed that the mathematica nb takes about ~98% of the lines coded in the project, so Julia seems not to be the important part of the code. 
I thus propose to erase the notebook and, if you still consider it important, simply include a tiny script which has the commands used to measure the performance in mathermatica.

lbenet:
Great that you noticed that `iaux` was actually irrelevant for `*`. The memory used in the test by Fateman is 1/4 of what it was!

I'm merging!

lbenet:
Ups... I just noticed that I didn't replace the mathematica notebook by a simple script...

PR #30, opened 2015-08-02
WIP: Use in-place multiplication

dpsanders:
This PR reduces memory allocation using the function mul! that adds `a*b` directly to `c`, instead of creating a new `HomogeneousPolynomial` for `a*b` which is only then afterwards added to `c`.

However, it does not seem to affect performance very much...

### DIFF
--- a/src/TaylorN.jl
+++ b/src/TaylorN.jl
@@ -334,83 +334,21 @@ end
 
 ## Multiplication ##
 function *(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
-
     T = promote_type( eltype(a), eltype(b) )
     order = a.order + b.order
     if order > get_order()
         return HomogeneousPolynomial(zero(T), get_order())
     end
 
-    iszero(b) && return HomogeneousPolynomial(zero(T), order)
-    # the code below checks quickly whether a is zero
-
-    @inbounds num_coeffs_a = size_table[a.order+1]
-    @inbounds num_coeffs_b = size_table[b.order+1]
-    @inbounds num_coeffs  = size_table[order+1]
     if eltype(a) != eltype(b)
         a, b = promote(a, b)
     end
 
-    coeffs = zeros(T, num_coeffs)
-    @inbounds posTb = pos_table[order+1]
-
-    @inbounds for na = 1:num_coeffs_a
-        ca = a.coeffs[na]
-        ca == zero(T) && continue
-        inda = index_table[a.order+1][na]
-
-        @inbounds for nb = 1:num_coeffs_b
-            cb = b.coeffs[nb]
-            cb == zero(T) && continue
-            indb = index_table[b.order+1][nb]
-
-            pos = posTb[inda + indb]
-            coeffs[pos] += ca * cb
-        end
-    end
-
-    return HomogeneousPolynomial{T}(coeffs, order)
+    res = HomogeneousPolynomial(zero(T), order)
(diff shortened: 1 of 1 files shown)


lbenet:
To shorten the code, and test performance, we could replace the current implementation of
`*(a::HomogeneousPolynomial, b::HomogeneousPolynomial)` by something involving
`mul!`.

What do you think about this?

dpsanders:
It's certainly worth making a branch with that idea.

dpsanders:
I think we should also rewrite `square` to use `mul!` everywhere and check the performance and memory differences. So far it doesn't seem to make much difference. Have you checked?

dpsanders:
For me, the `in_place_mulitplication` branch gives quite a lot less memory allocation, but it is, if anything, slower!  Not sure what's going on. Probably needs more playing around and work. It was really just supposed to show the lines I was thinking along.

lbenet:
I do think that if memory allocation is less, this will eventually pay back, so it is certainly worth pursuing. 
Precisely because of this, I guess that `mul!` should then be used consistently everywhere, and the easiest is to implement it in `*` among `HomogeneousPolynomial`. This will certainly have consequences everywhere, since all elementary operations use it at some stage.

I'll take a look tonight.

dpsanders:
Not exactly. The point is that `*` creates a new `HomogeneousPolynomial`, whereas `mul!` doesn't, which is why the allocation is significantly less for `mul!` (which also needs a third argument, namely the place to store the result of the multiplication).

lbenet:
I've checked what you mentioned. Indeed, there is a slow down despite of improving memory usage. 

I actually modified `*(a::HomogeneousPolynomial, b::HomogeneousPolynomial)` to use `mul!`, but nothing changed... I'll be playing with this because I think this may indeed improve the performance.

lbenet:
The Fateman tests that are more sensible to these changes are fateman2 and fateman4...

src/TaylorN.jl
```diff
@@ -585,6 +626,9 @@ function square(a::HomogeneousPolynomial)
     return HomogeneousPolynomial{T}(coeffs, order)
 end
 
+square(a::HomogeneousPolynomial) = a*a
```
lbenet: I think this line is part of the problem; the function defined on line 599 on is optimized (roughly a factor 2 in comparison with a*a). While the differences are not too bad,, I'm not quite getting the results without `mul!`...

lbenet:
I've been playing all day with this, and I can't understand why it is not improving the speed. 

I've tried to improve different aspects of `mul!`, and extending it to also work in `square`, without important improvements (though from time to time I got marginally better results than the current master).

dpsanders:
Is your branch available?

lbenet:
I just pushed the branch named "test_in_place_mult". It is essentially your branch, rebased to the current master, with new uses of `mult!` in `square`, and cleaning-up.

lbenet:
Interestingly, `fateman1` and `fateman3` functions seem to be more sensitive to the changes introduced by `mul!`, in the negative sense...

lbenet:
I have rebased to the last master, to allow #35 to be merged in this branch, and then merge it to master. Do you have anything against this?

lbenet:
I will clean-up this before merging to master...

lbenet:
I just cleaned-up the mess i built-up. Sorry. Waiting for the test to pass to then merge

PR #31, opened 2015-08-07
Replace Mathematica .nb by a .m script, and adapt fateman.jl

lbenet:
This saves LOT OF lines, and seems to be the correct way of comparing
the time of the calculation. The time returned is the minimum time
for performing 5 times the calculation.

To be fair, fateman.jl also computes the minimum of 5 runs. A fair comparison of 
what Mathematica does, one has to compare with fateman2 or fateman4 output.

### DIFF
--- a/perf/fateman.jl
+++ b/perf/fateman.jl
@@ -45,8 +45,13 @@ function run_fateman(N)
         println("Running $f")
         @time result = f(N)
         push!(results, result)
+        t = Inf
+        for i = 1:5
+            ti = @elapsed f(N)
+            t = min(t,ti)
+        end
+        println("\t Min time of 5 runs:", t)
     end
-
     results
 end
 
--- a/perf/timing_Fateman.m
+++ b/perf/timing_Fateman.m
@@ -1,0 +1,27 @@
+(* Timing a test by Fateman
+
+In a mac, run this as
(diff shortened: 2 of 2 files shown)


lbenet:
I must apologize for first adding a file with nearly 90k lines, and now removing it. I simply don't have a lot of experience with Mathematica, and even less to time its performance.

With regards to performance, in my machine I obtained:
```
$ /Applications/Mathematica.app/Contents/MacOS/MathematicaScript -script ~/perf_fateman.m
mathematica,fateman,3.074024

$ julia perf/fateman.jl 
Running Fateman with order 20...
Running fateman1
   2.631 seconds      (8980 allocations: 58489 KB, 0.21% gc time)
	 Min time of 5 runs:2.601419828
Running fateman2
   1.328 seconds      (7002 allocations: 41768 KB, 0.30% gc time)
	 Min time of 5 runs:1.318467365
Running fateman3
   2.704 seconds      (12998 allocations: 118 MB, 0.31% gc time)
	 Min time of 5 runs:2.703078027
Running fateman4
   1.345 seconds      (11400 allocations: 106 MB, 0.52% gc time)
	 Min time of 5 runs:1.341863851
Done.
```

So, with regards to performance, Mathematica effectively simplifies and then should compared to either fateman2 or fateman4's results. Then, Mathematica is a factor 2 slower than our implementation.

dpsanders:
Nice! Merging.

Issue #24, opened 2015-07-23
Nasty behaviour of `sqrt(x)`


dpsanders:
I think the following should throw an error instead.

```
julia> x, y = set_variables("x y", order=8)
2-element Array{TaylorSeries.TaylorN{Float64},1}:
  1.0 x + 𝒪(‖x‖⁹)
  1.0 y + 𝒪(‖x‖⁹)

julia> sqrt(x)
 Inf x - NaN y - Inf x² - NaN x y - NaN y² + Inf x³ - NaN x² y - NaN x y² - NaN y³ - Inf x⁴ - NaN x³ y - NaN x² y² - NaN x y³ - NaN y⁴ + Inf x⁵ - NaN x⁴ y - NaN x³ y² - NaN x² y³ - NaN x y⁴ - NaN y⁵ - Inf x⁶ - NaN x⁵ y - NaN x⁴ y² - NaN x³ y³ - NaN x² y⁴ - NaN x y⁵ - NaN y⁶ + Inf x⁷ - NaN x⁶ y - NaN x⁵ y² - NaN x⁴ y³ - NaN x³ y⁴ - NaN x² y⁵ - NaN x y⁶ - NaN y⁷ - Inf x⁸ - NaN x⁷ y - NaN x⁶ y² - NaN x⁵ y³ - NaN x⁴ y⁴ - NaN x³ y⁵ - NaN x² y⁶ - NaN x y⁷ - NaN y⁸ + 𝒪(‖x‖⁹)
```

dpsanders:
I guess when the constant term is zero.

dpsanders:
Same for `log`

dpsanders:
I now get 
```julia
x, y = set_variables("x y", order=8)
julia> sqrt(x^2)
ERROR: ArgumentError("The 0-th order TaylorN coefficient must be non-zero\nin order to expand `sqrt` around 0")
 in sqrt at /Users/dsanders/.julia/v0.3/TaylorSeries/src/TaylorN.jl:615
```
but it works for one variable:
```
julia> t = taylor1_variable(10)
 1.0 t + 𝒪(t¹¹)

julia> sqrt(t^2)
 1.0 t + 𝒪(t¹¹)
```

lbenet:
Yes, you are right; this is something known. 

[...]

PR #33, opened 2015-08-20
Solves issue #24 throwing an ErrorException; includes tests

lbenet:


lbenet:
I don't love the error messages, but are ok for the time being. Feel free to change them (also in Taylor1). I don't know yet how to use `throw` properly...

Tests pass.

dpsanders:
To use `throw` you need to pick what type of exception it is: http://julia.readthedocs.org/en/latest/manual/control-flow/#exception-handling

E.g. for the `sqrt`, it's either a DomainError or an ArgumentError. So something like

```julia
if p0 == 0
    throw(ArgumentError("First Taylor coefficient must be non-zero for sqrt"))
end
```
This is tested with
```julia
@fact_throws ArgumentError sqrt(0)
```

lbenet:
Thanks. I'll do the changes later... no time right now.

lbenet:
Thanks for the comments. Errors are now handled with `throw`, including some tests that were commented.

Tests pass...

lbenet:
I added one new commit, to replace `FloatingPoint` by `AbstractFloat`, and also removed unnecessary @compat statements

Tests are not passing though, because the tagged version of `Compat.jl` doesn't yet include the changes related to `AbstractFloat`... I opened an issue: https://github.com/JuliaLang/Compat.jl/issues/135

This can be merged once a new tagged version of Compat.jl appears.

lbenet:
New version tagged for Compat.jl; tests pass. I'm merging this...

dpsanders:
Looks nice!

dpsanders:
Hmmm, don't you need to add the required minimum version of `Compat` in the `REQUIRE` file?

lbenet:
I'm not quite sure what do you mean. I thought that `Compat.jl` precisely was taking care of this.

dpsanders:
You need to specify which version of `Compat.jl` is required for `TaylorSeries.jl`. Otherwise, if someone tries to use the package with a previous version of `Compat.jl`, the code will break, e.g. with Julia 0.3 and Compat 0.5.1 I get

```
julia> using TaylorSeries
ERROR: AbstractFloat not defined
```
If you change `Compat` to `Compat 0.6` in `REQUIRE`, then when I try to install `TaylorSeries` it should make sure that necessary version of `Compat` is installed.

lbenet:
Thanks for clarifying this.

I just tried to push a commit with the corresponding change, but I can't see it in this branch...

Issue #32, opened 2015-08-17
Tag a new version


dpsanders:
I think we should tag at least a new patch version.

lbenet:
I've been thinking about this as well. Shall we include in the tagged version an improvement related to #24, and perhaps #30 as well?

lbenet:
I will update the documents with the new improvements before tagging a new version

lbenet:
New version is tagged. In the documentation I only updated the bechmarks vs Mathematica (after tagging). I close this

PR #35, opened 2015-08-25
Test in place mult

lbenet:
I've extended the use of `mul!` to other functions (and brought back a check with `iszero`); maybe its use can be extended even more. My benchmarks point out that there is an improvement in memory, which you already noticed, and regarding the time there is not much improvement, but rather fine oscillations that make it difficult to judge if the changes improve or not.

In any case, and despite it cannot be automatically merged, I am opening this pull request to add my changes to the branch you opened, and go on and merge this PR into master (before tagging a new version #32). The thing is that in my other machine, the benchmarks look better, probably because I have more RAM, and less use of browsers... 

What do you think?

PS: I think the merge does not work automatically because your branch lies back with respect to master, so you have to rebase (some of the commits below are already in master).

### DIFF
--- a/src/TaylorN.jl
+++ b/src/TaylorN.jl
@@ -334,83 +334,21 @@ end
 
 ## Multiplication ##
 function *(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
-
     T = promote_type( eltype(a), eltype(b) )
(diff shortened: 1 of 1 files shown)


lbenet:
Merging this...

PR #36, opened 2015-08-27
Update the documentation to reflect bechmark with Mathematica

lbenet:
Fateman's test is faster with TaylorSeries than with Mathematica,
even when avoiding certain possible optimizations

:-)


lbenet:
Once this is merged, I'll update gh-pages

PR #42, opened 2016-01-31
Precompilation and solves a printing issue for T<:Real but not the usual cases

lbenet:
An example is when T is Interval; in that case, showing the
Taylor series displayed odd results, though the coefficients
were correct

It also adds precompilation

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,3 @@
-julia 0.3
+julia 0.4
 Compat 0.6
-Docile
 FactCheck 0.3.0
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -6,10 +6,6 @@ __precompile__(true)
 
 module TaylorSeries
 
-if VERSION < v"0.4.0-dev"
-    using Docile
-end
-
 using Compat
 
 import Base: ==, +, -, *, /, ^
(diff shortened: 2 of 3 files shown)


lbenet:
The following is a simple example; currently in master one gets
```julia
julia> using TaylorSeries
julia> using ValidatedNumerics
julia> Taylor1([0,@interval(-1,1)],3)
 - [0.0, 1.0] t + 𝒪(t⁴)
```

while, in the "printing" branch,
```julia
julia> using TaylorSeries
INFO: Recompiling stale cache file /Users/benet/.julia/lib/v0.4/TaylorSeries.ji for module TaylorSeries.
julia> using ValidatedNumerics
julia> Taylor1([0,@interval(-1,1)],3)
 [-1.0, 1.0] t + 𝒪(t⁴)
```

This is related to the fact that `numbr2str` operates using the sign, testing if the coefficient is >0, and if not, it puts a minus sign and attaches the abs. This is ok for numeric coefficients, but not for interval coefficients.

lbenet:
The commit that should be tagged as version 0.1.2 should be 3d9ba8f, the current in master. I guess we should tag that version before merging this. I don't know if we will need to rebase this afterwards or not.

lbenet:
This PR should not include precompilation, simply because the printing improvement can also be used in julia 0.3 without any problem; sorry, my mistake.

So, I am closing this PR, and will open another only with the printing improvements that we have  discussed here. This version should be tagged as the last version compatible with julia 0.3. Then, in a new PR we will include the precompilation and all the corresponding changes.

dpsanders:
Sounds good.

PR #43, opened 2016-03-04
Corrects a printing issue for T<:Real (but not the usual cases)

lbenet:
An example is when T is Interval, thus T<:Real; in that case, displaying the Taylor series shows odd results, though the coefficients internally are correct. An example of the problem is given in [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/42#issuecomment-177601984). 

This follows from #42; it only deals with the output of the series, and incorporates all suggestions of #42.

### DIFF
--- a/src/printing.jl
+++ b/src/printing.jl
@@ -91,12 +91,15 @@ function homogPol2str{T<:Number}(a::HomogeneousPolynomial{T})
     end
     return strout[1:end-1]
 end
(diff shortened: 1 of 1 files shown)


lbenet:
@dpsanders There is a problem with the tests only in the nightly version for OSX; it seems to be a problem related with the dmg image... I'll try again later.

lbenet:
Merging...

PR #44, opened 2016-04-03
Adds `abs` function with tests

lbenet:
... and slightly corrects some error messages.


lbenet:
I'm merging to then tag the corresponding version in master as the last one that is compatible with Julia 0.3. (Our set up for travis is not testing 0.3, so I have checked it locally and tests pass nicely.)

PR #47, opened 2016-04-29
Docs documenter

blas-ko:
Perdón que cerré el Pull Request pasado.

Documenté todo el `Taylor1.jl`incluyendo las relaciones de recurrencia para los coeficientes de `HomogCoefs`. Se utilizaron caracteres de UNICODE ya que `Documenter.jl` aún no funciona bien con ecuaciones para la documentación interna.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,15 +1,13 @@
 using Documenter, TaylorSeries
 
 makedocs(
-    # modules = TaylorSeries,
+    modules = Module[TaylorSeries],
     clean   = false,
     doctest = false,
 )
 
 deploydocs(
-    deps = Deps.pip("pygments", "mkdocs", "mkdocs-bootstrap", "python-markdown-math"),
+    deps = Deps.pip("pygments", "mkdocs", "mkdocs-cinder", "python-markdown-math"),
     repo   = "github.com/JuliaDiff/TaylorSeries.jl.git",
-    julia  = "release",
-    osname = "osx",
 )
 
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -3,8 +3,8 @@ site_description: 'TaylorSeries.jl: A julia package to manipulate Taylor expansi
 repo_url:         https://github.com/JuliaDiff/TaylorSeries.jl
 site_author:      Luis Benet
 
-# theme: cinder
-theme: bootstrap
+theme: cinder
+# theme: bootstrap
 
 extra_css:
   - assets/Documenter.css
@@ -22,7 +22,7 @@ markdown_extensions:
 
 docs_dir: 'build'
 
-copyright: "Copyright &copy; 2014-2016, Luis Benet and David P. Sanders.<br>
+copyright: "Copyright &copy; 2014-2016, Luis Benet and David P. Sanders.
 <a https://github.com/JuliaDiff/TaylorSeries.jl>TaylorSeries.jl</a> is licensed under the
 <a href='https://github.com/JuliaDiff/TaylorSeries.jl/blob/master/LICENSE.md'>
 MIT Expat license</a>."
@@ -30,6 +30,6 @@ MIT Expat license</a>."
 pages:
     - 'Home': 'index.md'
     - 'Background': 'background.md'
-    - 'User guide': 'user_guide.md'
+    - 'User guide': 'userguide.md'
     - 'Examples': 'examples.md'
     - 'API': 'api.md'
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -1,37 +1,42 @@
-    {meta}
-    CurrentModule = TaylorSeries
-
 # Library
 
 ---
 
+```@meta
+CurrentModule = TaylorSeries
+```
+
 ## Types
 
-    {docs}
-    TaylorSeries.Taylor1
-    TaylorSeries.HomogeneousPolynomial
-    TaylorSeries.TaylorN
-    TaylorSeries.ParamsTaylorN
+```@docs
+Taylor1
+HomogeneousPolynomial
+TaylorN
+ParamsTaylorN
+```
 
 ## Functions and methods
 
-    {docs}
-    TaylorSeries.taylor1_variable()
-    TaylorSeries.diffTaylor()
-    TaylorSeries.integTaylor()
-    TaylorSeries.deriv()
-    TaylorSeries.evaluate()
+```@docs
+taylor1_variable
+taylorN_variable
+set_variables
+show_params_TaylorN
+get_coeff
+```
 
-    TaylorSeries.taylorN_variable()
-    TaylorSeries.set_variables()
-    TaylorSeries.show_params_TaylorN()
-    TaylorSeries.get_coeff()
+```@docs
+evaluate
+derivative
+integrate
+gradient
+jacobian
+hessian
+```
 
-## Internal use
+## Internals
 
-    {docs}
-    TaylorSeries.generate_tables
-    TaylorSeries.coeff_table
-    TaylorSeries.index_table
-    TaylorSeries.size_table
-    TaylorSeries.pos_table
+```@docs
+generate_tables
+mulHomogCoef
+```
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -1,7 +1,3 @@
-
-    {meta}
-    CurrentModule = TaylorSeries
-
 # Examples
 
 ---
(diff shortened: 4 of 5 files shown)


lbenet:
No problem, but use English here. I check your PR in the next few days.

lbenet:
As we discussed, I suggest:
- Use one line imperative statement about the function, and in another paragraph provide details,
- Use TeX format for formulas instead of UTF-strings

blas-ko:
Changes about the suggestions were made. 
Docstrings for `Taylor1.jl` were all included in a `src/helpDB.jl` script which, by the way, I'm not sure if it's the best place to put it. 

Expansion coefficients for the `HomogCoef`are now in LateX format which can be seen nicely in a notebook but not in the REPL.

lbenet:
Thanks for everything!

I have few comments on the style:
- Use imperative form, i.e., "Compute" instead of "Computes".
- When referring to functions, I think it is better to use latex-style; check what you get for `? exp` (in Base they use latex instead of markdown).
- I would avoid any html commands such as `<br>` or `<center>`, as well as intermediate empty lines (before and after the equations); they are not nicely formatted.
- There are some commits which are not needed; I will clean those before merging.

I think you can add to `helpDB.jl` what already exists in `TaylorN.jl` (and in the other files).

dpsanders:
Is there a particular reason why you chose to have a separate `helpDB.jl` file? This is something that will be gradually removed from Base Julia. I would recommend instead putting the docstrings together with their functions.

src/helpDB.jl
```diff
@@ -0,0 +1,296 @@
...
+component is the coefficient of degree \$i-1\$ of the expansion.
+- `order  :: Int64` Maximum order (degree) of the polynomial.
+"""->
+Taylor1
+
+@doc """
+    taylor1_variable(T, [order=1])
+    taylor1_variable([order=1])
+
+Short-cut to define the independent variable as a `Taylor1` polynomial of
```
dpsanders: Prefer `shortcut` to `short-cut`

src/helpDB.jl
```diff
@@ -0,0 +1,296 @@
...
+@doc """
+    get_coeff(a, n)
+
+Return the coefficient of order `n::Int` of a `a::Taylor1` polynomial.
+""" ->
+get_coeff
+
+## Differentiating ##
+doc"""
+    diffTaylor(a)
```
dpsanders: Could this just be called `diff`?

src/helpDB.jl
```diff
@@ -0,0 +1,296 @@
...
+
+Return the `Taylor1` polynomial of the differential of `a::Taylor1`. 
+
+The last coefficient is set to zero.
+"""
+diffTaylor
+
+## Integrating ##
+doc"""
+    integTaylor(a, x)
```
dpsanders: `integrate`?

src/helpDB.jl
```diff
@@ -0,0 +1,296 @@
...
+    evaluate(a, x)
+
+Returns the substitution of `x::Taylor1` as independent variable in
+`a::Taylor1`.
+"""
+evaluate
+
+## Returns de n-th derivative of a series expansion
+doc"""
+    deriv(a, [n=1])
```
dpsanders: Could this be `derivative` to agree with `ForwardDiff`?

src/helpDB.jl
```diff
@@ -0,0 +1,296 @@
...
+"""
+deriv
+
+
+                                        ### Function Extensions ###
+
+
+## Exp ##
+doc"""
+    exp(a)
```
dpsanders: I do not see these docstrings when I do `?exp` at the REPL.

lbenet:
> Is there a particular reason why you chose to have a separate helpDB.jl file? This is something that will be gradually removed from Base Julia. I would recommend instead putting the docstrings together with their functions.

That was a suggestion I made to Blas based on what is (currently) done in Base (though i think `helpdb.jl` is a nicer name, without the "DB" in capitals); I think Blas disliked this idea. 
What I do not like of having the docstrings together with the code is that Julia internally (re)orders them in a way which I'd like to control (and do not know how). For instance, `? get_coeff` currently display the methods for `TaylorN` then for `HomogeneousPolynomial` and finally for `Taylor1`. I must admit that what I like of having the docstrings together with the code is precisely the essence of literate programing...

If you are strongly in favor of having them together with the code, I think this is the good time to revert things and clean up some commits. (Sorry Blas about thist!)

src/helpDB.jl
lbenet: `diff` is used (in Base) for finite differences (of a matrix or vector). Here we return the polynomial (`Taylor1` or `TaylorN`) which corresponds to the differential (or partial derivative). While I agree that the name is not the best, I think we should distinguish its functionality.

src/helpDB.jl
lbenet: Fully agree! Thanks!

src/helpDB.jl
lbenet: I agree!

dpsanders:
Yes, I'm definitely strongly in favour of having them together with the corresponding functions in the code; as you say, this is indeed literate programming.

As regards the order of appearence of the functions when you do `? get_coeff`, this is a good question, and something that should probably be raised on the Julia-users list or as an issue in the main Julia repo.

src/helpDB.jl
lbenet: At the beginning when I was playing with this, I was fed up with recompiling the package constantly, so I commented the relevant line in `TaylosSeries.jl`; if you uncomment it, it will recompile it and everything should work.

src/helpDB.jl
dpsanders: OK, agreed. What is the difference between `diffTaylor` and `deriv` then? Can they somehow be combined into `derivative`?

lbenet:
@blas-ko My mistake, Blas...

This is indeed in the spirit of literate programing, but not quite... But that's another thing.

dpsanders:
According to https://groups.google.com/forum/m/#!topic/julia-users/WPHhQtVxCVs
In Julia 0.5, methods in documentation will be listed in order if their definition.

lbenet:
Thanks for the info!

src/helpDB.jl
blas-ko: How do `deriv` and `diffTaylor` should be modified then ? Are they going to keep their names as they are, or should both be different cases for a `derivative` function?

blas-ko:
I moved back the docstrings into the `Taylor1.jl` script. If everything's okay, I'll work on documentation for `TaylorN.jl` and delete `helpDB.jl` so we can merge.

lbenet:
Thanks for pushing an updated version. I will correct the docstrings and clean up a little bit this PR, because there is a bunch of commits which do not provide anything. If I can't keep the authorship of the relevant commits, you will clean up the commits.

lbenet:
Checking today the last commit by @blas-ko , I noticed that I never responded to [this](https://github.com/JuliaDiff/TaylorSeries.jl/pull/47#discussion_r62714691) and [this](https://github.com/JuliaDiff/TaylorSeries.jl/pull/47#discussion_r63308993) comments. The question is:

> What is the difference between `diffTaylor` and `deriv` then? Can they somehow be combined into `derivative`?

- `diffTaylor` returns either a `Taylor1` or a `TaylorN` polynomial, which correspond to the derivative or partial derivative with respect to a given variable. I think we could call this `differential`.

- `deriv` returns the *value* of the n-th derivative of a `Taylor1` polynomial. (Something alike could be implemented for generic partial derivatives of `TaylorN` polynomials.) In Base, there is `diff`, which returns the finite difference operator. I think we should stick to `deriv`.

@dpsanders You are good at naming functions. How do you like this proposal?

lbenet:
Once we decide this, I'll change the code, the docs and the tests accordingly.

dpsanders:
Can't they both be called derivative? derivative(p)  is the polynomial, and derivative(p, a)  is the derivative evaluated at the point a?

lbenet:
Currently, we have `diffTaylor(p)` and `diffTaylor(p,[r=1::Int])` for `Taylor1` and `TaylorN`; the former returns the Taylor1 polynomial of the derivative while the latter returns the TaylorN polynomial of the (partial) derivative with respect to the `r` variable. `deriv(p, n::Int)` for a `Taylor1` polynomial, returns the *value* of the n-th derivative (of the Taylor expansion at the value around which the expansion is made). 

My point is that it makes things ambiguous if we use the same name to return different quantities (polynomial of the differential vs value of a given derivative). Moreover, we could define  `diffTaylor(p::Taylor1,[r=1::Int])` to return the polynomial corresponding to the r-th derivative, which would collide with `deriv(p,n)`, if we use the same names.

What about `diffTaylor` be `differential` and `deriv` be `derivative`?

dpsanders:
I would find that very confusing. I suggest derivative for both (the Julian style :) )  but change the order of the arguments, so derivative(i, p) is the ith derivative and derivative(p, a) is evaluated at a. And derivative(i, P, a) is the ith derivative evaluated at a!

dpsanders:
Note that all tests are failing on travis.

lbenet:
I like `derivative` for all possibilities, with the subtlety of the order! I'll implement this later in a different PR, so we are not stepping in our toes.

PR #48, opened 2016-05-10
Added A_mul_B! and some small improvements

pwl:
This greatly improves the performance of multiplying `Vector{Taylor1{T}}` by `Matrix{T}`.  I have to admit that there is a small numerical error of the order `10*eps` when compared to the original `Base.A_mul_B!` but I can't quite locate the source of this inconsistency.  It only shows up for particular arguments, so tests pass most of the time.  Perhaps someone should take a closer look before merging this.  The tests are failing for that reason.

### DIFF
--- a/src/Taylor1.jl
+++ b/src/Taylor1.jl
@@ -676,7 +676,7 @@ end
 
 Multiply a*b and save the result in y.  Extends y to fit the results if necessary.
 """
-function A_mul_B!{T<:Number}(y::Vector{Taylor1{T}},a::Matrix{T},b::Vector{Taylor1{T}})
+function A_mul_B!{T<:Number}(y::Vector{Taylor1{T}},a::Union{Matrix,SparseMatrixCSC},b::Vector{Taylor1{T}})
 
     # determine the maximal order of b
     order = maximum([b1.order for b1 in b])
--- a/src/Taylor1.jl
+++ b/src/Taylor1.jl
@@ -28,12 +28,15 @@ immutable Taylor1{T<:Number} <: Number
     function Taylor1(coeffs::Array{T,1}, order::Int)
         lencoef = length(coeffs)
         order = max(order, lencoef-1)
-        order == lencoef-1 && return new(coeffs, order)
-        resize!(coeffs, order+1)
-        for i = lencoef+1:order+1
-            coeffs[i] = zero(T)
+        if order == lencoef-1
+            return new(coeffs, order)
+        else
+            resize!(coeffs, order+1)
+            for i = lencoef+1:order+1
+                coeffs[i] = zero(T)
+            end
+            return new(coeffs, order)
         end
-        new(coeffs, order)
     end
 end
 
@@ -105,17 +108,17 @@ function firstnonzero{T<:Number}(a::Taylor1{T})
     nonzero
 end
 
-function fixshape{T<:Number, S<:Number}(a::Taylor1{T}, b::Taylor1{S})
-    eltype(a) == eltype(b) && a.order == b.order && return a, b
-    if eltype(a) != eltype(b)
-        a, b = promote(a, b)
-    end
+function fixshape{T<:Number}(a::Taylor1{T}, b::Taylor1{T})
     if a.order == b.order
         return a, b
     elseif a.order < b.order
         return Taylor1(a, b.order), b
     end
     return a, Taylor1(b, a.order)
+end
+
+function fixshape{T<:Number, S<:Number}(a::Taylor1{T}, b::Taylor1{S})
+    fixshape(promote(a,b)...)
 end
 
 ## real, imag, conj and ctranspose ##
@@ -665,4 +668,54 @@ Return the value of the `n`-th derivative of `a`.
 function deriv{T<:Number}(a::Taylor1{T}, n::Int=1)
     @assert a.order >= n >= 0
     factorial( widen(n) ) * a.coeffs[n+1] :: T
+end
+
+
+"""
+    A_mul_B!(y, a, b)
+
+Multiply a*b and save the result in y.  Extends y to fit the results if necessary.
+"""
+function A_mul_B!{T<:Number}(y::Vector{Taylor1{T}},a::Matrix{T},b::Vector{Taylor1{T}})
+
+    # determine the maximal order of b
+    order = maximum([b1.order for b1 in b])
+
+    n,k = size(a)
+
+    # extend y to fit all the coefficients and zero the
+    if !isdefined(y)
+        for i = 1:n
+            y[i] = Taylor1(zero(T),order)
+        end
+    end
+
+    for i = 1:n
+        if y[i].order < order
+            y[i] = Taylor1(zero(T),order)
+        else
+            fill!(y[i].coeffs,zero(T))
+        end
+    end
+
+
+    # do all of b1 have the same order?
+    if reduce(&,[b1.order == order for b1 in b])
+        # if yes this variant is faster
+        B = hcat([b1.coeffs for b1 in b ]...)'
+        Y=a*B
+        for i = 1:n
+            copy!(y[i].coeffs,Y[i,:])
+        end
+    else
+        # slower variant
+        for i = 1:n
+            for l = 1:k
+                for j = 1:length(b[l].coeffs)
+                    y[i].coeffs[j]+=a[i,l]*b[l].coeffs[j]
+                end
+            end
+        end
+    end
+    return y
 end
--- a/src/Taylor1.jl
+++ b/src/Taylor1.jl
@@ -671,25 +671,40 @@ function deriv{T<:Number}(a::Taylor1{T}, n::Int=1)
 end
 
 
-"""
-    A_mul_B!(y, a, b)
+# fix the ambiguities for A_mul_B!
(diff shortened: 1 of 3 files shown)


pwl:
Oh, and I tried to change `Matrix{T}` to `AbstractMatrix{T}` but I got a lot of warnings about ambiguous definitions.  I guess there is a more clever way of extending this definition.

lbenet:
Thanks for contributing! 

I will take a look (in the next few days) and see if I can find the source of the loss of accuracy.

src/Taylor1.jl
```diff
@@ -105,11 +108,7 @@ function firstnonzero{T<:Number}(a::Taylor1{T})
     nonzero
 end
 
-function fixshape{T<:Number, S<:Number}(a::Taylor1{T}, b::Taylor1{S})
-    eltype(a) == eltype(b) && a.order == b.order && return a, b
-    if eltype(a) != eltype(b)
-        a, b = promote(a, b)
-    end
+function fixshape{T<:Number}(a::Taylor1{T}, b::Taylor1{T})
```
lbenet: Good catch! Thanks.

src/Taylor1.jl
```diff
@@ -666,3 +669,53 @@ function deriv{T<:Number}(a::Taylor1{T}, n::Int=1)
...
+Multiply a*b and save the result in y.  Extends y to fit the results if necessary.
+"""
+function A_mul_B!{T<:Number}(y::Vector{Taylor1{T}},a::Matrix{T},b::Vector{Taylor1{T}})
+
+    # determine the maximal order of b
+    order = maximum([b1.order for b1 in b])
+
+    n,k = size(a)
+
+    # extend y to fit all the coefficients and zero the
```
lbenet: I think we may (and should) assume that `y` is already defined, though we need to ensure that it has the proper dimensionality (length `n`). Otherwise, we may run into an error. As an example, using your tests, but changing the dimensionality of Y to something less than 90 yields a `BoundsError`.

test/runtests.jl
```diff
@@ -305,4 +305,37 @@ facts("High order polynomials test inspired by Fateman (takes a few seconds))")
...
+    fake_order = max(n1,k1,order+1)
+    B1 = randn(n1,fake_order)
+    Y1 = randn(k1,fake_order)
+
+    A  = randn(k1,n1)
+
+    # B and Y contain elements of different orders
+    B  = Taylor1{Float64}[Taylor1(collect(B1[i,1:i]),i) for i=1:n1]
+    Y  = Taylor1{Float64}[Taylor1(collect(Y1[k,1:k]),k) for k=1:k1]
+    Bcopy = deepcopy!(B)
```
lbenet: I'm getting an error in the tests on this line; it should read `deepcopy` (without the `!`).

test/runtests.jl
```diff
@@ -305,4 +305,37 @@ facts("High order polynomials test inspired by Fateman (takes a few seconds))")
...
+    # B should be unchanged
+    @fact B==Bcopy --> true
+
+    # is the result compatible with the matrix multiplication?  We
+    # only check the zeroth order of the Taylor series.
+    y1=sum(Y).coeffs[1]
+    Y=A*B1[:,1]
+    y2=sum(Y)
+    # Sometimes this fails due to a small numerical error
+    @fact abs(y1-y2) < tol1 --> true
```
lbenet: I still do not know where the loss of accuracy is happening. Yet, I think that `tol1` should refer to the *local* `eps` with respect yo `y1` or `y2`, or you should be comparing a normalized difference.

test/runtests.jl
```diff
@@ -305,4 +305,37 @@ facts("High order polynomials test inspired by Fateman (takes a few seconds))")
...
 end
 
+facts("Matrix multiplication for Taylor1") do
+    order = 30
+    n1 = 100
+    k1 = 90
+
+    tol1 = eps(1.0)
+
+    fake_order = max(n1,k1,order+1)
```
lbenet: `fake_order` should be identical to `order+1`; it is related to the order of the `Taylor1` polynomials. Or did I miss something?

src/Taylor1.jl
pwl: Without the initialization calling `A_mul_B!` through `C=A*B` results in an error, this is because `C=A*B` expands roughly to `A_mul_B!(similar(B),A,B)` and `similar` doesn't initialize the contents of its result properly.  Another way to solve it would be to implement `similar` for `Array{Taylor,N}`.

test/runtests.jl
pwl: Sorry, that's a typo.

test/runtests.jl
pwl: You are right, the way I wrote it was confusing, fixed that.

test/runtests.jl
pwl: Good point, it looks like the tests are passing if the tolerance is proportional to `y1` and `y2` and the size of the array.  Still this should be fixed in the future.

src/Taylor1.jl
pwl: I tried defining similar for `AbstractArray{Taylor1,N}`, but got into some trouble with ambiguity of the method.

pwl:
I also added the support of `SparseMatrixCSC`.   In principle this method should work for any `AbstractMatrix`, but if I try to definie it for `AbstractMatrix` I run into lots of ambiguous method warnings.

src/Taylor1.jl
lbenet: My point (with the assumption that `y` is already defined) is the convention about naming a function with `!`, which changes the contents of `y`.

lbenet: Regarding `a::AbstractArray{Taylor1,N}`, shouldn't it be `a::AbstractArray{Taylor1,2}` or even  `a::AbstractArray{Taylor1{T},2}`?

lbenet: I've been playing with this; so far, the best I got is with `A_mul_B!{T<:Number}(y::AbstractVector{Taylor1{T}},a::AbstractMatrix{T},b::AbstractVector{Taylor1{T}})`; this still yields two warnings with `AbstractTriangular`.

pwl: Basing on your experiments I managed to fix the ambiguity warnings and I added the dimension checks.

Also I don't get what you mean by the naming convention and changing `y`.  Can you explain it?

lbenet: Thanks for adding the checks on the dimensions; incidentally, you may consider `@assert` next time, which is maybe not elegant, but it is exploited at compile time.

lbenet: > Also I don't get what you mean by the naming convention and changing y. Can you explain it?

In Julia the convention is that functions ending with `!` imply that (at least) one of the arguments is changed. In this case it is `y`, which means that it must be already defined. In other words, it makes no sense to use this specialized formula if `y` is not already defined. If you test `A_mul_B!(z,A1,B)` using the definitions in the tests, it will through an error because `z` doesn't exist. So the `if` block in lines 700-705 is never executed.

lbenet: A possible optimization of your code that occurred to me, based on your "faster variant", could be the following: Create matrices `B` that hold the coefficients of `b`, and `Y` which will hold the coefficients of `y`, the latter set to zero. Compute `A_mul_B!(Y, A, B)`, and form `y`.

I think this may avoid some `for` loops and exploits  `A_mul_B!` from Base, as in your faster variant.

lbenet: I've done some benchmarks about the proposal above. It indeed improves the timing in a factor 5-10. If you don't implement it, I'll push the changes.

pwl: I completely agree that it doesn't make sense to initialize `y` in `A_mul_B!` and that the convention is to assume that `y` is ready to be written on.  But this convention relies on an implementation of a function `similar`, which `TaylorSeries` doesn't have yet.  Once it's there you could safely remove the lines you mentioned but I would leave them for now as a temporary measure.

As for the improvement, feel free to merge this PR and implement the correction yourself.  Although I have to admit I don't quite get what you have in mind.  Isn't it how it already works?  Calling `Y=A*B` in my version is roughly equivalent to `B=A_mul_B!(similar(B),A,B)`.

lbenet: Ok, I'll do the changes. It would be nice if you can also benchmark the changes.

As for your comment on `similar`, the following in master (inspired in your PR, but without it!) currently works; note that I am *not* creating `Y1`, and `Y` is obtained from the operation:
```julia
julia> using TaylorSeries

julia> order = 30;

julia> n1 = 4;

julia> k1 = 3;

julia> B1 = randn(n1,order);

julia> A1  = randn(k1,n1);

julia> B  = Taylor1{Float64}[Taylor1(collect(B1[i,1:i]),i) for i=1:n1];

julia> Y = A_mul_B!(similar(B, k1), A1, B)
3-element Array{TaylorSeries.Taylor1{Float64},1}:
  0.27316308514920934 + 2.3361726531529268 t + 1.2648372715206755 t² + 0.04987013186300208 t³ + 𝒪(t⁵)
      1.896823400442119 - 0.8006875447982573 t - 2.232382532282572 t² + 0.8926690617877661 t³ + 𝒪(t⁵)
   - 0.2830616479144453 - 3.470193293491136 t - 5.1783732101128725 t² + 2.8494688946414692 t³ + 𝒪(t⁵)

```

The result is at most of order 5 because that is the maximum order of `B`.

lbenet:
@pwl Merged! Thanks for contributing!

PR #55, opened 2016-06-27
Remove warnings (in Julia 0.5) using `@compat`

lbenet:


lbenet:
Merged. It only cleans up the output, which is nice for the tests.

Issue #56, opened 2016-06-27
Make TaylorSeries work with DoubleDouble


dpsanders:
There are a string of conversion errors currently.

lbenet:
Could you be more specific on this? I haven't worked with DoubleDluble...

dpsanders:
```
julia> using TaylorSeries, DoubleDouble

julia> t = taylor1_variable(Double{Float64}, 10)
 Double(1.0, 0.0)
 - value: 1 t + 𝒪(t¹¹)

julia> t + t^2
 Double(1.0, 0.0)
 - value: 1 t + Double(1.0, 0.0)
 - value: 1 t² + 𝒪(t¹¹)

julia> (t+1)*(t+1)
 Double(1.0, 0.0)
 - value: 1 + Double(2.0, 0.0)
 - value: 2 t + Double(1.0, 0.0)
 - value: 1 t² + 𝒪(t¹¹)

julia> sin(t)
ERROR: MethodError: Cannot `convert` an object of type DoubleDouble.Double{Float64} to an object of type AbstractFloat
This may have arisen from a call to the constructor AbstractFloat(...),
since type constructors fall back to convert methods.
 in sincos(::TaylorSeries.Taylor1{DoubleDouble.Double{Float64}}) at /Users/dpsanders/.julia/v0.5/TaylorSeries/src/Taylor1.jl:718
 in sin(::TaylorSeries.Taylor1{DoubleDouble.Double{Float64}}) at /Users/dpsanders/.julia/v0.5/TaylorSeries/src/Taylor1.jl:703
 in eval(::Module, ::Any) at ./boot.jl:234
 in macro expansion at ./REPL.jl:92 [inlined]
 in (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:46
```

This is after I fixed one missing conversion in DoubleDouble; it's possible that the problem is there, but I don't understand where the convert to AbstractFloat is happening in the TaylorSeries code.

lbenet:
The problem is related to the fact that `sin(Double(0.0))`, as well as other elementary functions, are not defined for `Double`s; see also https://github.com/simonbyrne/DoubleDouble.jl/issues/8. This is needed by `TaylorSeries`, since the very first coefficient of a Taylor series is the function (`sin`) evaluated around the point you expand, in this case, `sin(Double(0.0))`.

dpsanders:
[...]

PR #59, opened 2016-06-28
Proof of concept of not calculating exp(coeffs[0])

dpsanders:


lbenet:
I agree that this does work; however, it should be part of the implementation of `sin` for `Double`, and not of `sin` for `Taylor1` or `TaylorN`. See https://github.com/JuliaDiff/TaylorSeries.jl/issues/56#issuecomment-229141215

lbenet:
See [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/issues/263#issuecomment-850921898) for a possible implementation of a `qsqrt`; `qexp` would have a similar implementation, though i think in general they won't work.

PR #69, opened 2016-10-19
Explicit method for `Taylor1{Rational{S}}` division by a `Real` or `Complex`

lbenet:
This makes that dividing a Taylor1{Rational{S}} by an integer
returns a Taylor1{Rational{S}}; up to now, it was returning a
Taylor1{Float64}

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ os:
 
 julia:
     - 0.4
-    - release
+    - 0.5
     - nightly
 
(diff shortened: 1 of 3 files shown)


lbenet:
I just corrected an inconsistency that made tests fail in 0.4, added new tests to increase coverage (and also retouched `acosHomogCoef`, which was not being used.

I'll wait to travis before merging.

PR #70, opened 2016-10-28
Adds nicer output and conversion rules for Taylor1{TaylorN{T}}-like cases

lbenet:


lbenet:
Tests are passing, so I merge this.

lbenet:
Added one more test...

PR #71, opened 2016-11-03
Added gaussian and erf functions.

blas-ko:
Added **gaussian** and **erf** functions to the one variable expansions. 
As these functions are not analytical, convergence is not tested.

An extra method `gaussian(μ,σ,order)` is included to automatically create a gaussian function with a given average (`μ`)and standard deviation (`σ`).


lbenet:
Thanks for the contribution! Tests are failing.

blas-ko:
Test pass now. Maybe you can now check this @lbenet

PR #72, opened 2016-11-04
Add methods for convert and jacobian

lbenet:
This PR adds methods for convert and one for the Jacobian, including a minor change.


lbenet:
Merging...

PR #76, opened 2016-11-06
Add series reversion

### LINKED ISSUE #74: Implement Lagrange series inversion?

MikaelSlevinsky:
Closes #74.

only about t = 0 and only if f(0) == 0.

The reason is that it is numerically unstable to re-expand about the
point (t-f(0))^n, and the recursion for the coefficients is simplified
when reverting about t = 0.

Added three tests. 1. checks that reverse(exp(t)-1) == log(1+t)
2. checks the Taylor series of the LambertW function (from Wikipedia)
3. checks the ArgumentError

### DIFF
--- a/src/Taylor1.jl
+++ b/src/Taylor1.jl
@@ -1235,6 +1235,39 @@ function evaluate{T<:Number,S<:Number}(a::Taylor1{T}, x::Taylor1{S})
 end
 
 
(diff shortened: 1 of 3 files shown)


lbenet:
Looks good to me. I'm merging!

Thanks for the contributions!

PR #83, opened 2017-02-09
Deprecations0.6

lbenet:
This is a slight modification of #82, which includes changes so tests also pass in Julia v0.4.

I'm checking about the minimum required Compat.jl version modify `REQUIRE`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ os:
   - osx
 
 julia:
-    - 0.4
     - 0.5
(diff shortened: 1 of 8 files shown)


lbenet:
Tests pass locally...

lbenet:
... but seem to be stalled in travis.

lbenet:
Though there are warnings (in 0.6) not yet addressed, I'm merging to tag a new version.

Issue #85, opened 2017-02-18
Cannot use more than 65 variables with TaylorN


PerezHz:
Was just doing some work on my machine with TaylorN (OS X 10.10.5, julia-0.5; TaylorSeries v0.4.0), and then I came across some unexpected behaviour:
```julia
julia> using TaylorSeries
julia> δv = set_variables("δ", order=1, numvars=65);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₁ + 𝒪(‖x‖²) #this is OK
julia> δv = set_variables("δ", order=1, numvars=66);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₂ + 𝒪(‖x‖²) #the subscript on the δ should be 1 instead of 2
julia> δv = set_variables("δ", order=1, numvars=70);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₆ + 𝒪(‖x‖²) #the subscript on the δ should be 1 instead of 6
julia> δv = set_variables("δ", order=1, numvars=5);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₁ + 𝒪(‖x‖²) #this is OK
julia> δv = set_variables("δ", order=1, numvars=65);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₁ + 𝒪(‖x‖²) #this is OK
```

There seems to be a problem with `TaylorN` when `order==1` and `numvars>65`...

If I use `order=2` instead of `order=1`, everything works fine (this is the workaround I'm using):
```julia
julia> δv = set_variables("δ", order=2, numvars=65);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₁ + 1.0 δ₁² + 𝒪(‖x‖³) #OK
julia> δv = set_variables("δ", order=2, numvars=66);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₁ + 1.0 δ₁² + 𝒪(‖x‖³) #OK
julia> δv = set_variables("δ", order=2, numvars=70);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₁ + 1.0 δ₁² + 𝒪(‖x‖³) #OK
julia> δv = set_variables("δ", order=2, numvars=5);
julia> (1+δv[1])^2
 1.0 + 2.0 δ₁ + 1.0 δ₁² + 𝒪(‖x‖³) #OK
julia> δv = set_variables("δ", order=2, numvars=65);
[...]

PR #86, opened 2017-02-19
Make `index_table` a Vector{Vector{UInt128}}, and add tests

lbenet:
This is a small modification addressing [#85]. It simply carries
on the hashing of the vectors of integers in `UInt128` arithmetic.

### DIFF
--- a/src/hash_tables.jl
+++ b/src/hash_tables.jl
@@ -15,9 +15,9 @@ and `pos_table`. Internally, these are treated as `const`.
 The $i+1$-th component contains a vector with the vectors of all the possible
 combinations of monomials of a `HomogeneousPolynomial` of order $i$.
 
-    index_table :: Array{Array{Int64,1},1}
+    index_table :: Array{Array{UInt128,1},1}
 
-The $i+1$-th component contains a vector of (hashed) indices that represent
+The $i+1$-th component contains a vector of `UInt128` indices that represent
 the distinct monomials of a `HomogeneousPolynomial` of order (degree) $i$.
 
     size_table :: Array{Int64,1}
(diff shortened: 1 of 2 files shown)


dpsanders:
This will slow down (I would guess by rather a lot) every single calculation with the package.
It will also have its own upper bound on the number of possible variables.

I think a better solution is the one I posted in https://github.com/JuliaDiff/TaylorSeries.jl/issues/85#issuecomment-280924935

although that will require more work to make everything type stable again.

lbenet:
> This will slow down (I would guess by rather a lot) every single calculation with the package.
It will also have its own upper bound on the number of possible variables.

Both statements are correct. Running `perf/fateman.jl` with Julia v0.5 yields almost a factor ~1.7 slower if #86 would be merged, though (interestingly) no more allocations. (Using `Int128` instead of `UInt128` yields essentially the same results.)

> I think a better solution is the one I posted in #85 (comment)
although that will require more work to make everything type stable again.

I also agree with this. Yet, this is quite tricky, since the choice must depend on the *values*. 

A possible approach could be to define a (global) parameter which fixes the parameterization of `generate_tables{T}` and `in_base{T}`, which eventually defines the type of `index_table`. Throwing an error message would also be nice, but I'm not sure how to do that.

lbenet:
@PerezHz The performance hit is important, and as @dpsanders states, it is not common to have situations where the used hashed numbers is so important. I suggest that you use the branch "index_table" to get access to what has been done here.

dpsanders:
I think we should move away from any kind of global table. Then each Taylor polynomial can store a reference to an object of the relevant type / size.

lbenet:
I agree that we should look for alternatives (including moving away from the global table scheme), but if such alternatives involve memory overhead, I don't think that's the way to go.

lbenet:
I closed this, `push -f` the corresponding commits adapted to the new organization of the files (#87), and keep this open, in case it may be needed.

PerezHz:
Thanks @lbenet and @dpsanders for the work you are putting into this; it is helping me a lot! Perhaps a note should be added in the documentation regarding the fact that `TaylorN` works with at most 65 variables and if more are needed then the current solution is to use the `index_table` branch?

lbenet:
Addressed by #314 and #316

PR #87, opened 2017-03-02
Reorganize files

lbenet:
This reorganizes the contents differently. In particular, it allows to exploit metaprogramming tools, though I have only done that to to a limited extent. My idea is to reorganize things and  provide a more consistent use of the internal methods.


dpsanders:
I started commenting but think it will be easier for me to do a PR with some code simplification once this is merged.

I certainly agree with splitting up the huge files into pieces. Is this all that is done in the PR? It would be best to make a separate PR for any actual code changes.

src/arithmetic.jl
```diff
@@ -0,0 +1,501 @@
...
+#
+
+# Arithmetic operations: +, -, *, /
+
+## Equality ##
+for T in (:Taylor1, :TaylorN)
+    @eval begin
+        =={T<:Number,S<:Number}(a::$T{T}, b::$T{S}) = ==(promote(a,b)...)
+        function =={T<:Number}(a::$T{T}, b::$T{T})
+            a, b = fixorder(a, b)
```
lbenet: `fixorder` is needed because you may be comparing two `Taylor1` (or `TaylorN`) polynomials with different order.

src/arithmetic.jl
```diff
@@ -0,0 +1,501 @@
...
+            @simd for i in eachindex(a.coeffs)
+                @inbounds (a.coeffs[i] == b.coeffs[i]) || return false
```
lbenet: Maybe, though I'm not sure if that would be faster or not.

src/arithmetic.jl
```diff
@@ -0,0 +1,501 @@
...
+        end
+    end
+end
+
+==(a::HomogeneousPolynomial, b::HomogeneousPolynomial) = a.coeffs == b.coeffs
+
+
+function iszero{T}(a::HomogeneousPolynomial{T})
+    @simd for i in eachindex(a.coeffs)
+        @inbounds (a.coeffs[i] == zero(T)) || return false
```
lbenet: Thanks for the advice! I will test it now.

src/arithmetic.jl
```diff
@@ -0,0 +1,501 @@
...
+    end
+    return true
+end
+
+
+## zero and one ##
+zero{T<:Number}(a::Taylor1{T}) = Taylor1(zero(a.coeffs[1]), a.order)
+one{T<:Number}(a::Taylor1{T}) = Taylor1(one(a.coeffs[1]), a.order)
```
lbenet: This is necessary for mixed types, e.g. `Taylor`{TaylorN{T}}`.

lbenet:
I just responded to some of your comments. I will test a replacement involving `all` that you point out somewhere.

Sorry about this PR, which includes some other changes beside reorganizing the files. Some things are related to use some metaprogramming to avoid repeating the same lines of code; other changes include SIMD optimizations, or using the new syntax for inner constructors.

Let me check your suggestion, and then it is a good idea to merge (otherwise, I'll continue putting things in top of this)...

src/arithmetic.jl
lbenet: I just implemented this; tests are failing ...

lbenet:
Regarding [this suggestion](https://github.com/JuliaDiff/TaylorSeries.jl/pull/87#discussion_r104237125), tests are failing in several places. So I will stop here making additions... Sorry, once again.

src/arithmetic.jl
lbenet: The idea of using `iszero` was for optimizing in places like `mul!`; without such optimizations, we loose a bit of performance in `perf/fateman.jl`, around ~15%.

lbenet: Sorry, I did something wrong regarding my check with your proposal; indeed tests pass. Regarding performance (`perf/fateman.jl`) it is slightly degraded (few percent), the number of allocations grows as well as the total memory used grows. So the current implementation, though not so nice, is a bit better.

lbenet:
@dpsanders One commit more 😄  : separate tests in several files and use Base.Test...

lbenet:
Indeed, `Number` includes `Real` and `Complex`, but also `Taylor1` and the other types defined in `TaylorSeries`. The reason to restrict it is to consider *only* those cases. 

(Things like `Taylor1{Taylor1{T}}` should still be possible, but I didn't check this. That actually allows to construct polynomials in many variables in a tree form, but it is something I haven't had time to explore more.)

src/arithmetic.jl
dpsanders:
```
julia v0.5> x, y = set_variables("x y")
  1.0 x + 𝒪(‖x‖⁷)
  1.0 y + 𝒪(‖x‖⁷)

julia v0.5> t = Taylor1(x, 10)
  1.0 x + 𝒪(‖x‖⁷) + 𝒪(t¹¹)

julia v0.5> one(eltype(t))
 1.0 + 𝒪(‖x‖¹)
```
`one(a.coeffs[1])` is the same as `one(typeof(a.coeffs[1]))` so my suggestion should always work.

src/arithmetic.jl
dpsanders: I would guess it would be the same speed and much less writing. You are just writing out by hand testing equality of two vectors as far as I can tell.

If you don't  use `fixorder` then testing equality would indeed be more complicated.

src/arithmetic.jl
dpsanders: But you are creating a whole new `Taylor1` object just to check equality, where you could just be checking that the coeffs that match are equal, and the extra coeffs in the longer vector are all 0.

lbenet: You are right; I'll see if I can do something to deal only with the coefficients.

lbenet:
Ok. I'll merge this and address the `fixorder` and the issue about `one` in a different PR. Thanks for the review!

Issue #89, opened 2017-03-07
Issues with mixtures of Taylor1 and TaylorN


lbenet:
After merging #87, and despite tests pass, @PerezHz made me note (offline) that certain functionality is no longer working as expected, such as `t * x` with `t::Taylor1{Float64}` and
`x::TaylorN{Taylor1{Float64}}`. This is to remind us about that.

PerezHz:
In [`src/arithmetic.jl`, line 179](https://github.com/JuliaDiff/TaylorSeries.jl/blob/master/src/arithmetic.jl#L179), `T` is being used with two meanings in the expression `b::$T{Taylor1{T}`!!

lbenet:
I think I've solved it; can you take a look in the branch "mixtures" ?

PerezHz:
Looks good to me! It seems to be working fine!

PerezHz:
Thanks for looking into this!

PerezHz:
Forget my comment about `b::$T{Taylor1{T}}`; there's no syntactic ambiguity there (but perhaps it's prone to confusion for a human reader to have `$T` and `T`? I'd suggest using a different letter e.g. `U` or whatever)

lbenet:
You are right, it leads to some confusion, but I checked that possibility too... I'd like to add some more tests before merging (and then I rebase the branch index_table).

lbenet:
This should be solved by #90. Closing

PR #90, opened 2017-03-07
Fix * and / for TaylorN{Taylor1{T}}

lbenet:
This solves some strange issue (#89) that appeared after reorganizing the files with some other minor changes.


lbenet:
Once I get green light here, I shall merge.

Issue #92, opened 2017-03-09
Division is not properly defined for some cases of Taylor1{TaylorN{Float64}}


PerezHz:
I'm working with `Taylor1{TaylorN{Float64}}` variables in Julia 0.5 using TaylorSeries latest master, but are having some trouble with two cases for the division:

+ Division of two `Taylor1{TaylorN{Float64}}`'s
+ Division of a `Taylor1{TaylorN{Float64}}` by a `Float64`

```julia
julia> using TaylorSeries

julia> x = 1.0 + TaylorN(Float64,1,order=5)
 1.0 + 1.0 x₁ + 𝒪(‖x‖⁶)

julia> y = Taylor1(x, 16)
  1.0 + 1.0 x₁ + 𝒪(‖x‖⁶) + 𝒪(t¹⁷)

julia> y/y # division of Taylor1{TaylorN{Float64}}'s
ERROR: MethodError: no method matching isinf(::TaylorSeries.TaylorN{Float64})
Closest candidates are:
  isinf(::BigFloat) at mpfr.jl:792
  isinf(::Float16) at float16.jl:118
  isinf(::Real) at float.jl:362
  ...
 in divfactorization(::TaylorSeries.Taylor1{TaylorSeries.TaylorN{Float64}}, ::TaylorSeries.Taylor1{TaylorSeries.TaylorN{Float64}}) at /Users/Jorge/.julia/v0.5/TaylorSeries/src/arithmetic.jl:430
 in /(::TaylorSeries.Taylor1{TaylorSeries.TaylorN{Float64}}, ::TaylorSeries.Taylor1{TaylorSeries.TaylorN{Float64}}) at /Users/Jorge/.julia/v0.5/TaylorSeries/src/arithmetic.jl:382

julia> y/1.0 # divide a Taylor1{TaylorN{Float64}} by a Float64
Segmentation fault: 11
```

It seems that in order to properly define division of two `Taylor1{TaylorN{Float64}}`'s, either a method `isinf(::TaylorSeries.TaylorN{Float64})` or `divfactorization{T<:Real}(a1::Taylor1{TaylorN{T}}, b1::Taylor1{TaylorN{T}})` must be defined. Also, in order to define the division of a `Taylor1{TaylorN{Float64}}` by a `Float64`, a specialized method `/{T<:Real}(a::Taylor1{TaylorN{T}}, b::T)` must be defined. Currently, what I'm doing to work around this is:

```julia
[...]

PR #93, opened 2017-03-09
Extend isinf and isnan to TaylorN and HomogeneousPolynomial

lbenet:
This solves #92


lbenet:
The problem in #92 is solved; yet there are issues which are not yet producing the expected results; I'll merge this and try to solve that separately.

Issue #94, opened 2017-03-09
Problems with promotion of mixed-types


lbenet:
As reported in #92, arithmetic operations are subtle when dealing with `Taylor1{TaylorN{T}}`. The simplest case I've come to, which yields to a segmentation fault is:
```julia
julia> using TaylorSeries

julia> x = 1.0 + TaylorN(Float64,1,order=5)
 1.0 + 1.0 x₁ + 𝒪(‖x‖⁶)

julia> y = Taylor1(x, 16)
  1.0 + 1.0 x₁ + 𝒪(‖x‖⁶) + 𝒪(t¹⁷)

julia> x + y
```

The problem seems to be related with promotion.

dpsanders:
What is the result of that operation expected to be? What if the `TaylorN` had two variables? Should this even be defined?

lbenet:
To make myself clearer, reconsider what happens above, but in the following form:

```julia
julia> using TaylorSeries

julia> δx, δy = set_variables("δx δy")
2-element Array{TaylorSeries.TaylorN{Float64},1}:
  1.0 δx + 𝒪(‖x‖⁷)
  1.0 δy + 𝒪(‖x‖⁷)

julia> x = 1+Taylor1(δx,10)
  1.0 + 1.0 δx + 𝒪(‖x‖⁷) + 𝒪(t¹¹)

```

So, in a way, `x` is time dependent (`Taylor1`), and `δx` represent possible variations close to 1. (There is also a δy, so the `TaylorN` part is actually a 2-variable polynomial.) So altogether, it is a polynomial in 3 variables. So my example would be to consider `x + δx`, which should result in `1.0 + 2.0 δx + 𝒪(‖x‖⁷) + 𝒪(t¹¹)`.

The idea of having this type of things is that you integrate the time dependence, and are able to say things about small deviations close to the reference trajectory.

PerezHz:
Using the latest commit of `fuse_operations` branch I noticed there seems to be still some trouble with `TaylorN{Taylor1{Float64}}`:

```julia
julia> using TaylorSeries

julia> p = set_variables("ξ", numvars=4, order=1);

julia> x1N = Taylor1([2+TaylorN(Float64,1,order=1),1+TaylorN(Float64,2,order=1)],10)
[...]

Issue #78, opened 2016-11-16
the order of importing TalyorSeries should not matter


JeffreySarnoff:
[the second package to find order dependance](https://github.com/MikeInnes/MacroTools.jl/issues/22)

I had the same problem many months ago.  Seeing Mike's note allowed me to inform you of that.

lbenet:
Thanks for the note Jeffrey! Can you please be more specific about Mike's notes? I guess you refer to @MikeInnes, don't you?

JeffreySarnoff:
Mike is @MikeInnes, instead of "note" read "issue" -- there is no other relevant note.

fhenneke:
I have a similar problem with TaylorSeries and ForwardDiff.
```julia
using TaylorSeries, ForwardDiff
a = ForwardDiff.Dual(1.)
a * a
```
works as intended, whereas
```julia
using ForwardDiff, TaylorSeries
a = ForwardDiff.Dual(1.)
a * a
```
produces an error.
```julia
ERROR: * not defined for ForwardDiff.Dual{0,Float64}
 in *(::ForwardDiff.Dual{0,Float64}, ::ForwardDiff.Dual{0,Float64}) at ./promotion.jl:256
```

lbenet:
Thanks for reporting. I truly do not know what's going on here.

JeffreySarnoff:
@lbenet ask the dev-list?

lbenet:
Closing since it is (somehow) fixed in #95.

Issue #62, opened 2016-07-21
code should not require three argument multiply


JeffreySarnoff:
in Taylor1 there is this line

``` julia
        coefhomog += (kcoef-i) * ac[kcoef-i+1] * coeffs[i+1]
```

which looks for *(::Int64, :::Float64, ::Float64)
It would be more portable to supply the parens

JeffreySarnoff:
hmm --- looks like vectors -- never mind

JeffreySarnoff:
reopening this because something is awkward -- deserves a look

lbenet:
Thanks for taking a look on this!

This kind of expressions appear often when functions of polynomials are computed. For concreteness, I'll assume that you are looking into `TaylorSeries.expHomogCoef`, and I will use the corresponding formula given in the [documentation](http://www.juliadiff.org/TaylorSeries.jl/latest/api/#TaylorSeries.expHomogCoef). Note that this formula provides a *recursive* relation to compute the coefficients of `c=exp(a)`, with `a` a polynomial in one variable.

`Taylor1{T}` objects are stored as a vector of `T`-coefficients, and the order (maximum degree) of the polynomial. So, indeed, `ac` and `coeffs` seem to be related with the  coefficients of `a` (the polynomial) and the already computed coefficients of `c = exp(a)`, in the example I am considering.

> reopening this because something is awkward -- deserves a look 

Can you be more specific in what you find as awkward? (I agree that something is awkward  deserves a look).

JeffreySarnoff:
```julia
using TaylorSeries
using ArbFloats
setprecision(ArbFloat,28);

t1=Taylor1([ArbFloat(2), ArbFloat(3)])
# 2.00000000 + 3.00000000 t + 𝒪(t²)
evaluate(t1,ArbFloat(0.5))
# 3.50000000
#   that is good
quit()
```
```julia
using ArbFloats
using TaylorSeries
setprecision(ArbFloat,28);

 t1=Taylor1([ArbFloat(2), ArbFloat(3)])
# 2.00000000 + 3.00000000 t + 𝒪(t²)
evaluate(t1,ArbFloat(0.5))
ERROR: * not defined for ArbFloats.ArbFloat{28}    
 in evaluate(::TaylorSeries.Taylor1{ArbFloats.ArbFloat{28}}, ::ArbFloats.ArbFloat{28}) at      
[...]

PR #95, opened 2017-03-10
Fuse operations and methods for mixed types

lbenet:
This PR began as an attempt to get rid of the use `fixorder`, which creates copies of Taylor1 objects. This has been accomplished to a certain extent using dot-operatotions and dot-functions. But I couldn't get rid of some cases. 

While working on this, #92 appeared followed by #94. I think I have solved this now by introducing new methods, and restricting more carefully some methods according to their methods.

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -468,10 +468,9 @@ function divfactorization(a1::Taylor1, b1::Taylor1)
     # order of first factorized term; a1 and b1 assumed to be of the same order
     a1nz = findfirst(a1)
     b1nz = findfirst(b1)
+    a1nz = a1nz ≥ 0 ? a1nz : a1.order
+    b1nz = b1nz ≥ 0 ? b1nz : a1.order
     orddivfact = min(a1nz, b1nz)
-    if orddivfact < 0
-        orddivfact = a1.order
-    end
     cdivfact = a1.coeffs[orddivfact+1] / b1.coeffs[orddivfact+1]
 
     # Is the polynomial factorizable?
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -41,9 +41,7 @@ one{T<:Number}(a::Taylor1{T}) = Taylor1(one(a.coeffs[1]), a.order)
 function zero{T<:Number}(a::HomogeneousPolynomial{T})
     a.order == 0 && return HomogeneousPolynomial([zero(a.coeffs[1])], 0)
     v = Array{T}( size_table[a.order+1] )
-    @simd for ind in eachindex(a.coeffs)
-        @inbounds v[ind] = zero(a.coeffs[ind])
-    end
+    v .= zero.(a.coeffs)
     return HomogeneousPolynomial(v, a.order)
 end
 
@@ -62,9 +60,7 @@ zeros{T<:Number}(::Type{HomogeneousPolynomial{T}}, order::Int) =
 function one{T<:Number}(a::HomogeneousPolynomial{T})
     a.order == 0 && return HomogeneousPolynomial([one(a.coeffs[1])], 0)
     v = Array{T}( size_table[a.order+1] )
-    @simd for ind in eachindex(a.coeffs)
-        @inbounds v[ind] = one(a.coeffs[ind])
-    end
+    v .= one.(a.coeffs)
     return HomogeneousPolynomial{T}(v, a.order)
 end
 
@@ -85,62 +81,78 @@ end
 ones{T<:Number}(::Type{HomogeneousPolynomial{T}}, order::Int) =
     ones( HomogeneousPolynomial([one(T)], 0), order)
 
-zero{T<:Number}(a::TaylorN{T}) = TaylorN(zero(T), a.order)
-one{T<:Number}(a::TaylorN{T}) = TaylorN(one(T), a.order)
+zero{T<:Number}(a::TaylorN{T}) = TaylorN(zero(a.coeffs[1]), a.order)
+one{T<:Number}(a::TaylorN{T}) = TaylorN(one(a.coeffs[1]), a.order)
 
 
 
 ## Addition and substraction ##
 for f in (:+, :-)
-    for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
+    dotf = ifelse( f == :+, :.+, :.-)
+    for T in (:Taylor1, :TaylorN)
         @eval begin
             ($f){T<:Number,S<:Number}(a::$T{T}, b::$T{S}) = $f(promote(a,b)...)
             function $f{T<:Number}(a::$T{T}, b::$T{T})
-                a, b = fixorder(a, b)
-                v = similar(a.coeffs)
-                @simd for i in eachindex(v)
-                    @inbounds v[i] = $f(a.coeffs[i], b.coeffs[i])
+                aord = a.order
+                bord = b.order
+                ac = view(a.coeffs, :)
+                bc = view(b.coeffs, :)
+                if aord == bord
+                    v = similar(ac)
+                    v .= $dotf(ac, bc)
+                elseif aord < bord
+                    v = similar(bc)
+                    @inbounds v[1:aord+1] .= $dotf(ac[1:aord+1], bc[1:aord+1])
+                    @inbounds v[aord+2:bord+1] .= $f(bc[aord+2:bord+1])
+                else
+                    v = similar(ac)
+                    @inbounds v[1:bord+1] .= $dotf(ac[1:bord+1], bc[1:bord+1])
+                    @inbounds v[bord+2:aord+1] .= ac[bord+2:aord+1]
                 end
-                return $T(v, a.order)
+                return $T(v)
             end
             function $f(a::$T)
                 v = similar(a.coeffs)
-                @simd for i in eachindex(v)
-                    @inbounds v[i] = $f(a.coeffs[i])
-                end
+                broadcast!($f, v, a.coeffs)
                 return $T(v, a.order)
             end
-            if $T != HomogeneousPolynomial
-                function $f(a::$T, b::RealOrComplex)
-                    R = promote_type(eltype(a.coeffs), eltype(b))
-                    coeffs = Array{R}(a.order+1)
-                    @simd for i in eachindex(coeffs)
-                        @inbounds coeffs[i] = a.coeffs[i]
-                    end
-                    @inbounds coeffs[1] = $f(a.coeffs[1], b)
-                    return $T(coeffs, a.order)
-                end
-                function $f(b::RealOrComplex, a::$T)
-                    R = promote_type(eltype(a.coeffs), eltype(b))
-                    coeffs = Array{R}(a.order+1)
-                    @simd for i in eachindex(coeffs)
-                        @inbounds coeffs[i] = $f(a.coeffs[i])
-                    end
-                    @inbounds coeffs[1] = $f(b, a.coeffs[1])
-                    return $T(coeffs, a.order)
-                end
+            function $f(a::$T, b::RealOrComplex)
+                R = promote_type(eltype(a.coeffs), eltype(b))
+                coeffs = Array{R}(a.order+1)
+                coeffs .= a.coeffs
+                @inbounds coeffs[1] = $f(a.coeffs[1], b)
+                return $T(coeffs, a.order)
+            end
+            function $f(b::RealOrComplex, a::$T)
+                R = promote_type(eltype(a.coeffs), eltype(b))
+                coeffs = Array{R}(a.order+1)
+                coeffs .= $f(a.coeffs)
+                @inbounds coeffs[1] = $f(b, a.coeffs[1])
+                return $T(coeffs, a.order)
             end
         end
     end
     @eval begin
+        ($f){T<:Number,S<:Number}(a::HomogeneousPolynomial{T},
+            b::HomogeneousPolynomial{S}) = $f(promote(a,b)...)
+        function $f{T<:Number}(a::HomogeneousPolynomial{T},
+                b::HomogeneousPolynomial{T})
+            @assert a.order == b.order
+            v = similar(a.coeffs)
+            v .= $dotf(a.coeffs, b.coeffs)
+            return HomogeneousPolynomial(v, a.order)
+        end
+        function $f(a::HomogeneousPolynomial)
+            v = similar(a.coeffs)
+            v .= $f(a.coeffs)
+            return HomogeneousPolynomial(v, a.order)
+        end
         function ($f){T<:RealOrComplex,S<:RealOrComplex}(
                 a::TaylorN{Taylor1{T}}, b::Taylor1{S})
             @inbounds aux = $f(a.coeffs[1].coeffs[1], b)
             R = typeof(aux)
-            coeffs = Array{HomogeneousPolynomial{Taylor1{R}}}(length(a.coeffs))
-            @simd for i in eachindex(coeffs)
-                @inbounds coeffs[i] = a.coeffs[i]
-            end
+            coeffs = Array{HomogeneousPolynomial{Taylor1{R}}}(a.order+1)
+            coeffs .= a.coeffs
             @inbounds coeffs[1] = aux
             return TaylorN(coeffs, a.order)
         end
@@ -148,10 +160,8 @@ for f in (:+, :-)
             b::Taylor1{S}, a::TaylorN{Taylor1{T}})
             @inbounds aux = $f(b, a.coeffs[1].coeffs[1])
             R = typeof(aux)
-            coeffs = Array{HomogeneousPolynomial{Taylor1{R}}}(length(a.coeffs))
-            @simd for i in eachindex(coeffs)
-                @inbounds coeffs[i] = $f(a.coeffs[i])
-            end
+            coeffs = Array{HomogeneousPolynomial{Taylor1{R}}}(a.order+1)
+            coeffs .= $f(a.coeffs)
             @inbounds coeffs[1] = aux
             return TaylorN(coeffs, a.order)
         end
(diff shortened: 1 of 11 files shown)


lbenet:
Pushed another commit to increase coverage

lbenet:
To my surprise (at least in Julia 0.5), this PR seems to solve the cases described in #62 and #78 related to the order of loading `TaylorSeries` and another package. 

Here is an example:
```julia
julia> using TaylorSeries, ArbFloats

julia> t1=Taylor1([ArbFloat(2), ArbFloat(3)])
 2 + 3 t + 𝒪(t²)

julia> exp(t1)
 7.389056098930650227230427 + 22.16716829679195068169128 t + 𝒪(t²)
```

and the other way around

```julia
julia> using ArbFloats, TaylorSeries

julia> t1=Taylor1([ArbFloat(2), ArbFloat(3)])
 2 + 3 t + 𝒪(t²)

julia> exp(t1)
 7.389056098930650227230427 + 22.16716829679195068169128 t + 𝒪(t²)

```

@PerezHz @JeffreySarnoff @fhenneke Can you confirm this indeed the case?

PerezHz:
Wow, quite unexpected! On my machine, the order of loading seems to be fixed for ArbFloats, ForwardDiff, and ValidatedNumerics!

lbenet:
@dpsanders Do you want to take a look on this PR? (Sorry for making it so complicate.) Tests pass and have been enlarged to cover other situations (#94), including unexpected ones (#62 and #78).

dpsanders:
Yes, but I won't be able to take a proper look at it until Sunday.
If you prefer to merge now then go ahead.

lbenet:
Take your time. You have always good comments and insight! Thanks!

dpsanders:
Hmm, I see that not using `fixorder` does complicate the code quite a lot.

It's also a bit unfortunate that what was now relatively simple code for `Taylor1` is mixed up with more complicated code for `TaylorN`?

lbenet:
I agree, though your idea in #96 may help a little bit...

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
-one{T<:Number}(a::TaylorN{T}) = TaylorN(one(T), a.order)
+zero(a::TaylorN) = TaylorN(zero(a.coeffs[1]), a.order)
+one(a::TaylorN) = TaylorN(one(a.coeffs[1]), a.order)
 
 
 
 ## Addition and substraction ##
 for f in (:+, :-)
-    for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
+    dotf = ifelse( f == :+, :(.+), :(.-))
```
dpsanders: `dotf = Symbol(".", f)`

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
         @eval begin
             ($f){T<:Number,S<:Number}(a::$T{T}, b::$T{S}) = $f(promote(a,b)...)
             function $f{T<:Number}(a::$T{T}, b::$T{T})
-                a, b = fixorder(a, b)
-                v = similar(a.coeffs)
-                @simd for i in eachindex(v)
-                    @inbounds v[i] = $f(a.coeffs[i], b.coeffs[i])
+                aord = a.order
+                bord = b.order
+                ac = view(a.coeffs, :)
```
dpsanders: Why is this needed?

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
+    dotf = ifelse( f == :+, :(.+), :(.-))
+    for T in (:Taylor1, :TaylorN)
         @eval begin
             ($f){T<:Number,S<:Number}(a::$T{T}, b::$T{S}) = $f(promote(a,b)...)
             function $f{T<:Number}(a::$T{T}, b::$T{T})
-                a, b = fixorder(a, b)
-                v = similar(a.coeffs)
-                @simd for i in eachindex(v)
-                    @inbounds v[i] = $f(a.coeffs[i], b.coeffs[i])
+                aord = a.order
```
dpsanders: I would suggest using `la = length(a)` again.

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
+                bord = b.order
+                ac = view(a.coeffs, :)
+                bc = view(b.coeffs, :)
+                if aord == bord
+                    v = similar(ac)
+                    v .= $dotf(ac, bc)
+                elseif aord < bord
+                    v = similar(bc)
+                    @inbounds v[1:aord+1] .= $dotf(ac[1:aord+1], bc[1:aord+1])
+                    @inbounds v[aord+2:bord+1] .= $f(bc[aord+2:bord+1])
```
dpsanders: Use `end` instead of `bord+1`.

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
-                    @simd for i in eachindex(coeffs)
-                        @inbounds coeffs[i] = $f(a.coeffs[i])
-                    end
-                    @inbounds coeffs[1] = $f(b, a.coeffs[1])
-                    return $T(coeffs, a.order)
-                end
+            ($f){T<:Number,S<:Number}(a::$T{T}, b::S) = $f(promote(a,b)...)
+            function $f{T<:Number}(a::$T{T}, b::T)
+                R = eltype(a.coeffs)
+                coeffs = Array{R}(a.order+1)
```
dpsanders: Aren't these two lines just `coeffs = similar(a, a.order+1)`?

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
+                coeffs .= a.coeffs
+                @inbounds coeffs[1] = $f(a.coeffs[1], b)
+                return $T(coeffs, a.order)
+            end
+            ($f){T<:Number,S<:Number}(b::S, a::$T{T}) = $f(promote(b,a)...)
+            function $f{T<:Number}(b::T, a::$T{T})
```
dpsanders: Please add a new line before each new function definition.

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
-                end
+            ($f){T<:Number,S<:Number}(a::$T{T}, b::S) = $f(promote(a,b)...)
+            function $f{T<:Number}(a::$T{T}, b::T)
+                R = eltype(a.coeffs)
+                coeffs = Array{R}(a.order+1)
+                coeffs .= a.coeffs
+                @inbounds coeffs[1] = $f(a.coeffs[1], b)
+                return $T(coeffs, a.order)
+            end
+            ($f){T<:Number,S<:Number}(b::S, a::$T{T}) = $f(promote(b,a)...)
```
dpsanders: Shouldn't all these promotions be automatic for standard function like `+`?

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
+        function $f{T<:NumberNotSeriesN}(a::HomogeneousPolynomial{T},
+                b::HomogeneousPolynomial{T})
+            @assert a.order == b.order
+            v = similar(a.coeffs)
+            v .= $dotf(a.coeffs, b.coeffs)
+            return HomogeneousPolynomial(v, a.order)
+        end
+        function $f(a::HomogeneousPolynomial)
+            v = similar(a.coeffs)
+            v .= $f(a.coeffs)
```
dpsanders: These two lines can be reduced to `v = $f(a.coeffs)`.

src/arithmetic.jl
```diff
@@ -73,88 +70,119 @@ function ones{T<:Number}(a::HomogeneousPolynomial{T}, order::Int)
...
                 a::TaylorN{Taylor1{T}}, b::Taylor1{S})
             @inbounds aux = $f(a.coeffs[1].coeffs[1], b)
-            R = typeof(aux)
-            coeffs = Array{HomogeneousPolynomial{Taylor1{R}}}(length(a.coeffs))
-            @simd for i in eachindex(coeffs)
-                @inbounds coeffs[i] = a.coeffs[i]
-            end
+            R = eltype(aux)
+            coeffs = Array{HomogeneousPolynomial{Taylor1{R}}}(a.order+1)
+            coeffs .= a.coeffs
```
dpsanders: These two lines should be equivalent to a `convert`?

src/arithmetic.jl
```diff
@@ -165,37 +193,39 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
...
+    end
+end
+
+for T in (:HomogeneousPolynomial, :TaylorN)
+    @eval begin
+        function *{T<:NumberNotSeries,S<:NumberNotSeries}(a::Taylor1{T}, b::$T{Taylor1{S}})
+            @inbounds aux = a * b.coeffs[1]
+            R = typeof(aux)
+            coeffs = Array{R}(length(b.coeffs))
+            coeffs .= a .* b.coeffs
```
dpsanders: It feels like these three lines form a recurring pattern that should be factored out into a function somehow.

src/arithmetic.jl
```diff
@@ -165,37 +193,39 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
...
+            return $T(coeffs, b.order)
         end
+        *{T<:NumberNotSeries,R<:NumberNotSeries}(b::$T{Taylor1{R}}, a::Taylor1{T}) = a * b
+        function *{T<:NumberNotSeries,S<:NumberNotSeries}(a::$T{T}, b::Taylor1{$T{S}})
+            @inbounds aux = a * b.coeffs[1]
+            R = typeof(aux)
+            coeffs = Array{R}(length(b.coeffs))
+            coeffs .= a .* b.coeffs
```
dpsanders: There is a lot of repetition here with the previous definitions.

src/arithmetic.jl
```diff
@@ -262,7 +292,7 @@ Inputs are the `kcoef`-th coefficient, and the vectors of the expansion coeffici
 """
 function mulHomogCoef{T<:Number}(kcoef::Int, ac::Array{T,1}, bc::Array{T,1})
     kcoef == 0 && return ac[1] * bc[1]
-    coefhomog = zero(T)
+    coefhomog = zero(ac[1])
```
dpsanders: `zero(T)`.

src/auxiliary.jl
```diff
@@ -8,27 +8,35 @@
 
 ## Auxiliary function ##
 
-function check_taylor1_order!{T<:Number}(coeffs::Array{T,1}, order::Int)
+"""
+    resize_coeffs1!{T<Number}(coeffs::Array{T,1}, order::Int)
```
lbenet: I moved part of the constructor of `Taylor1` to this function to simplify the code. I'm not sure how to answer the question, since this function resizes the coefficients if you ask for a larger order. For instannce, you may call `Taylor1([0,1],10)` and at some point this function is called to resize the coefficients to size 11.

dpsanders: My point is that there will no longer be any need to resize the coefficients, ever!

src/constructors.jl
```diff
@@ -179,3 +182,11 @@ julia> TaylorN(Rational{Int},2)
 TaylorN{T<:Number}(::Type{T}, nv::Int; order::Int=get_order()) =
     return TaylorN( [HomogeneousPolynomial(T, nv)], order )
 TaylorN(nv::Int; order::Int=get_order()) = TaylorN(Float64, nv, order=order)
+
+
+# A `Number` which is not an `AbstractSeries`
+const NumberNotSeries = Union{setdiff(subtypes(Number), [AbstractSeries])...}
```
lbenet: Somehow beyond my understanding, I think this is at the core why problems reported in #62 and #94 are solved.... or Julia-magic.

src/auxiliary.jl
```diff
@@ -8,27 +8,35 @@
...
     lencoef = length(coeffs)
-    order = max(order, lencoef-1)
-    order == lencoef-1 && return nothing
+    order ≤ lencoef-1 && return nothing
     resize!(coeffs, order+1)
-    @simd for i = lencoef+1:order+1
-        @inbounds coeffs[i] = zero(coeffs[1])
-    end
-    nothing
+    coeffs[lencoef+1:order+1] .= zero(coeffs[1])
```
lbenet: I guess it should work... I'll push another commit with this change.

src/auxiliary.jl
lbenet: I truly hope so!

src/evaluate.jl
```diff
@@ -96,43 +96,42 @@ end
...
-
-    numVars = get_numvars()
-    @assert length(vals) == numVars
+# function evaluate{T<:Number}(a::HomogeneousPolynomial{T}, vals::Array{T,1} )
+function evaluate{T<:Number,S<:NumberNotSeriesN}(a::HomogeneousPolynomial{T},
+        vals::Array{S,1} )
+    @assert length(vals) == get_numvars()
     R = promote_type(T,S)
+    numVars = get_numvars()
     suma = convert(TaylorN{R}, a)
```
lbenet: Type stability; if you simply use assignment (`suma = a`) it may change type inside the loop.

src/evaluate.jl
```diff
@@ -96,43 +96,42 @@ end
...
 
-    for nv = 1:numVars
+    @inbounds for nv = 1:numVars
         suma = horner(suma, (nv, vals[nv]))
     end
 
     return suma.coeffs[1].coeffs[1]
```
lbenet: Maybe. The way it is done here is that a `TaylorN` polynomial is evaluated substituting each variable, one by one; maybe it's not the best way of doing it...

src/evaluate.jl
```diff
@@ -96,43 +96,42 @@ end
...
 end
 
-evaluate{T<:Number}(a::TaylorN{T}) = evaluate(a, zeros(T, get_numvars()))
+evaluate{T<:Number}(a::TaylorN{T}) = a.coeffs[1].coeffs[1]
```
lbenet: Evaluating a polynomial on zero is equivalent to the independent order zero) coefficient.

src/arithmetic.jl
```diff
@@ -13,37 +13,36 @@ for T in (:Taylor1, :TaylorN)
     @eval begin
         =={T<:Number,S<:Number}(a::$T{T}, b::$T{S}) = ==(promote(a,b)...)
         function =={T<:Number}(a::$T{T}, b::$T{T})
-            a, b = fixorder(a, b)
-            @simd for i in eachindex(a.coeffs)
-                @inbounds (a.coeffs[i] == b.coeffs[i]) || return false
+            if a.order == b.order
+                return all( a.coeffs .== b.coeffs )
+            elseif a.order < b.order
+                res1 = all( a.coeffs[1:a.order+1] .== b.coeffs[1:a.order+1] )
```
lbenet: For `Taylor1` this is essentially equivalent; for `TaylorN` this is a bit more involved. In any case, it's clearer the way you present it; I'll implement it.

src/arithmetic.jl
```diff
@@ -13,37 +13,36 @@ for T in (:Taylor1, :TaylorN)
...
-    return true
+for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
+    @eval iszero(a::$T) = iszero(a.coeffs)
 end
 
-
 ## zero and one ##
-zero{T<:Number}(a::Taylor1{T}) = Taylor1(zero(a.coeffs[1]), a.order)
-one{T<:Number}(a::Taylor1{T}) = Taylor1(one(a.coeffs[1]), a.order)
+zero(a::Taylor1) = Taylor1(zero(a.coeffs[1]), a.order)
```
lbenet: No. It's subtle, I agree, but things like `Taylor1{TaylorN{T}}` don't work correctly, because `zero(::Type{TaylorN{T}})` has no way of knowing about the order of the polynomial.

src/arithmetic.jl
lbenet: I'll implement it like that. Thanks!

src/arithmetic.jl
lbenet: I'll do it.

src/arithmetic.jl
lbenet: Not really needed, but the expressions were getting too long. Maybe with `la` things are easier to read.

src/arithmetic.jl
lbenet: I will. Thanks.

src/arithmetic.jl
lbenet: I think so, but there were subtleties related with the mixed types.

src/arithmetic.jl
lbenet: For `+` and `-` it's worth the code, since promotion would add lots of coefficients to zero.

src/arithmetic.jl
lbenet: I will! Sorry...

src/arithmetic.jl
lbenet: I guess so... Thanks for the catch!

src/arithmetic.jl
lbenet: I guess they appear like that to assure type stability.

src/arithmetic.jl
lbenet: Same as some comment above; the difference appears for `Taylor1{TaylorN{T}}`...

lbenet:
I'll revise the code and push another commit.

lbenet:
@dpsanders Thanks for your review! And, again, I'm sorry that this PR became so extended and messy.

I just pushed another commit which address many of your observations, but not all. As you suggest, this can be part of another commit. Once we get green lights, I think this should be merged.

Somewhere, which i can't find now, you also suggested that the tests should reflect more the file that they are actually testing. I will reorganize them as well, but in another PR.

lbenet:
Got green lights. Merging.

dpsanders:
Nice!

PR #97, opened 2017-03-15
Add getindex; replace a.coeffs[i] with a[i] 

dpsanders:
- Replace all occurrences of `a.coeffs[i]` by `a[i]`.
- Add corresponding `getindex` and `setindex!` methods.


lbenet:
I just approved this to be merged, but it seems that tests (at least in 0.5) are not passing. I'll check this again.

src/auxiliary.jl
```diff
@@ -69,7 +69,12 @@ end
 Return the coefficient of order `n::Int` of a `a::Taylor1` polynomial.
 """
 get_coeff(a::Taylor1, n::Int) = (@assert 0 ≤ n ≤ a.order+1;
-    return a.coeffs[n+1])
+    return a[n+1])
+
+getindex(a::Taylor1, n::Int) = a.coeffs[n]
+getindex(a::Taylor1, n::UnitRange) = a.coeffs[n]
```
lbenet: Don't we need also methods for `HomogeneousPolynomial` and `TaylorN`? This may be the cause that the tests are not passing.

dpsanders: Yes we do. I wanted to begin just with Taylor1 to get it right. I won't be able to get to the rest for a while if you want to have a go.

lbenet: I'll do it then.

One question: Should we also include `setindex`?

dpsanders: Ah yes, I guess we should.

dpsanders:
setindex! 

Is it surprising that the Taylor1 tests pass without setindex defined?

lbenet:
I guess this is because we haven't done anything like `a[1] = b[1]+c[1]`, with all quantities `Taylor1`. This could simplify things even further! Yet, `Taylor1` and the rest are immutable, so I don't know if it works completely...

dpsanders:
But I changed all a.coeffs accesses to use getindex, so I'm not sure how
any entries in a TaylorSeries are getting changed...

dpsanders:
Ah I see, they're immutable so you only use the constructor. Then there's
no need or use to define setindex!?

lbenet:
Including `setindex!` the tests are passing. I'll include further tests.

lbenet:
I just pushed a new commit. With this functionality, I think we can simplify even further the code. I will work on that later (on a new PR)! 

I'll wait to get the green lights and then merge.

dpsanders:
Tests are passing on Travis for 0.5 and 0.6 (on Linux). 
The Mac queues seem to be eternal these days, so I suggest merging this.

lbenet:
I agree with you.

One silly question: shall I rebase and merge, or squash and merge?

dpsanders:
I always just use the Squash and merge button.

lbenet:
Thanks a lot!!

PR #98, opened 2017-03-17
0.6 update: type -> mutable struct

miguelraz:
I believe this should get this script up to speed with [version 0.6](https://github.com/JuliaLang/julia/blob/master/NEWS.md)


lbenet:
I'm closing this, simply because it is too early to implement such changes since we want to still support Julia 0.5; thanks anyway!

PR #64, opened 2016-09-14
Added hyperbolic trigonometric functions.

blas-ko:
Added functions `sinh`, `cosh` , `tanh` to `Taylor1.jl` alongside with some tests. 

*NOTE:* For proving that  ` -sinh(t)^2 + cosh(t)^2 == 1`  a random number `rand()` is evaluated on the tests. Is this legit enough?

PR #49, opened 2016-05-22
Taylor1 functions

blas-ko:
Inverse trigonometric functions `asin`, `acos`, `atan` were added for `Taylor1.jl`. 

Although the recurrence formula I used is correct, there are some problems in obtaining this expansions because an auxiliary expansion `r(x) = sqrt(1 - a^2)` is used for `asin` and `acos`. 

Notice that if `a.coeffs[1] = 1`, then `1 / r.coeffs[1] = Inf`, which is causing trouble when calculating the `HomogCoefs`.

`atan` works perfectly in the tests I did, where `atan(tan(a)) = a = tan(atan(a)) ` with `a::Taylor1`.

PR #99, opened 2017-03-19
Refactor functions

lbenet:
This PR aims to build a more consistent API, in the sense that a function `f` of `Taylor1` and functions of `TaylorN` should be treated as close as possible. Now, most of the functions have more or less the same structure (except maybe for the first coefficient), and call a related function which mutates the coefficients of the resulting function. The functions that mutate the coefficients (which are the recurrence relations on the coefficients) have nicer names and a more homogeneous structure across all function; note that not all functions have been implemented in this way. By doing this, we have extended  with (essentially) no effort the inverse trigonometric functions #49 and the hyperbolic functions #64 to `TaylorN`.

I have also deleted some documentation of functions (e.g., `exp(a)` for `a::Taylor1` or `a::TaylorN`), which do not really provided anything new but the extension to these types.

There are still some tests to be added, and the example notebooks should also be updated.


src/arithmetic.jl
```diff
@@ -250,93 +248,85 @@ for T in (:HomogeneousPolynomial, :TaylorN)
     end
 end
 
+for (T, W) in ((:Taylor1, :Number), (:TaylorN, :NumberNotSeriesN))
+    @eval *{T<:$W, S<:$W}(a::$T{T}, b::$T{S}) = *(promote(a,b)...)
 
+    @eval function *{T<:$W}(a::$T{T}, b::$T{T})
+        corder = max(a.order, b.order)
+        if $T == Taylor1
+            c = Taylor1(zero(a[1]), corder)
```
dpsanders: Just add a method `zero(a, order)` to make an object of that order filled with zeros, then this should work here and in the next line.

src/arithmetic.jl
```diff
@@ -250,93 +248,85 @@ for T in (:HomogeneousPolynomial, :TaylorN)
...
-"""
-*{T<:Number,S<:Number}(a::Taylor1{T}, b::Taylor1{S}) = *(promote(a,b)...)
-
-function *{T<:Number}(a::Taylor1{T}, b::Taylor1{T})
-    a, b = fixorder(a, b)
-    coeffs = similar(a.coeffs)
-    @inbounds for k = 0:a.order
-        coeffs[k+1] = mulHomogCoef(k, a.coeffs, b.coeffs)
+        @inbounds for ord = 0:corder
+            mul!(c, a, b, ord) # updates c[ord+1]
```
dpsanders: Is there not a nicer way of writing this?

src/arithmetic.jl
```diff
@@ -250,93 +248,85 @@ for T in (:HomogeneousPolynomial, :TaylorN)
...
-## Multiplication ##
-*{T<:NumberNotSeriesN,S<:NumberNotSeriesN}(a::HomogeneousPolynomial{T}, b::HomogeneousPolynomial{S}) =
-    *(promote(a,b)...)
-
-function *{T<:NumberNotSeriesN}(a::HomogeneousPolynomial{T}, b::HomogeneousPolynomial{T})
     order = a.order + b.order
 
-    order > get_order() && return HomogeneousPolynomial([zero(a[1])], get_order())
+    order > get_order() &&
+        return HomogeneousPolynomial([zero(a[1])], get_order())
```
dpsanders: This line and the next can be combined with something like  `min(order, get_order())`.
Isn't there a better way of writing this zero?

src/arithmetic.jl
```diff
@@ -407,89 +365,70 @@ function /{T<:Integer, S<:NumberNotSeries}(a::Taylor1{Rational{T}}, b::S)
...
-        /{T<:NumberNotSeries}(a::$T{T}, b::T) = a * inv(b)
-        function /{T<:NumberNotSeries,S<:NumberNotSeries}(a::$T{T}, b::S)
-            R = promote_type(T,S)
-            return convert($T{R}, a) * inv(convert(R, b))
-        end
-        /{T<:NumberNotSeries}(a::$T, b::T) = a * inv(b)
+
+    @eval /{T<:NumberNotSeries}(a::$T{T}, b::T) = a * inv(b)
+
+    @eval function /{T<:NumberNotSeries,S<:NumberNotSeries}(a::$T{T}, b::S)
```
dpsanders: Any particular reason for splitting up the `@eval begin...end` into separate `@eval` here?

src/arithmetic.jl
```diff
@@ -407,89 +365,70 @@ function /{T<:Integer, S<:NumberNotSeries}(a::Taylor1{Rational{T}}, b::S)
...
+        R = promote_type(T,S)
+        return convert($T{R}, a) * inv(convert(R, b))
     end
+
+    @eval /{T<:NumberNotSeries}(a::$T, b::T) = a * inv(b)
```
dpsanders: Why is there this one and the one on line 369?

src/arithmetic.jl
```diff
@@ -407,89 +365,70 @@ function /{T<:Integer, S<:NumberNotSeries}(a::Taylor1{Rational{T}}, b::S)
...
+            b::$T{Taylor1{S}}, a::Taylor1{T})
+        @inbounds aux = b[1] / a
+        R = typeof(aux)
+        coeffs = Array{R}(length(b.coeffs))
+        coeffs .= b.coeffs ./ a
+        return $T(coeffs, b.order)
     end
-end
 
+    @eval function /{T<:NumberNotSeries,S<:NumberNotSeries}(
```
dpsanders: This is exactly the same code as the previous one. Can they be unified?

src/auxiliary.jl
```diff
@@ -71,7 +71,10 @@ Return the coefficient of order `n::Int` of a `a::Taylor1` polynomial.
 get_coeff(a::Taylor1, n::Int) = (@assert 0 ≤ n ≤ a.order+1;
     return a[n+1])
 
-getindex(a::Taylor1, n::Int) = a.coeffs[n]
+function getindex(a::Taylor1, n::Int)
+    (1 ≤ n ≤ length(a.coeffs)) && return a.coeffs[n]
```
dpsanders: This can now be written as
```
(1 <= n <= length(a)) && return a[n]
```

dpsanders: Ah no, sorry, that's begin defined here!

src/auxiliary.jl
```diff
@@ -111,10 +114,17 @@ function get_coeff(a::TaylorN, v::Array{Int,1})
     get_coeff(a[order+1], v)
 end
 
-getindex(a::TaylorN, n::Int) = a.coeffs[n]
+function getindex(a::TaylorN, n::Int)
+    1 ≤ n ≤ length(a.coeffs) && return a.coeffs[n]
+    n ≤ get_order()+1 &&
+        return zero( HomogeneousPolynomial([zero(eltype(a))], n-1) )
```
dpsanders: Again there needs to be a simpler way to write this `zero`.

src/functions.jl
```diff
@@ -6,649 +6,646 @@
...
-    coefhomog
-end
+# Functions
+for T in (:Taylor1, :TaylorN)
+    @eval begin
+        ## exp ##
+        function exp(a::$T)
+            if $T == Taylor1
+                order = a.order
+                @inbounds aux = exp( a[1] )
```
dpsanders: There should be a way to unify these. Otherwise it doesn't make much sense to have them mixed up.

dpsanders: e.g. Replace `get_order()` by `order(a)` and define `order(a)` for Taylor1 and TaylorN.

src/functions.jl
```diff
@@ -6,649 +6,646 @@
...
+                c = $T( aux, order )
+            else
+                order = get_order()
+                @inbounds aux = exp( a[1][1] )
```
dpsanders: What is `aux` for here?

dpsanders: Same comments presumably apply to all the following functions.

src/functions.jl
```diff
@@ -6,649 +6,646 @@
...
+        sin(a::$T) = sincos(a)[1]
+        cos(a::$T) = sincos(a)[2]
+        function sincos(a::$T)
+            if $T == Taylor1
+                order = a.order
+                @inbounds s = $T( sin(a[1]), order )
+                @inbounds c = $T( cos(a[1]), order )
+            else
+                order = get_order()
+                @inbounds s = $T( sin(a[1][1]), order )
```
dpsanders: How about defining a function `constant_term` (or similar name) that does this correctly for Taylor1 and TaylorN.

dpsanders: i.e.
```
constant_term(a::Taylor1) = a[1]
constant_term(a::TaylorN) = a[1][1]
```

src/functions.jl
```diff
@@ -6,649 +6,646 @@
...
-    @inbounds coscoeffs[1] = cos( a[1] )
-    @inbounds for k = 1:a.order
-        sincoeffs[k+1], coscoeffs[k+1] = sincosHomogCoef(k, v, sincoeffs, coscoeffs)
-    end
-    return Taylor1(sincoeffs, a.order), Taylor1(coscoeffs, a.order)
-end
+# Recursive functions (homogeneous coefficients)
+for T in (:Taylor1, :TaylorN)
+    @eval begin
+        function exp!(c::$T, a::$T, k::Int)
```
dpsanders: How does `exp!` differ from `exp`?
The usual trick is to define `exp!` correctly, and then `exp` is just `exp(x) = (y = copy(x); exp!(y)`

src/functions.jl
```diff
@@ -6,649 +6,646 @@
...
-        coeffs[k+1] = tanHomogCoef(k, v, coeffst2)
-        coeffst2[k+1] = squareHomogCoef(k, coeffs)
-    end
-    Taylor1( coeffs, a.order )
-end
-
-# Homogeneous coefficients for tan
 doc"""
-    tanHomogCoef(kcoef, ac, coeffst2)
+    reverse(f)
```
dpsanders: `inverse`?

dpsanders: Or can we overload `inv`?

src/other_functions.jl
```diff
@@ -6,30 +6,23 @@
...
-
+for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
 
-## real, imag, conj and ctranspose ##
-for TT in (:HomogeneousPolynomial, :TaylorN), f in (:real, :imag, :conj)
-    @eval ($f){T<:Number}(a::($TT){T}) = ($TT)(($f)(a.coeffs), a.order)
-end
+    ## real, imag, conj and ctranspose ##
+    for f in (:real, :imag, :conj)
+        fdot = Symbol(f, ".")
```
dpsanders: You don't seem to use `fdot`.

dpsanders:
Overall this looks like a very nice improvement, thanks!

I think it can be made even better following some of my comments, trying to unify *even more* the functions for `Taylor1` and `TaylorN`. I gave some concrete suggestions that I think will help achieve this to a large extent. 

Basically, every time there is nearly-duplicated code, it often just needs an extra couple of definitions of extra functions to take care of the small differences, and reduce it to a single function definition.

lbenet:
Did you submit your review observations? I haven't received anything and I am very much looking forward. In any case, thanks a lot for the feedback and the review!

I agree with you that there is indeed a pattern in the mutating functions, and that this should be exploited.

dpsanders:
Sorry, done. Not used to the new interface!

lbenet:
Thanks! I'll read it now!

src/arithmetic.jl
lbenet: Good point! I'll implement it. This should definitely get rid of some of these kind of `if`s.

src/arithmetic.jl
lbenet: Let me try to explain the "design principle". Essentially, we need the same mutating functions for the coefficients of either `Taylor1` or `TaylorN`, which must include "what to update" (`c`), the info to update it (`a` and `b` here because this is a binary function), and the order of the coefficient to be updated. So, as far as I can see, this is some sort of minimum required. For `mul!` I tried to be as close as what we already had for `HomogeneousPolynomial`.

Do you have some suggestion on this?

src/arithmetic.jl
lbenet: I'm not sure if I understand your point here. 

The case is the following: The multiplication of two `HomogeneousPolynomial` is another one whose order is the addition of the orders of the polynomials involved. Due to the fact that we have a maximum order for `HomogeneousPolynomial` (and `TaylorN`), whenever the resulting polynomial is of larger degree we return a zero of the largest degree. This is what is implemented here. We can certainly return zero with another order, but I'm not sure this helps at all...

src/arithmetic.jl
lbenet: Not really...

src/arithmetic.jl
dpsanders: I was thinking something like `zero(HomogeneousPolynomial{T}, get_order(a))`
for whatever is a suitable `T`. Maybe no better.

src/functions.jl
dpsanders: I think this is the key comment that will eliminate code duplication here!

src/arithmetic.jl
lbenet: Good catch! Indeed, line 369 can be deleted.

src/arithmetic.jl
lbenet: I think both functions are needed: the first function is e.g. for `\(b::TaylorN{Taylor1{T}}, a::Taylor1{S})`, while the second one is for `\(b::Taylor1{TaylorN{T}}, a::TaylorN{S})`. Note that the code in 376 does not include this case.

In any case, you have a point: the body of the functions are the same, while it is only the function definition what changes. I was simply too lazy to think on collapsing this further.

src/auxiliary.jl
lbenet: Maybe you are right. 

The subtlety is that for `a.order+1 < n <= get_order()` we can't rely on `a[n]` or `a.coeffs[n]`, since the length of `a.coeffs` is `a.order+1`.

src/functions.jl
lbenet: I guess it is enough to add methods for `get_order(a)` for `Taylor1`, `HomogeneousPolynomial` and `TaylorN`. I'll do that. If you really really want it, then we change the name.

dpsanders: Writing `order(a)` seems pretty natural to me.

dpsanders: But I guess we can leave that for the next PR.

lbenet: Actually, except for Taylor1, the methods were there...

lbenet: I just noticed that, here, I made a design decision. Suppose that the maximum order (for `TaylorN`) is 6. By what ever reason, you may have the (first) independent variable of order 2, `x=TaylorN(1, order=2)`; yet, `exp(x)` should be returned to maximum possible order (so `get_order()`, instead of `get_order(x) = x.order`...

dpsanders: That seems to be due to an asymmetry between how Taylor1 and TaylorN are conceived, i.e. the fact that there is no "global" order for Taylor1.

lbenet: I agree with you. A way out of this would be to force all `TaylorN` to be of order `get_order()`...

src/functions.jl
lbenet: You are right, this appears all over the place. `aux` is some sort of initial value which allows initializing the returned `Taylor1` or `TaylorN`, and also defines the initial value for the recurrence relations.

src/functions.jl
lbenet: Very nice! I'll do it!

src/functions.jl
lbenet: I have defined a function `max_order(a)` which for `a::Taylor1` returns `a.order` and for `a::TaylorN` returns `get_coeff()`. Probably not the nicest name, but that is left for another PR.

src/functions.jl
lbenet: `exp(a)` returns a polynomial of type `typeof(a)`. 

`exp!(c, a, k)` mutates `c[k+1]` with the k-th order coefficient of `c = exp(a)`. The recursion relations are actually coded in `exp!`.

dpsanders: Maybe it would be useful to use https://github.com/alsam/OffsetArrays.jl
to do away with all these annoying `+1` everywhere.

src/other_functions.jl
lbenet: Good catch! Indeed, that was somehow problematic, and at the end $f.(...) worked fine. I'll erase it.

src/functions.jl
lbenet: We should indeed consider it, but let's leave it (again) for another PR.

lbenet:
Thank you very much for the good suggestions and the whole review. 

I just pushed few commits addressing your observations. There are some I haven't yet implemented (`get_order` be replaced by `order`), but I got rid of lots of code duplication and, actually, of the nasty if-block inside some `@eval` blocks. I introduced a couple of new functions (whose name we may indeed change).

Lots of tests are still to be implemented, but this is left for tomorrow. 

Once again, thanks a lot for the review and ideas!

src/functions.jl
```diff
@@ -6,649 +6,479 @@
...
-    @inbounds for i = 0:kcoef-1
-        coefhomog += (kcoef-i) * ac[kcoef-i+1] * coeffs[i+1]
-    end
-    coefhomog = coefhomog/kcoef
-    coefhomog
-end
+# Functions
+for T in (:Taylor1, :TaylorN)
+    @eval begin
+        ## exp ##
```
dpsanders: These comments can now be removed ;)

src/functions.jl
```diff
@@ -6,649 +6,479 @@
...
+        function exp(a::$T)
+            order = max_order(a)
+            @inbounds aux = exp( constant_term(a) )
```
dpsanders: The `@inbounds` can be moved inside the `constant_term` function.

src/functions.jl
```diff
@@ -6,649 +6,479 @@
...
+            c = $T( aux, order )
```
dpsanders: Maybe get rid of `aux` everywhere and just write
```
c = $T( exp(constant_term(a)), order )
```
Can we think of a better name for `constant_term`?

src/functions.jl
```diff
@@ -6,649 +6,479 @@
...
+        function acos(a::$T)
+            a0 = constant_term(a)
+            a0^2 == one(a0) && throw(ArgumentError(
+                """
+                Recursion formula diverges due to vanishing `sqrt`
+                in the denominator.
+                """))
+
+            order = max_order(a)
+            c = $T( asin(a0), order )
```
dpsanders: `acos`?

src/functions.jl
```diff
@@ -6,649 +6,479 @@
...
+    @eval begin
+        function exp!(c::$T, a::$T, k::Int)
+            if k == 0
+                @inbounds c[1] = exp(constant_term(a))
+                return nothing
+            end
 
-Return the Taylor expansion of $\sin(a)$, of order `a.order`, for
-`a::Taylor1` polynomial.
+            c[k+1] = zero_korder(c, k)
```
dpsanders: What does `zero_korder` do?

src/functions.jl
```diff
@@ -6,649 +6,479 @@
...
+            if k == 0
+                @inbounds aux = tan( constant_term(a) )
+                @inbounds c[1] = aux
+                @inbounds c2[1] = aux^2
+                return nothing
+            end
+
+            c[k+1] = zero_korder(c, k)
+            @inbounds for i = 0:k-1
+                c[k+1] += (k-i)*a[k-i+1]*c2[i+1]
```
dpsanders: Add spaces around operators for consistency.

src/functions.jl
dpsanders: What is the correct term for this function?

src/other_functions.jl
```diff
@@ -38,68 +30,40 @@ for op in (:mod, :rem)
...
-## Division functions: rem and mod
-for op in (:mod, :rem)
-    @eval begin
-        @inbounds function ($op){T<:Real,S<:Real}(a::TaylorN{T}, x::S)
+        function ($op){T<:Real,S<:Real}(a::TaylorN{T}, x::S)
             coeffs = copy(a.coeffs)
-            y = ($op)(a[1][1], x)
-            coeffs[1] = HomogeneousPolynomial([y], 0)
+            # y = ($op)(a[1][1], x)
+            @inbounds coeffs[1] = HomogeneousPolynomial([($op)(a[1][1], x)], 0)
```
dpsanders: Use `constant_term(a)` here?

src/other_functions.jl
```diff
@@ -38,68 +30,40 @@ for op in (:mod, :rem)
...
             return TaylorN( coeffs, a.order )
         end
-        @inbounds function ($op){T<:Real}(a::TaylorN{Taylor1{T}}, x::T)
+
+        function ($op){T<:Real}(a::TaylorN{Taylor1{T}}, x::T)
             coeffs = copy(a.coeffs)
-            y = ($op)(a[1][1], x)
-            coeffs[1] = HomogeneousPolynomial([y], 0)
+            # y = ($op)(a[1][1], x)
+            @inbounds coeffs[1] = HomogeneousPolynomial([($op)(a[1][1], x)], 0)
```
dpsanders: And here.

src/other_functions.jl
```diff
@@ -108,34 +72,48 @@ for op in (:mod, :rem)
     end
 end
 
+
+function mod2pi{T<:Real}(a::Taylor1{T})
+    coeffs = copy(a.coeffs)
+    @inbounds coeffs[1] = mod2pi( a[1] )
```
dpsanders: Probably can unify some of these functions now.

src/power.jl
```diff
@@ -7,156 +7,251 @@
...
-
-    while (t -= 1) > 0
-        x = square(x)
+for T in (:Taylor1, :TaylorN)
+    @eval function ^{T<:Number}(a::$T{T}, n::Integer)
+        n == 0 && return one(a)
+        n == 1 && return a
+        n == 2 && return square(a)
+        n < 0 && return inv( a^(-n) )
+        return Base.power_by_squaring(a, n)
```
dpsanders: Maybe use some kind of binomial expansion instead?
Since after normalising we have something of the form `[1 + f(t)]^n`.
Hmm, sounds complicated.

dpsanders: Or just use `exp(n*log(a))`  here too -- have you compared efficiency?

src/power.jl
```diff
@@ -7,156 +7,251 @@
...
+for T = (:Taylor1, :TaylorN)
+    @eval function sqr!(c::$T, a::$T, k::Int)
+        if k == 0
+            @inbounds c[1] = a[1]^2
+            return nothing
+        end
 
-with $a$ a `Taylor1` polynomial.
+        if $T == Taylor1
+            @inbounds c[k+1] = zero(c[1])
```
dpsanders: Can these two lines be unified?

src/power.jl
```diff
@@ -7,156 +7,251 @@
...
+    iszero(a) && return nothing
+
+    T = eltype(c)
+    @inbounds num_coeffs_a = size_table[a.order+1]
+    @inbounds num_coeffs  = size_table[c.order+1]
+
+    @inbounds posTb = pos_table[c.order+1]
+
+    for na = 1:num_coeffs_a
+        @inbounds ca = a[na]
```
dpsanders: I think you can just put the `@inbounds` outside the `for` loop.

src/power.jl
```diff
@@ -7,156 +7,251 @@
...
+        ca == zero(T) && continue
+        @inbounds inda = index_table[a.order+1][na]
```
dpsanders: Try given a name to `index_table[a.order+1]` outside the loop. I expect that would be a bit more efficient. (Just like with `posTb`.)

src/power.jl
```diff
@@ -195,211 +282,96 @@ function sqrt(a::Taylor1)
...
+    @inbounds c[lnull+1] = aux
     @inbounds for k = lnull+1:a.order-l0nz
-        coeffs[k+1] = sqrtHomogCoef(k, v, coeffs, lnull)
+        sqrt!(c, a, k, lnull)
+    end
+
+    return c
+end
+
+function sqrt(a::TaylorN)
```
dpsanders: Can this function now be unified with the previous one?

src/functions.jl
lbenet: Sure!

src/functions.jl
lbenet: You are right; I thought I removed them from here...

src/functions.jl
lbenet: I agree. There is at least one exception: `tan`. There, there is an auxiliary function which is the square, so we save one (expensive?) calculation...

lbenet: Actually, there are other cases where it is convenient to have `aux` or `a0`... sorry for the randomness in naming this.

src/functions.jl
lbenet: Do you mean the definition of `r`? I'm not sure...

dpsanders: I mean that this is in the `acos` function, but the definition of `c` has `asin` in it.

src/functions.jl
lbenet: It avoids the nasty `if $T == Taylor1 ... end` block that were there. For `Taylor1` it simply returns `zero(a[1])`, while for `TaylorN` it returns `HomogeneousPolynomial(zero(a[1][1]),k`.

src/functions.jl
lbenet: Good point; I'll do it!

src/functions.jl
lbenet: I guess `inverse` or `inversion` would be ok; it implements Lagrange series inversion. 

Cc: @MikaelSlevinsky

src/other_functions.jl
lbenet: Good catch! I forgot about these functions...

src/functions.jl
dpsanders: That's why I was wondering if overloading `inv` makes sense here?

lbenet: I prefer not to use `inv` since it usually refers to the inverse multiplication, i.e. `1/a`.

src/other_functions.jl
lbenet: Good point! I'll do it.

src/auxiliary.jl
lbenet: I am writing it now as `zero_korder(a, n-1)`...

src/functions.jl
lbenet: It's actually a nice trick that @blas-ko came up with: the recursion formula is essentially the same as for `asin` except for a sign and the constant term.

dpsanders: OK, thanks. Maybe this should be mentioned in a comment somewhere.

src/power.jl
lbenet: The advantage of `power_by_squaring` is that, if you have `a::Taylor1{Int}`, the result is of the same type, as it would be expected. This is not preserved by `exp(n*log(a))`, which would return `Taylor1{Float64}`.

Long time ago I did some comparisons with what we have now and the same type of implementation for `^(a,r::Real)`; `power_by_squaring was better, at least for small `n`. 

The same subtlety about the type of the result appears here. The nice thing of preserving the type is that you get exact answers for Fateman-like tests.

src/power.jl
lbenet: I forgot completly this file; I'll implement the same changes as in `functions.jl` here.

src/functions.jl
MikaelSlevinsky: I'm ok with whatever term you choose. For the case that is implemented (which is numerically stable), I believe inversion and reversion are equivalent.

src/power.jl
lbenet: Done!

src/power.jl
lbenet: Good point; done!

src/power.jl
lbenet: The catch here is that it is simple to factorize `Taylor1` polynomials of the form `a2 t^2 + a3 t^3 +...` into something like `t^2 (a2 + a3 t +...)` so then taking the `sqrt` makes it simpler. The same trick doesn't apply in a straight forward way to `TaylorN` polynomials. (Yet, `Taylor1{Taylor1{T}}` behaves nicely again!)

src/functions.jl
lbenet: Good point.

src/functions.jl
lbenet: Ok. So, I'll change it into `inverse`.

lbenet:
I just pushed the changes we discussed. I'll add some more tests related to the new functions, perhaps some docstrings, before merging. Thanks again for the code review!

dpsanders:
Looks good and tests pass! Shall I merge?

lbenet:
Hold on... I'm checking performance, and it seems that it is worst.

lbenet:
Memory allocation has jump ~10x... I'll check this later

lbenet:
I just pushed another commit which addresses some optimizations in memory allocation I observed using `perf/fateman.jl`. The problem was that expressions like `c[i] += a[j] * b[k]` for `TaylorN`s, where creating temporary `HomogeneousPolynomial`s and therefore allocating memory. So I replaced them by `mul!(c[i], a[j], b[k])` which does not allocate such temporaries.

This achieves much better memory allocation and performs better than current master for `perf/fateman.jl`.

lbenet:
I'm merging... What ever is left, will be the subject of a new PR.

Issue #102, opened 2017-03-31
Notable performance slowdown after PR #99 in julia-0.6.0-pre.alpha


PerezHz:
While working with TaylorIntegration.jl and TaylorSeries.jl, I came across a notable slowdown after PR #99. While benchmarking TaylorIntegration.jl tests using commit [3a8e12e](https://github.com/PerezHz/TaylorIntegration.jl/pull/18/commits/3a8e12ede3b9f80777d70f19692f38092ce1a2d2) in PerezHz/TaylorIntegration.jl#18 in julia 0.6.0-pre.alpha.315 I got the following typical values (as always, I ran a "warmup lap" before benchmarking to discard compilation overhead):

Julia 0.6.0-pre.alpha.315 TaylorIntegration.jl tests benchmarks for different commits of TaylorSeries:

+ TaylorSeries latest master (commit 47b201c, PR #101):
`22.468 seconds (124.61 M allocations: 19.621 GiB)`

+ TaylorSeries commit 0697deb (PR #99):
`22.786 seconds (124.60 M allocations: 20.336 GiB, 15.47% gc time)`

+ TaylorSeries commit 487cb8b (PR #97):
`16.958 seconds (112.10 M allocations: 10.923 GiB, 14.04% gc time)`

So there's a slowndown of about 30% in typical execution times, about 11% more allocations, and about 50% more total bytes allocated when using commits 47b201c and 0697deb vs using commit 487cb8b in julia 0.6.0-pre.alpha. I also tested some "heavier" benchmarks on two different machines, for a more complicated problem than the ones included in TaylorIntegration.jl's tests, and the slowdown is even bigger, about 50%.

Interestingly enough, this slowdown does not occur in julia 0.5.1. Still, julia-0.5.1 fastest time is slower than julia-0.6.0-pre.alpha slowest time:

Julia 0.5.1 TaylorIntegration.jl tests benchmarks for different commits of TaylorSeries:

+ TaylorSeries commit 47b201c: 
`32.264 seconds (161.27 M allocations: 20.527 GB, 15.54% gc time)`

+ TaylorSeries commit 0697deb: 
`32.350 seconds (161.27 M allocations: 20.528 GB, 14.73% gc time)`

+ TaylorSeries commit 487cb8b: 
`32.849 seconds (161.27 M allocations: 20.527 GB, 14.51% gc time)`

lbenet:
[...]

PR #103, opened 2017-04-05
Solve performance regresion

lbenet:
See #102 for details of the problem.

### DIFF
--- a/perf/fateman.jl
+++ b/perf/fateman.jl
@@ -10,9 +10,9 @@ function fateman1(degree::Int)
     s = TaylorN( [oneH, HomogeneousPolynomial([one(T),one(T),one(T),one(T)],1)], degree )
     s = s^degree
     # s is converted to order 2*ndeg
-    s = TaylorN(s, 2*degree)
+    s = TaylorN(s.coeffs, 2*degree)
 
-    s * ( s+TaylorN(oneH, 2*degree) )
+    s * ( s+1 )
 end
 
 function fateman2(degree::Int)
@@ -22,7 +22,7 @@ function fateman2(degree::Int)
     s = TaylorN( [oneH, HomogeneousPolynomial([one(T),one(T),one(T),one(T)],1)], degree )
     s = s^degree
     # s is converted to order 2*ndeg
-    s = TaylorN(s, 2*degree)
+    s = TaylorN(s.coeffs, 2*degree)
     return s^2 + s
 end
 
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -447,7 +447,7 @@ c_k =  \frac{1}{b_0} (a_k - \sum_{j=0}^{k-1} c_j b_{k-j}).
 For `Taylor1` polynomials, `ordfact` is the order of the factorized
 term of the denominator.
 """
-@inline function div!(c::Taylor1, a::Taylor1, b::Taylor1, k::Int, ordfact::Int=0)
+function div!(c::Taylor1, a::Taylor1, b::Taylor1, k::Int, ordfact::Int=0)
     if k == 0
         @inbounds c[1] = a[ordfact+1] / b[ordfact+1]
         return nothing
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -15,21 +15,10 @@ for T in (:Taylor1, :TaylorN)
         =={T<:Number,S<:Number}(a::$T{T}, b::$T{S}) = ==(promote(a,b)...)
 
         function =={T<:Number}(a::$T{T}, b::$T{T})
-            la = a.order+1
-            lb = b.order+1
-            if a.order == b.order
-                # return all( a.coeffs .== b.coeffs )
-                return a.coeffs == b.coeffs
-            elseif a.order < b.order
-                # res1 = all( a[1:la] .== b[1:la] )
-                res1 = a[1:la] == b[1:la]
-                res2 = iszero(b[la+1:lb])
-            else a.order > b.order
-                # res1 = all( a[1:lb] .== b[1:lb] )
-                res1 = a[1:lb] == b[1:lb]
-                res2 = iszero(a[lb+1:la])
+            if a.order != b.order
+                a, b = fixorder(a, b)
             end
-            return res1 && res2
+            return a.coeffs == b.coeffs
         end
     end
 end
@@ -43,17 +32,11 @@ end
 
 ## zero and one ##
 for T in (:Taylor1, :TaylorN), f in (:zero, :one)
-    @eval begin
-        ($f)(a::$T) = $T([($f)(a[1])], a.order)
-
-        ($f)(a::$T, order) = $T([($f)(a[1])], order)
-    end
+    @eval ($f)(a::$T) = $T(($f)(a[1]), a.order)
 end
 
 function zero{T<:Number}(a::HomogeneousPolynomial{T})
-    a.order == 0 && return HomogeneousPolynomial([zero(T)], 0)
-    v = Array{T}( size_table[a.order+1] )
-    @__dot__ v = zero(a.coeffs)
+    v = zeros(a.coeffs)
     return HomogeneousPolynomial(v, a.order)
 end
 
@@ -70,9 +53,7 @@ zeros{T<:Number}(::Type{HomogeneousPolynomial{T}}, order::Int) =
     zeros( HomogeneousPolynomial([zero(T)], 0), order)
 
 function one{T<:Number}(a::HomogeneousPolynomial{T})
-    a.order == 0 && return HomogeneousPolynomial([one(a[1])], 0)
-    v = Array{T}( size_table[a.order+1] )
-    @__dot__ v = one(a.coeffs)
+    v = ones(a.coeffs)
     return HomogeneousPolynomial{T}(v, a.order)
 end
 
@@ -99,21 +80,12 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
             ($f){T<:Number,S<:Number}(a::$T{T}, b::$T{S}) = $f(promote(a,b)...)
 
             function $f{T<:Number}(a::$T{T}, b::$T{T})
-                la = a.order+1
-                lb = b.order+1
-                if a.order == b.order
-                    v = similar(a.coeffs)
-                    @__dot__ v = $f(a.coeffs, b.coeffs)
-                elseif a.order < b.order
-                    v = similar(b.coeffs)
-                    @inbounds @__dot__ v[1:la] = $f(a[1:la], b[1:la])
-                    @inbounds @__dot__ v[la+1:lb] = $f(b[la+1:lb])
-                else
-                    v = similar(a.coeffs)
-                    @inbounds @__dot__ v[1:lb] = $f(a[1:lb], b[1:lb])
-                    @inbounds @__dot__ v[lb+1:la] = a[lb+1:la]
+                if a.order != b.order
+                    a, b = fixorder(a, b)
                 end
-                return $T(v)
+                v = similar(a.coeffs)
+                @__dot__ v = $f(a.coeffs, b.coeffs)
+                return $T(v, a.order)
             end
 
             function $f(a::$T)
@@ -125,8 +97,7 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
             ($f){T<:Number,S<:Number}(a::$T{T}, b::S) = $f(promote(a,b)...)
 
             function $f{T<:Number}(a::$T{T}, b::T)
-                coeffs = similar(a.coeffs)
-                coeffs .= a.coeffs
+                coeffs = copy(a.coeffs)
                 @inbounds coeffs[1] = $f(a[1], b)
                 return $T(coeffs, a.order)
             end
@@ -258,8 +229,10 @@ for (T, W) in ((:Taylor1, :Number), (:TaylorN, :NumberNotSeriesN))
     @eval *{T<:$W, S<:$W}(a::$T{T}, b::$T{S}) = *(promote(a,b)...)
 
     @eval function *{T<:$W}(a::$T{T}, b::$T{T})
-        a, b = fixorder(a, b)
-        c = zero(a, a.order)
+        if a.order != b.order
+            a, b = fixorder(a, b)
+        end
+        c = $T(zero(a[1]), a.order)
         for ord = 0:c.order
             mul!(c, a, b, ord) # updates c[ord+1]
         end
@@ -287,6 +260,7 @@ end
 for T in (:Taylor1, :TaylorN)
     @eval function mul!(c::$T, a::$T, b::$T, k::Int)
 
+        # c[k+1] = zero( a[k+1] )
         @inbounds for i = 0:k
             if $T == Taylor1
                 c[k+1] += a[i+1] * b[k-i+1]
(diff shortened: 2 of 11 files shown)


lbenet:
@dpsanders Can you review this PR? It seems that #102 is solved with the last commit.

dpsanders:
Yes, I should be able to review it tonight or tomorrow.

lbenet:
Thanks, that would be great!

perf/fateman.jl
```diff
@@ -10,9 +10,9 @@ function fateman1(degree::Int)
     s = TaylorN( [oneH, HomogeneousPolynomial([one(T),one(T),one(T),one(T)],1)], degree )
     s = s^degree
     # s is converted to order 2*ndeg
-    s = TaylorN(s, 2*degree)
+    s = TaylorN(s.coeffs, 2*degree)
```
dpsanders: Why is this change to `s.coeffs` necessary?

perf/fateman.jl
```diff
@@ -10,9 +10,9 @@ function fateman1(degree::Int)
     s = TaylorN( [oneH, HomogeneousPolynomial([one(T),one(T),one(T),one(T)],1)], degree )
```
dpsanders: `s = 1 + w + x + y + z`?

dpsanders: Why don't you fix the degree here to be `2*degree` already?

perf/fateman.jl
```diff
@@ -22,7 +22,7 @@ function fateman2(degree::Int)
...
     s = s^degree
     # s is converted to order 2*ndeg
-    s = TaylorN(s, 2*degree)
+    s = TaylorN(s.coeffs, 2*degree)
```
dpsanders: Same comments here.

src/arithmetic.jl
```diff
@@ -15,18 +15,10 @@ for T in (:Taylor1, :TaylorN)
...
-            lb = b.order+1
-            if a.order == b.order
-                return all( a.coeffs .== b.coeffs )
-            elseif a.order < b.order
-                res1 = all( a[1:la] .== b[1:la] )
-                res2 = iszero(b[la+1:lb])
-            else a.order > b.order
-                res1 = all( a[1:lb] .== b[1:lb] )
-                res2 = iszero(a[lb+1:la])
+            if a.order != b.order
```
dpsanders: I suggest putting the `a.order != b.order` test inside `fixorder`, and just doing `a,b = fixorder(a, b)` here.

src/arithmetic.jl
```diff
@@ -40,17 +32,11 @@ end
 
 ## zero and one ##
 for T in (:Taylor1, :TaylorN), f in (:zero, :one)
-    @eval begin
-        ($f)(a::$T) = $T([($f)(a[1])], a.order)
-
-        ($f)(a::$T, order) = $T([($f)(a[1])], order)
-    end
+    @eval ($f)(a::$T) = $T(($f)(a[1]), a.order)
```
dpsanders: (In the future) we should think if `TaylorN` can literally be written in terms of `Taylor1` to avoid all this duplication of similar methods.

src/arithmetic.jl
```diff
@@ -96,19 +80,12 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
...
-                    @inbounds @__dot__ v[1:la] = $f(a[1:la], b[1:la])
-                    @inbounds @__dot__ v[la+1:lb] = $f(b[la+1:lb])
-                else
-                    @inbounds @__dot__ v[1:lb] = $f(a[1:lb], b[1:lb])
-                    @inbounds @__dot__ v[lb+1:la] = a[lb+1:la]
+                if a.order != b.order
+                    a, b = fixorder(a, b)
                 end
-                return $T(v)
+                v = similar(a.coeffs)
```
dpsanders: Why do you need this? Can't you just do `v = $f(a.coeffs, b.coeffs)` straight away?

src/arithmetic.jl
```diff
@@ -280,10 +256,11 @@ function *{T<:NumberNotSeriesN}(a::HomogeneousPolynomial{T}, b::HomogeneousPolyn
...
 
-# Homogeneous coefficient for the multiplication
+# Internal multiplication functions
 for T in (:Taylor1, :TaylorN)
-    @eval function mul!(c::$T, a::$T, b::$T, k::Int)
+    @eval @inline function mul!(c::$T, a::$T, b::$T, k::Int)
 
+        # c[k+1] = zero( a[k+1] )
         @inbounds for i = 0:k
             if $T == Taylor1
```
dpsanders: I suggest defining `mul!` for scalars instead to remove this special case.

src/arithmetic.jl
```diff
@@ -317,7 +294,7 @@ c_k = \sum_{j=0}^k a_j b_{k-j}.
 Return `c = a*b` with no allocation; all arguments are `HomogeneousPolynomial`.
 
 """
-function mul!(c::HomogeneousPolynomial, a::HomogeneousPolynomial,
+@inline function mul!(c::HomogeneousPolynomial, a::HomogeneousPolynomial,
         b::HomogeneousPolynomial)
 
     (iszero(b) || iszero(a)) && return nothing
```
dpsanders: You can just use `return`.

src/auxiliary.jl
```diff
@@ -18,7 +18,8 @@ function resize_coeffs1!{T<:Number}(coeffs::Array{T,1}, order::Int)
     lencoef = length(coeffs)
     order ≤ lencoef-1 && return nothing
     resize!(coeffs, order+1)
-    coeffs[lencoef+1:end] .= zero(coeffs[1])
+    z = zero(coeffs[1])
+    @__dot__ coeffs[lencoef+1:order+1] = z
```
dpsanders: This change is not needed I think.

dpsanders: It does the same as the simpler version. Or even do `coeffs[lencoef+1:end] .= 0`?

src/auxiliary.jl
```diff
@@ -35,7 +36,8 @@ function resize_coeffsHP!{T<:Number}(coeffs::Array{T,1}, order::Int)
     @assert order ≤ get_order() && lencoef ≤ num_coeffs
     num_coeffs == lencoef && return nothing
     resize!(coeffs, num_coeffs)
-    coeffs[lencoef+1:end] .= zero(coeffs[1])
+    z = zero(coeffs[1])
+    @__dot__ coeffs[lencoef+1:num_coeffs] = z
```
dpsanders: Same comment.

src/auxiliary.jl
```diff
@@ -71,10 +73,7 @@ Return the coefficient of order `n::Int` of a `a::Taylor1` polynomial.
 get_coeff(a::Taylor1, n::Int) = (@assert 0 ≤ n ≤ a.order+1;
     return a[n+1])
 
-function getindex(a::Taylor1, n::Int)
-    (1 ≤ n ≤ length(a.coeffs)) && return a.coeffs[n]
-    return zero(a.coeffs[1])
-end
+getindex(a::Taylor1, n::Int) = a.coeffs[n]
```
dpsanders: No bounds check required?

src/auxiliary.jl
```diff
@@ -167,17 +162,22 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
...
-For `a::Taylor1` returns `a.order` while for `a::TaylorN` returns
-`get_order()`.
-"""
-max_order(a::Taylor1) = a.order
+## fixorder ##
+for T in (:Taylor1, :TaylorN)
+    @eval begin
+        function fixorder(a::$T, b::$T)
+            a.order == b.order && return a, b
+            a.order < b.order && return $T(copy(a.coeffs), b.order),
```
dpsanders: No `copy` required here?

src/auxiliary.jl
```diff
@@ -167,17 +162,22 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
...
+                $T(copy(b.coeffs), b.order)
+            return $T(copy(a.coeffs), a.order), $T(copy(b.coeffs), a.order)
```
dpsanders: Or here?

src/constructors.jl
```diff
@@ -31,16 +31,19 @@ immutable Taylor1{T<:Number} <: AbstractSeries{T}
     ## Inner constructor ##
     function (::Type{Taylor1{T}}){T}(coeffs::Array{T,1}, order::Int)
         resize_coeffs1!(coeffs, order)
-        return new{T}(coeffs, length(coeffs)-1)
+        return new{T}(coeffs, order)
```
dpsanders: Do you need `@compat` here?

src/power.jl
```diff
@@ -40,6 +40,34 @@ for T in (:Taylor1, :TaylorN)
 end
 
 
+# power_by_squaring; slightly modified from base/intfuncs.jl
```
dpsanders: What is the modification? Why is it required?

src/power.jl
```diff
@@ -184,31 +211,33 @@ c_k & = & 2 \\sum_{j=0}^{(k-2)/2} a_{k-j} a_j + (a_{k/2})^2,
...
+        @inline function sqr!(c::$T, a::$T, k::Int)
+            if k == 0
+                @inbounds c[1] = constant_term(a)^2
+                return nothing
+            end
+
+            kodd = k%2
+            kend = div(k - 2 + kodd, 2)
+            @inbounds for i = 0:kend
+                if $T == Taylor1
```
dpsanders: Use `mul!`

src/power.jl
```diff
@@ -184,31 +211,33 @@ c_k & = & 2 \\sum_{j=0}^{(k-2)/2} a_{k-j} a_j + (a_{k/2})^2,
...
+            end
+            @inbounds c[k+1] = 2 * c[k+1]
+            kodd == 1 && return nothing
 
-        kodd = k%2
-        kend = div(k - 2 + kodd, 2)
-        @inbounds for i = 0:kend
             if $T == Taylor1
-                c[k+1] += a[i+1] * a[k-i+1]
+                @inbounds c[k+1] += a[div(k,2)+1]^2
```
dpsanders: Define `sqr!` for scalars and remove the special case.

dpsanders: (These tiny functions will be inlined.)

test/fateman40.jl
```diff
@@ -15,7 +15,7 @@ using Base.Test
             [oneH, HomogeneousPolynomial([one(T),one(T),one(T),one(T)],1)], degree)
         s = s^degree
         # s is converted to order 2*ndeg
-        s = TaylorN(s, 2*degree)
+        s = TaylorN(s.coeffs, 2*degree)
```
dpsanders: Again not sure why this change is necessary.

perf/fateman.jl
lbenet: The correct answer is "for historical reasons", which somehow do not apply any longer. As it is now, when you compute `s^degree` it only calculates up to `order=degree`, instead of putting lots of zeros (which are homogeneous polynomials of the correct order). I checked the impact in performance of putting `2*degree` there and, with respect to time there is none, but memory increases.

If you want, I can change it, but I don't think it is necessary for this PR.

perf/fateman.jl
lbenet: The reason is related to a method I deleted of `TaylorN` (as well as the corresponding for `Taylor1` and `HomogeneousPolynomial`). The deleted method (in constructors.jl) was `TaylorN{T<:Number}(x::TaylorN{T}, order::Int) = TaylorN{T}(x.coeffs, order)`. Those methods have side effects, which may yield some unexpected behaviour...

src/arithmetic.jl
lbenet: I tested that option and retested it now again including `@inline` for `fixorder`, and the benchmarks are slightly worst with respect to time, due to more memory use. I am leaving this as it is.

src/arithmetic.jl
lbenet: I fully agree. I have done some work along this, but it is largely unfinished. Actually, after reading different sources, it seems to me that it is better (in terms of memory) to have `Taylor1{Taylor1{T}}`, which effectively is a `TaylorN{T}` of two variables.

src/arithmetic.jl
lbenet: You certainly can simply do `v = $f(a.coeffs, b.coeffs)`, but as far as I understand, using `@__dot__` avoids creating some temporary arrays and copying the results, which is therefore faster. For the type of benchmarks that Jorge was checking in #102, without `@__dot__` in the addition functions produces ~30% more memory usage, and ~10% slower code.

dpsanders: In general, I agree. In this particular case, though, I don't see where the temporary will be.

src/arithmetic.jl
lbenet: We can certainly define a `mul` like method that returns `c[k+1] += a[i+1] * b[k-i+1]`. But my guess is that it will get inlined and produce the same code. Surely this will avoid the ugly code separating the `Taylor1` and `TaylorN` cases, but this happens at parse time, so I don't think we should care too much on this.

src/arithmetic.jl
lbenet: After reading your comment I also thought that this should produce no temporary array, but after checking it, it turn out that it does!

src/auxiliary.jl
lbenet: It is not really needed, you are right, but it yields a tiny performance improvement (at a very low cost in memory). The change you propose (`coeffs[lencoef+1:end] .= 0`) could yield the same or even better performance; yet, for `Taylor1{Taylor1{T}}` cases, and other mixtures, using `0` instead of `zero(coeffs[1])` produces slightly different things with respect to the order of the first coefficient. While this in general should not be important, I can think of cases where it may have consequences in the sense that the answer depends on the order of the coefficient.

For the time being, I'll leave it as it is, though it is a fair point to check those cases in detail because it may yield another performance improvement.

src/auxiliary.jl
lbenet: I don't think we require checking the bounds because, if we are out of the bounds, an error will be thrown.

src/auxiliary.jl
lbenet: I included the `copy` (knowing about the penalty that this entails) because without it you can change the coefficients of `a` (or `b`) and affect the *original* `a` and `b`, i.e., produce a side effect. 

Probably if `a.order == b.order` should also `return $T(copy(a.coeffs), a.order), $T(copy(b.coeffs), b.order)`, but for the moment, that is where `fixorder` is used, there are no side effects. Should I write that case with the `copy`s, since we are filtering the case `a.order == b.order` ? (You commented on this previously.)

src/constructors.jl
lbenet: Since there are no warnings for both Julia 0.5 and 0.6, I guess not...

src/power.jl
lbenet: Previously, in  487cb8b, we had this function, which turns out that it improves the performance with respect to using `Base.power_by_squaring`. 

The reason is that, wherever you read `square(x)`, in Base it is written as `x*x`, and `square(x)` is faster and uses less memory.

src/power.jl
lbenet: It could *almost* be replaced by `mul!(c,a,a,kend)`, but it can't. The reason is that the core of the loop is identical to the loop in `mul!`, but it updates `c[k+1]` instead of `c[kend+1]` as `mul!` would do.

src/power.jl
lbenet: `sqr!` for scalars will be non-mutating, so you will have to return the updated value. Then, either you keep the `if`, or return a corresponding value for `sqr!` when it acts on `HomogenousPolynomials`; the latter will have a cost in memory...

test/fateman40.jl
lbenet: Same reason as above; get rid of a method that had side effects.

lbenet:
@dpsanders Thanks a lot for the review; I just finished answering all your comments. As a result of your review, I pushed another commit that inlines `fixorder`. My local tests show that this improves further some benchmarks.

I'll squash it and merge it...

PR #106, opened 2017-05-13
broken link due to typo.

miguelraz:


lbenet:
Thanks for noticing this and sending a patch! Yet, I think that the problem affects only the "stable" version of the documentation, but not the "latest" one. Can you confirm this? 

Note that in the "stable" documentation is generated (automatically) once we tag a new version, which should occur soon, while the latest is uploaded when a commit is merged into master. Because of this, I suggest that you make the PR directly to the `gh-pages` branch of the repo. Though, as I said, we should be tagging a new version soon, which changes a bit the documentation.

miguelraz:
Can confirm, thanks for the explanation, sorry for the noise!
Will PR to `gh-pages` in the future.
Thanks!

lbenet:
Just a clarification: To correct the stable documentation directly, you need to make a PR to gh-pages branch. However, the next tagged version will erase those changes, so a normal PR would indeed be useful. Currently, the "latest" documentation has already diverged from the "stable" one, and no patch version tag is planned but a minor tag, so your contribution would be unnoticed.

PR #107, opened 2017-05-15
Correction in divfactorization

lbenet:
Use `iszero` instead of `isnan` and `isinf` in `divfactorization` to check if the polynomial is factoriz…


lbenet:
@dpsanders Do you agree to merge this?

Also, what about tagging a new version?

lbenet:
> Also, what about tagging a new version?

Or shall we wait that Julia 0.6 is released, and drop 0.5 support?

PR #111, opened 2017-06-20
Changed evaluate method for TaylorN and HomogeneousPolynomial.

blas-ko:
I've been working with `TaylorSeries` and `TaylorIntegration` in getting neighborhood solutions of dynamical systems using Jet Transport. This method requieres **a lot** of evaluations of `TaylorN` polynomials and actual `evaluate` methods is really slow. 

This PR changes the `TaylorN` and `HomogeneousPolynomial`  `evaluate` method into a mucho faster one without using Horner's method. I've tried it for `TaylorN`s of 2,3 & 4 variables at `order=7` & `order=14` and benchmarks are inspiring.
Here's an example:

```julia
#evaluate2 is my new evaluate function; evaluate is the actual TaylorSeries one.
julia> x = TaylorN([HomogeneousPolynomial([2,0.]),
             HomogeneousPolynomial([3.,2,1]),
             HomogeneousPolynomial([1.,2,3,4]),
             HomogeneousPolynomial([1,2,3,4,5.]),
             HomogeneousPolynomial([1,2,3,4,5.,6]),
             HomogeneousPolynomial([1,2,3,4,5.,6,7]),         
            ]);
julia> vars = 2; dx = ones(vars);
julia> ex = Taylor1(rand(12)); dxT = [ex,ex];

julia> @time evaluate2(x,dx) 
  0.000006 seconds (5 allocations: 176 bytes)
82.0
julia> @time evaluate(x,dx)
  0.001045 seconds (16.77 k allocations: 934.188 KB)
82.0
julia> @time evaluate2(x,dxT)
  0.000054 seconds (377 allocations: 33.859 KB)
82.0 + 754.0 t + 4155.0 t² + 17300.0 t³ + 57252.2 t⁴ + 158232.4 t⁵ + 373441.9 t⁶ + 759991.4 t⁷ + 1.3505463599999999e6 t⁸ + 2.0965756999999997e6 t⁹ + 2.84503888e6 t¹⁰ + 3.3817241e6 t¹¹ + 𝒪(t¹²)
julia> @time evaluate(x,dxT)
  0.003382 seconds (82.87 k allocations: 5.112 MB)
82.0 + 754.0 t + 4155.0 t² + 17300.0 t³ + 57252.2 t⁴ + 158232.39999999997 t⁵ + 373441.9 t⁶ + 759991.4 t⁷ + 1.35054636e6 t⁸ + 2.0965757000000002e6 t⁹ + 2.8450388800000004e6 t¹⁰ + 3.3817241e6 t¹¹ + 𝒪(t¹²)
``` 
Although allocation and time gets considerably better (which is, I think, because of Horner's rule not being efficient in a many variable evaluation), there is little rounding precision difference

```julia
julia> set_variables("x", numvars=2, order=17);
julia> xH = HomogeneousPolynomial([1,0]); yH = HomogeneousPolynomial([0,1],1); xT = TaylorN(xH, 17); yT = TaylorN(Int64, 2, order=17)
julia> exy = exp( xT+yT );

julia> evaluate2(exy,[0.1im,0.01im])
0.9939560979566969 + 0.10977830083717484im
julia> exp(0.11im) 
0.9939560979566968 + 0.10977830083717481im
```

### DIFF
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -166,7 +166,41 @@ function evaluate{T<:Number,S<:NumberNotSeriesN}(a::TaylorSeries.TaylorN{T},
         suma[homPol] = sun
     end
 
-    #here it only works with <:Real numbers...
-    return sum(sort(suma))
+    return sort_n_sum(suma)
 end
 evaluate{T<:Number}(a::TaylorN{T}) = a[1][1]
+
+## auxiliary function for a precise evaluation of evaluate
+"""
+    sort_n_sum(a)
+
+Auxiliary function for an accurate evaluation of `evaluate`.
+"""
+function sort_n_sum{T<:NumberNotSeriesN}(a::Array{T,1})
+
+    if T <: Real
+        suma = sum(sort(a,by=abs))
+
+    elseif T <: Complex
+        suma = sum(sort(real(a),by=abs) + sort(imag(a),by=abs)*im)
+
+    elseif T <: Taylor1
+        ord = a[1].order+1
+        T_type = eltype(a[1])
+        a_length = length(a)
+
+        tmp1 = Array(T_type,ord)
+        tmp2 = Array(T_type,a_length)
+
+        for o in 1:ord
+
+            for i in 1:a_length
+                tmp2[i] = a[i][o]
+            end
+
+            tmp1[o] = sum(sort(tmp2,by=abs))
+        end
+        suma = Taylor1(tmp1,ord)
+    end
+    return suma
+end
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -91,49 +91,7 @@ end
 
 
 
-## Evaluation
-"""
-    evaluate(a, vals)
-
-Evaluate a `HomogeneousPolynomial` polynomial using Horner's rule (hand coded)
-at `vals`.
-"""
-function evaluate{T<:Number,S<:NumberNotSeriesN}(a::HomogeneousPolynomial{T},
-        vals::Array{S,1} )
-    @assert length(vals) == get_numvars()
-    R = promote_type(T,S)
-    numVars = get_numvars()
-    suma = convert(TaylorN{R}, a)
-
-    @inbounds for nv = 1:numVars
-        suma = horner(suma, (nv, vals[nv]))
-    end
-
-    return suma[1][1]
-end
-evaluate(a::HomogeneousPolynomial) = zero(a[1])
-
-"""
-    evaluate(a, [vals])
-
-Evaluate the `TaylorN` polynomial `a` using Horner's rule (hand coded) at `vals`.
-If `vals` is ommitted, it is evaluated at zero.
-"""
-function evaluate{T<:Number,S<:NumberNotSeriesN}(a::TaylorN{T}, vals::Array{S,1} )
-    @assert length(vals) == get_numvars()
-    numVars = get_numvars()
-    R = promote_type(T,S)
-    suma = convert(TaylorN{R}, a)
-
-    @inbounds for nv = 1:numVars
-        suma = horner(suma, (nv, vals[nv]))
-    end
-
-    return suma[1][1]
-end
-
-evaluate{T<:Number}(a::TaylorN{T}) = a[1][1]
-
+## Evaluation of multivariable
 function evaluate!{T<:Number}(x::Array{TaylorN{T},1}, δx::Array{T,1},
         x0::Array{T,1})
     @assert length(x) == length(x0)
@@ -151,52 +109,57 @@ function evaluate!{T<:Number}(x::Array{Taylor1{TaylorN{T}},1}, δt::T,
     nothing
 end
 
-## Evaluates HomogeneousPolynomials and TaylorN on a val of the nv variable
-## using Horner's rule on the nv variable
-function horner{T<:Number,S<:NumberNotSeriesN}(a::HomogeneousPolynomial{T},
-        b::Tuple{Int,S} )
-    nv, val = b
-    numVars = get_numvars()
-    @assert 1 <= nv <= numVars
-    R = promote_type(T,S)
-    @inbounds indTb = coeff_table[a.order+1]
-    suma = TaylorN(zero(R), a.order)
+"""
+    evaluate(a, vals)
 
-    # Horner's rule on the nv variable
-    for ord = a.order : -1 : 0
-        suma_ord = TaylorN(zero(R), a.order)
-        posOrd = order_posTb(a.order, nv, ord)
-        neworder = a.order-ord
-        for pos in posOrd
-            c = a[pos]
-            iIndices = copy(indTb[pos])
-            iIndices[nv] = 0
-            kdic = in_base(get_order(), iIndices)
-            newpos = pos_table[neworder+1][kdic]
-            zhp = HomogeneousPolynomial([zero(R)], neworder)
-            zhp[newpos] = a[pos]
-            suma_ord += TaylorN(zhp, a.order)
+Evaluate a `HomogeneousPolynomial` polynomial at `vals`.
+"""
+function evaluate{T<:Number,S<:TaylorSeries.NumberNotSeriesN}(a::HomogeneousPolynomial{T},
+        vals::Array{S,1} )
+    @assert length(vals) == get_numvars()
+
+    num_vars = get_numvars()
+    ct = TaylorSeries.coeff_table[a.order+1]
+    R = promote_type(T,S)
+    suma = zero(R)
+
+    for (i,a_coeffs) in enumerate(a.coeffs)
+        tmp = vals[1]^(ct[i][1])
+        for n in 2:num_vars
+            tmp *= vals[n]^(ct[i][n])
         end
-        if ord == a.order
-            suma += suma_ord
-        else
-            suma = suma*val + suma_ord
+        suma += a_coeffs * tmp
+    end
+
+    return suma
+end
+evaluate(a::HomogeneousPolynomial) = zero(a[1])
+
+
+"""
+    evaluate(a, [vals])
+
+Evaluate the `TaylorN` polynomial `a` at `vals`.
+If `vals` is ommitted, it's evaluated at zero.
+"""
+function evaluate{T<:Number,S<:TaylorSeries.NumberNotSeriesN}(a::TaylorSeries.TaylorN{T},
+        vals::Array{S,1})
+    @assert length(vals) == get_numvars()
+
+    num_vars = get_numvars()
+    ct = TaylorSeries.coeff_table
+    R = promote_type(T,S)
+    suma = zero(R)
+    for homPol in 1:length(a)
+        for (i,a_coeff) in enumerate(a.coeffs[homPol].coeffs)
+            tmp = vals[1]^(ct[homPol][i][1])
+            for n in 2:num_vars
+                tmp *= vals[n]^(ct[homPol][i][n])
+            end
+            suma += a_coeff * tmp
         end
     end
 
     return suma
 end
-
-function horner{T<:Number,S<:NumberNotSeriesN}(a::TaylorN{T}, b::Tuple{Int,S} )
-
-    nv, val = b
-    @assert 1 <= nv <= get_numvars()
-    R = promote_type(T,S)
-
-    suma = TaylorN(zero(R), a.order)
-    for ord = a.order:-1:0
-        @inbounds polH = a[ord+1]
-        suma += horner( polH, b)
-    end
-    suma
-end
+evaluate{T<:Number}(a::TaylorN{T}) = a[1][1]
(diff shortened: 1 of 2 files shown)


dpsanders:
You are benchmarking in global scope. Please use e.g.
```
using BenchmarkTools

@btime evaluate($x, $dx)
```
and re-report all the timings.

blas-ko:
Thanks @dpsanders! Results are as following: 

```julia
## Using Github's `evaluate` version
julia> using BenchmarkTools, TaylorSeries

julia> x = TaylorN([HomogeneousPolynomial([2,0.]),
                   HomogeneousPolynomial([3.,2,1]),
                   HomogeneousPolynomial([1.,2,3,4]),
                   HomogeneousPolynomial([1,2,3,4,5.]),
                   HomogeneousPolynomial([1,2,3,4,5.,6]),
                   HomogeneousPolynomial([1,2,3,4,5.,6,7]),         
                  ]);

julia> vars = 2; dx = ones(vars);

julia> ex = Taylor1(collect(1:12)); dxT = [ex,ex];

julia> @btime evaluate(x,dx)
  669.109 μs (16763 allocations: 934.03 KiB)
82.0

julia> @btime evaluate(x,dxT)
  3.353 ms (86057 allocations: 5.36 MiB)
 82.0 + 754.0 t + 4155.0 t² + 17300.0 t³ + 59665.0 t⁴ + 179094.0 t⁵ + 482511.0 t⁶ + 1.191448e6 t⁷ 
+ 2.737209e6 t⁸ + 5.91686e6 t⁹ + 1.2139474e7 t¹⁰ + 2.3802504e7 t¹¹ + 𝒪(t¹²)


## Using PR's `evaluate` version
julia> include("TaylorSeries.jl"); using .TaylorSeries, BenchmarkTools

julia> x = TaylorN([HomogeneousPolynomial([2,0.]),
                   HomogeneousPolynomial([3.,2,1]),
                   HomogeneousPolynomial([1.,2,3,4]),
                   HomogeneousPolynomial([1,2,3,4,5.]),
                   HomogeneousPolynomial([1,2,3,4,5.,6]),
                   HomogeneousPolynomial([1,2,3,4,5.,6,7]),         
                  ]);

julia> vars = 2; dx = ones(vars);

julia> ex = Taylor1(collect(1:12)); dxT = [ex,ex];

julia> @btime evaluate(x,dx)
  486.596 ns (1 allocation: 16 bytes)
82.0

julia> @btime evaluate(x,dxT)
  28.216 μs (373 allocations: 33.70 KiB)
 82.0 + 754.0 t + 4155.0 t² + 17300.0 t³ + 59665.0 t⁴ + 179094.0 t⁵ + 482511.0 t⁶ + 1.191448e6 t⁷
 + 2.737209e6 t⁸ + 5.91686e6 t⁹ + 1.2139474e7 t¹⁰ + 2.3802504e7 t¹¹ + 𝒪(t¹²)
````

**Comment by dpsanders** [comment]
As I said, you need `$` signs before each variable:
```
@btime evaluate($x, $dxT)
```

**Comment by blas-ko** [comment]
Sorry @dpsanders! Deleted previews comment so it doesn't turn confusing. 
Results are as following:

```julia
## Using Github's `evaluate` version
julia> using BenchmarkTools, TaylorSeries

julia> x = TaylorN([HomogeneousPolynomial([2,0.]),
                   HomogeneousPolynomial([3.,2,1]),
                   HomogeneousPolynomial([1.,2,3,4]),
                   HomogeneousPolynomial([1,2,3,4,5.]),
                   HomogeneousPolynomial([1,2,3,4,5.,6]),
                   HomogeneousPolynomial([1,2,3,4,5.,6,7]),         
                  ]);

julia> vars = 2; dx = ones(vars);

julia> ex = Taylor1(collect(1:12)); dxT = [ex,ex];

julia> @btime evaluate($x,$dx)
  644.928 μs (16762 allocations: 934.02 KiB)
82.0

julia> @btime evaluate($x,$dxT)
  3.392 ms (86057 allocations: 5.36 MiB)
 82.0 + 754.0 t + 4155.0 t² + 17300.0 t³ + 59665.0 t⁴ + 179094.0 t⁵ + 482511.0 t⁶ + 1.191448e6 t⁷ +
 2.737209e6 t⁸ + 5.91686e6 t⁹ + 1.2139474e7 t¹⁰ + 2.3802504e7 t¹¹ + 𝒪(t¹²)
```
```julia
## Using PR's `evaluate` version
julia> include("TaylorSeries.jl"); using .TaylorSeries, BenchmarkTools

julia> x = TaylorN([HomogeneousPolynomial([2,0.]),
                   HomogeneousPolynomial([3.,2,1]),
                   HomogeneousPolynomial([1.,2,3,4]),
                   HomogeneousPolynomial([1,2,3,4,5.]),
                   HomogeneousPolynomial([1,2,3,4,5.,6]),
                   HomogeneousPolynomial([1,2,3,4,5.,6,7]),         
                  ]);

julia> vars = 2; dx = ones(vars);

julia> ex = Taylor1(collect(1:12)); dxT = [ex,ex];

julia> @btime evaluate($x,$dx)
  428.005 ns (0 allocations: 0 bytes)
82.0

julia> @btime evaluate($x,$dxT)
  27.398 μs (373 allocations: 33.70 KiB)
 82.0 + 754.0 t + 4155.0 t² + 17300.0 t³ + 59665.0 t⁴ + 179094.0 t⁵ + 482511.0 t⁶ + 1.191448e6 t⁷ +
 2.737209e6 t⁸ + 5.91686e6 t⁹ + 1.2139474e7 t¹⁰ + 2.3802504e7 t¹¹ + 𝒪(t¹²)
````

**Comment by dpsanders** [comment]
Thanks! Those results look very nice!

File: src/evaluate.jl [hunk 1]
```diff
@@ -151,52 +109,57 @@ function evaluate!{T<:Number}(x::Array{Taylor1{TaylorN{T}},1}, δt::T,
...
-            suma_ord += TaylorN(zhp, a.order)
-        end
-        if ord == a.order
-            suma += suma_ord
-        else
-            suma = suma*val + suma_ord
+    suma = zero(R)
+
+    for (i,a_coeffs) in enumerate(a.coeffs)
+        tmp = vals[1]^(ct[i][1])
```
dpsanders: Why is this efficient?

blas-ko: It just access each of the `HomogeneousPolynomial`s of `a` just once. I'm not sure if it's the most efficient way to do it when `vals` is of type `Array{Taylor1{S}}`, but when `vals` is just an array of numbers it's just number operations, which is pretty efficient.

PerezHz:
Wow, this looks great @blas-ko! Just wanted to confirm that I was able to reproduce the performance improvement on my machine! 👍

lbenet:
Thanks for the contribution! I just got to see this; sorry for being that late. I'll send you some comments soon...

src/evaluate.jl
```diff
@@ -151,52 +109,57 @@ function evaluate!{T<:Number}(x::Array{Taylor1{TaylorN{T}},1}, δt::T,
...
-        b::Tuple{Int,S} )
-    nv, val = b
-    numVars = get_numvars()
-    @assert 1 <= nv <= numVars
+"""
+    evaluate(a, vals)
+
+Evaluate a `HomogeneousPolynomial` polynomial at `vals`.
+"""
+function evaluate{T<:Number,S<:TaylorSeries.NumberNotSeriesN}(a::HomogeneousPolynomial{T},
```
lbenet: Here, you don't need `TaylorSeries.NumberNotSeriesN`; it's enough to have `NumberNotSeriesN`.

src/evaluate.jl
```diff
@@ -151,52 +109,57 @@ function evaluate!{T<:Number}(x::Array{Taylor1{TaylorN{T}},1}, δt::T,
...
+        vals::Array{S,1} )
+    @assert length(vals) == get_numvars()
+
+    num_vars = get_numvars()
+    ct = TaylorSeries.coeff_table[a.order+1]
```
lbenet: Same comment as above; `coeff_table[a.order+1]` suffices.

src/evaluate.jl
lbenet: To answer @dpsanders, if I understand correctly the code, what you are doing is, for each coefficient of `a::HomoegeneousPolynomial{T}` you are computing the actual monomial term which means to know the actual powers in the different variables, and perform the actual product. The reason that this outperforms the old code, as you say, is because you are not computing `HomoegeneousPolynomial{T}` quantities (i.e., you are not allocating any vector), but only `T` (or the correctly promoted type). Nice!

Part of the reason the code was writen as it was, was to be accurate as possible. Here we are loosing some accuracy. Do you have any ideas on this? Perhaps if you calculate each term, save it in a vector, and them add them up from the smallest to the biggest we can get a bit more precision at the cost of time, though.

src/evaluate.jl
blas-ko: Good catch  @lbenet , thanks!

src/evaluate.jl
blas-ko: same here...

src/evaluate.jl
blas-ko: > Part of the reason the code was writen as it was, was to be accurate as possible. Here we are loosing some accuracy. Do you have any ideas on this? Perhaps if you calculate each term, save it in a vector, and them add them up from the smallest to the biggest we can get a bit more precision at the cost of time, though.

I'm gonna try to implement this and see if precision gets better; thanks for the suggestion. Note that, at the time being, the precision loss in rounding the operations... lets see what happens.

lbenet:
I suggest to test the idea of adding things from the smallest value (in magnitude) to the largest, and see if (i) we get better accuracy and (ii) not too slow performance. If these becomes too complicated, we could merge this, and try to improve accuracy in another PR.

@dpsanders Do you have other suggestions?

blas-ko:
Actually your idea is working @lbenet ; accuracy increases as performance gets _a little_ worse but in the same order of magnitude (see example benchmark below). The problem, which I don't know yet how to manage, is on automating the sorting depending on the element type of `vals`. What I mean is the workaround for sorting a `Complex{T}` array is different from a `Float64` array than from a `Taylor1{T}` array, for example.  

Any ideas/suggestions on this? @lbenet @dpsanders @PerezHz 

```julia
##as the original PR
julia> @btime evaluate($x,$dx)
  240.531 ns (0 allocations: 0 bytes)

##saving as an array an then making the sumations (with numbers)
julia> @btime evaluate($x,$dx)
  288.433 ns (1 allocation: 144 bytes)
```

lbenet:
Can you push the last changes? I see only those which are independent of the accuracy...

blas-ko:
Pushed... I haven't done the workaround with magnitude sort for other types different than `<: Real`.

lbenet:
Tests are not passing. As you use sort, the sum considers first the most negative numbers, which maybe the source of the problem. So the idea is a bit more tricky than I originally thought. 

I suggest to revert to the previous commit and merge, and maybe try it in a different PR. 

@dpsanders Do you agree or have an alternative suggestion?

dpsanders:
I don't have any suggestions, sorry (except maybe use the absolute value to avoid problems with negative numbers?)

blas-ko:
I'm currently working in a specialized `sort_&_sum` function for this specific purpose. I don't know yet if it's gonna work, but let me try some test and I'll comment my results.

lbenet:
The following could perhaps work (instead of the `return sum(sort(suma))` line:

```julia
indxs = sortperm(abs.(A))
return sum(A[indxs])
```

Please, benchmark the results, to see if we revert them, or merge them.

lbenet:
... maybe better:
```julia
indxs = sortperm(abs.(A))
return sum( view(A,indxs) )
```

blas-ko:
So i pushed a new function `sort_n_sum` that does the job. I don't know if it's coded in the best style so @lbenet please make notes about it. Nevertheless, accuracy is better; results are as following: 

```julia
julia> include("TaylorSeries.jl"); using .TaylorSeries, BenchmarkTools

julia> x = TaylorN([HomogeneousPolynomial([2,0.]),
                   HomogeneousPolynomial([3.,2,1]),
                   HomogeneousPolynomial([1.,2,3,4]),
                   HomogeneousPolynomial([1,2,3,4,5.]),
                   HomogeneousPolynomial([1,2,3,4,5.,6]),
                   HomogeneousPolynomial([1,2,3,4,5.,6,7]),         
                  ]);

julia> vars = 2; dx = ones(vars);

julia> ex = Taylor1(collect(1:12)); dxT = [ex,ex];

julia> @btime evaluate($x,$dx)
  1.109 μs (10 allocations: 704 bytes)
82.0

julia> @btime evaluate($x,$dxT)
  25.214 μs (557 allocations: 45.59 KiB)
 82.0 + 754.0 t + 4155.0 t² + 17300.0 t³ + 59665.0 t⁴ + 179094.0 t⁵ + 482511.0 t⁶ + 1.191448e6 t⁷ +
2.737209e6 t⁸ + 5.91686e6 t⁹ + 1.2139474e7 t¹⁰ + 2.3802504e7 t¹¹ + 𝒪(t¹³)
```

lbenet:
I think your approach is not very julian (in `sort_n_sum`, in the sense that you should not decide depending on the type). Yet, I don't see any type stability. So I will merge this and try to the code later.

Thanks a lot!

lbenet:
Merged!

PR #112, opened 2017-06-23
More specialization of evaluate methods

lbenet:
Just made a couple of specialized methods; this improves slightly more #111


lbenet:
Merging...

PR #113, opened 2017-06-23
Add/fix evaluate methods for Poincare mapping with TaylorIntegration

PerezHz:
This PR changes and adds some evaluate methods which are used for Poincaré mapping with TaylorIntegration. Locally, tests are passing only in julia 0.6.0 and 0.7-DEV.

### DIFF
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -55,14 +55,14 @@ representing the Taylor expansion for the dependent variables
 of an ODE at *time* `δt`. It updates the vector `x0` with the
 computed values.
 """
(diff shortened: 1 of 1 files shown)


lbenet:
I am in favor of merging this. Yet, I have three suggestions:

- Can you add some tests?

- I wonder if other `evaluate` methods should also incorporate this type of change.

- Maybe, `AbstractArray{T,1}` is too wide. You could also add an abbreviation for a `Union`
type, which includes `Array` and `SubArray`, and which can be made broader if needed.

Maybe @dpsanders can guide us on the last point.

lbenet:
LGTM; I'm merging it. Thanks!

PerezHz:
Thanks for merging! 😄

Issue #116, opened 2017-08-22
Add function-like behavior for Taylor1, HomogeneousPolynomial and TaylorN?


PerezHz:
Hi! I was reading the [Function-like objects](https://docs.julialang.org/en/stable/manual/methods/#Function-like-objects-1) section of the Julia language documentation, and also @dpsanders's recent [blog post](https://julialang.org/blog/2017/08/dsl), and thought that perhaps it would be nice to add function-like behavior for Taylor1, TaylorN and HomogeneousPolynomial variables? My idea is to implement something like:

```julia
julia> using TaylorSeries

julia> t = Taylor1(16)
 1.0 t + 𝒪(t¹⁷)

julia> p = sin(t)
 1.0 t - 0.16666666666666666 t³ + 0.008333333333333333 t⁵ - 0.0001984126984126984 t⁷ + 2.7557319223985893e-6 t⁹ - 2.505210838544172e-8 t¹¹ + 1.6059043836821616e-10 t¹³ - 7.647163731819817e-13 t¹⁵ + 𝒪(t¹⁷)

julia> p(0.1)
ERROR: MethodError: objects of type TaylorSeries.Taylor1{Float64} are not callable

julia> (p::Taylor1)(x) = evaluate(p, x)

julia> p(0.1)
0.09983341664682815

julia> sin(0.1) #just comparing last answer to the actual result of evaluating sin at 0.1
0.09983341664682815
```

And similar stuff for HomogeneousPolynomials and TaylorNs... If you think this is a worthwhile feature to add to `TaylorSeries`, I'd be more than happy to submit a PR! 😄

lbenet:
It sounds as it could be interesting. Any comments or ideas, @dpsanders ?

dpsanders:
Seems like a good idea! (Is it common to evaluate Taylor series like this?)

PerezHz:
Thanks for the feedback! Well I guess it wouldn't hurt to be able to do `p(0.1)` as a shorter (and perhaps more legible) alternative to `evaluate(p, 0.1)`, where `p` is a `Taylor1` (and equivalent things for `TaylorN`s). Moreover, mathematically I think it does make sense to think about `p(0.1)` as the value of a polynomial function `p` at `0.1` (is that the question you're asking @dpsanders or did I misunderstand you?)

This notation is helpful, for example, when working with Legendre polynomials. Using this trick, one would be able to do things such as:

```julia
julia> using TaylorSeries

[...]

PR #118, opened 2017-08-30
Add function-like behavior for Taylor1, TaylorN, HomogeneousPolynomial

PerezHz:
This PR implements a solution for #116, and adds corresponding tests; tests are passing on my fork, thanks everyone for your feedback!

### DIFF
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -148,6 +148,11 @@ end
 
 evaluate(a::HomogeneousPolynomial) = zero(a[1])
 
+#function-like behavior for HomogeneousPolynomial
+(p::HomogeneousPolynomial)(x) = evaluate(p, x)
+
+(p::HomogeneousPolynomial)() = evaluate(p)
+
 """
     evaluate(a, [vals])
 
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -115,6 +115,15 @@ function evaluate!{T<:Number}(x::Array{TaylorN{T},1}, δx::Array{T,1},
 end
 
 function evaluate!{T<:NumberNotSeries}(x::Array{TaylorN{T},1}, δx::Array{Taylor1{T},1},
+        x0::Array{Taylor1{T},1})
+    @assert length(x) == length(x0)
+    @inbounds for i in eachindex(x)
+        x0[i] = evaluate( x[i], δx )
+    end
+    nothing
+end
+
+function evaluate!{T<:NumberNotSeries}(x::Array{TaylorN{Taylor1{T}},1}, δx::Array{Taylor1{T},1},
     x0::Array{Taylor1{T},1})
 @assert length(x) == length(x0)
 @inbounds for i in eachindex(x)
@@ -219,6 +228,30 @@ function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
     return suma
 end
 
+function evaluate{T<:NumberNotSeries}(a::TaylorN{Taylor1{T}},
+        vals::Array{Taylor1{T},1})
+    @assert length(vals) == get_numvars()
+
+    num_vars = get_numvars()
+    ct = coeff_table
+    a_length = length(a)
+    ord = maximum( get_order.(vals) )
+    suma = Taylor1(zeros(T, ord))
+
+    for homPol in 1:length(a)
+        sun = zero(Taylor1{T})
+        for (i,a_coeff) in enumerate(a.coeffs[homPol].coeffs)
+            tmp = vals[1]^(ct[homPol][i][1])
+            for n in 2:num_vars
+                tmp *= vals[n]^(ct[homPol][i][n])
+            end
+            suma += a_coeff * tmp
+        end
+    end
+
+    return suma
+end
+
 evaluate{T<:Number}(a::TaylorN{T}) = a[1][1]
 
 function evaluate{T<:Number}(x::Array{TaylorN{T},1}, δx::Array{T,1})
@@ -227,9 +260,14 @@ function evaluate{T<:Number}(x::Array{TaylorN{T},1}, δx::Array{T,1})
     return x0
 end
 
-function evaluate{T<:NumberNotSeries,S<:Number}(x::Array{TaylorN{T},1}, δx::Array{S,1})
-    R = promote_type(T,S)
-    x0 = Array{R}( length(x) )
+function evaluate{T<:NumberNotSeries}(x::Array{TaylorN{T},1}, δx::Array{Taylor1{T},1})
+    x0 = Array{Taylor1{T}}( length(x) )
+    evaluate!( x, δx, x0 )
+    return x0
+end
+
+function evaluate{T<:NumberNotSeries}(x::Array{TaylorN{Taylor1{T}},1}, δx::Array{Taylor1{T},1})
+    x0 = Array{Taylor1{T}}( length(x) )
     evaluate!( x, δx, x0 )
     return x0
 end
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -91,6 +91,16 @@ function evaluate{T<:Number}(a::Taylor1{T}, x::Taylor1{T})
     end
     suma
 end
+function evaluate{T<:NumberNotSeries}(a::Taylor1{Taylor1{T}}, x::Taylor1{T})
+    if a.order != x.order
+        a, x = fixorder(a, x)
+    end
+    @inbounds suma = a[end]
+    @inbounds for k = a.order:-1:1
+        suma = suma*x + a[k]
+    end
+    suma
+end
 
 evaluate{T<:Number,S<:Number}(p::Taylor1{T},x::Array{S}) = evaluate.(p,x)
 
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -48,7 +48,7 @@ function evaluate{T<:Number}(x::Array{Taylor1{T},1}, δt::T)
     evaluate!(x, δt, xnew)
     return xnew
 end
-evaluate{T<:Number}(a::Array{Taylor1{T},1}) = evaluate(a, zero(T))
+evaluate{T<:Number}(a::Array{Taylor1{T},1}) = evaluate.(a)
 
 doc"""
     evaluate!(x, δt, x0)
@@ -112,6 +112,15 @@ function evaluate!{T<:Number}(x::Array{TaylorN{T},1}, δx::Array{T,1},
         x0[i] = evaluate( x[i], δx )
     end
     nothing
+end
+
+function evaluate!{T<:NumberNotSeries}(x::Array{TaylorN{T},1}, δx::Array{Taylor1{T},1},
+    x0::Array{Taylor1{T},1})
+@assert length(x) == length(x0)
+@inbounds for i in eachindex(x)
+    x0[i] = evaluate( x[i], δx )
+end
+nothing
 end
 
 function evaluate!{T<:Number}(x::Array{Taylor1{TaylorN{T}},1}, δt::T,
@@ -218,7 +227,7 @@ function evaluate{T<:Number}(x::Array{TaylorN{T},1}, δx::Array{T,1})
     return x0
 end
 
-function evaluate{T<:NumberNotSeries,S<:NumberNotSeries}(x::Array{TaylorN{T},1}, δx::Array{S,1})
+function evaluate{T<:NumberNotSeries,S<:Number}(x::Array{TaylorN{T},1}, δx::Array{S,1})
     R = promote_type(T,S)
     x0 = Array{R}( length(x) )
     evaluate!( x, δx, x0 )
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -28,6 +28,10 @@ function evaluate{T<:Number,S<:Number}(a::Taylor1{T}, dx::S)
     end
     suma
 end
+
+(p::Taylor1)(x) = evaluate(p, x)
+(p::Taylor1)() = evaluate(p)
+
 evaluate{T<:Number}(a::Taylor1{T}) = a[1]
 
 doc"""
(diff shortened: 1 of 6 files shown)


PerezHz:
PR tests are passing, if there is anything that I missed or that you think should be added, please let me know 😄

blas-ko:
> if there is anything that I missed or that you think should be added, please let me know.

It would be a nice feature to take advantage of this PR to implement an evaluate method for an array of values. I'm thinking of something like
```julia
julia> using TaylorSeries
julia> import TaylorSeries: evaluate;
julia> function evaluate(p::Taylor1{T},x::Vector{S}) where {T<:Number,S<:Number}
    ll = length(x)
    p_eval = zeros(x)
    for (i,xi) in enumerate(x)
        @inbounds p_eval[i] = evaluate(p,xi)
    end
    return p_eval
end
julia> (p::Taylor1)(x) = evaluate(p,x) #this now works for numbers & arrays!!
julia> p = Taylor1([1,2,3,4])
 1 + 2 t + 3 t² + 4 t³ + 𝒪(t⁴)
julia> p(1.0)
 10.0
julia> p([1.0,0.0])
2-element Array{Float64,1}:
 10.0
  1.0
```
The analogous could be done for `TaylorN` but instead an `NxM` matrix should be passed as an argument where `N`: number of variables, `M`: number of values to be evaluated.

PerezHz:
I agree that this functionality should be available for the cases you mention, @blas-ko! Granted, I didn't add the methods you are proposing; but actually, as the way it is now, such cases may already be handled via broadcasting:

```julia
#evaluating p at a matrix of Float64s:
julia> p.([1.0 0.0; -1.0 -2.0])
2×2 Array{Float64,2}:
 10.0    1.0
 -2.0  -23.0

#evaluating p at a 3x3x3 matrix of Float64s:
julia> p.(rand(3,3,3))
3×3×3 Array{Float64,3}:
[:, :, 1] =
 8.65957  1.17927  8.12569
 1.13276  2.57914  3.37285
 2.51999  5.42389  2.04953

[:, :, 2] =
 1.03394  1.19053  3.51668
 1.15536  9.98952  4.27719
 7.29441  2.3461   2.76938

[:, :, 3] =
 1.33343  1.11886  1.08465
 5.75461  1.47925  1.78689
 1.47734  2.19666  1.84005
```
and for TaylorNs:
```julia
julia> ξ1, ξ2 = set_variables("ξ", numvars=2, order=15);

julia> f = sin(ξ1+ξ1*ξ2); g = exp(-sin(im*ξ1)+ξ1*ξ2); h = [f, g];

julia> f = sin(ξ1+ξ1*ξ2); g = exp(-sin(ξ1)+ξ1*ξ2); h = [f, g];

#evaluating f. at a vector of evaluation points returns a vector of Float64 values:
julia> f.([rand(2),rand(2)])
2-element Array{Float64,1}:
 0.974375
 0.585455

#evaluating g. at a 2x2 matrix of evaluation points returns a 2x2 matrix of Float64 values:
julia> g.( map(x->rand(2), rand(2,2)) )
2×2 Array{Float64,2}:
 0.929124  0.932631
 0.825413  0.888743

#evaluating h=[f,g] at a 2x2 matrix of evaluation points returns a 2x2 matrix of Vector{Float64}(2) values:
julia> h.( map(x->rand(2), rand(2,2)) )
2×2 Array{Array{Float64,1},2}:
 [0.674398, 0.736717]  [0.991339, 0.994949]
 [0.865741, 0.959931]  [0.460701, 0.800402]

#and so on...
julia> h.( [rand(2),rand(2)] )
2-element Array{Array{Float64,1},1}:
 [0.748772, 0.728316]
 [0.165423, 0.964655]

julia> h.( map(x->rand(2), rand(3,2,2)) )
3×2×2 Array{Array{Float64,1},3}:
[:, :, 1] =
 [0.970154, 0.996443]  [0.952156, 0.628598]
 [0.434873, 0.979518]  [0.750666, 0.492932]
 [0.839201, 0.475184]  [0.151065, 0.894432]

[:, :, 2] =
 [0.966398, 0.597349]  [0.414589, 0.901062]
 [0.045531, 0.992213]  [0.58178, 0.701201] 
 [0.395325, 0.718587]  [0.951338, 0.66144] 
```

I think this is, in a sense, more julian, but also more general; otherwise one would have to define a method for each particular case... The only caveat is that this behavior is only supported by julia 0.6+

PerezHz:
So, for the particular case case of Taylor1 it might be enough to define:

```julia
evaluate{T<:Number,S<:Number}(p::Taylor1{T},x::Vector{S}) = evaluate.(p,x)
```

Thus

```julia
julia> p = Taylor1(rand(10))

julia> vr = rand(5)

julia> p(vr) == p.(vr)
true

julia> p(vr) == evaluate.(p,vr)
true
```

I'm gonna push this extra method for Taylor1s and corresponding tests! Thanks @blas-ko for your suggestion! I'd like to use some similar broadcasting tricks for TaylorN, but then the second argument would have to be a `Vector{Vector{S}}`...

PerezHz:
Is there an easy way to convert from a `Vector{Vector{T}}` to an `Array{T,2}`?

blas-ko:
>  such cases may already be handled via broadcasting

You're absolutely right, my fault there!

> Is there an easy way to convert from a Vector{Vector{T}} to an Array{T,2}?

Maybe `hcat`
```julia
julia> a = [[1,2,3],[4,5,6]];
julia> @time hcat(a...)
  0.000038 seconds (8 allocations: 384 bytes)
3×2 Array{Int64,2}:
 1  1
 1  1
 2  3
```

PerezHz:
Many thanks, @blas-ko!

PerezHz:
> Is everything type stable? It seems to me that this is the case...

Well, as this functionality is being proposed right now, the function-like behavior of Taylor1, etc., at the end of the day relies on `evaluate`, so in principle this new behavior should be as type-stable as `evaluate` is... Do you have any concrete cases in mind?

> What about mixed-types, i.e. Taylor1{TaylorN{T}}, or TaylorN{Taylor1{T}} ?

I'm gonna add some tests for the mixtures! That's actually something important that I didn't add! 😅 sorry about that!  Regarding the function-like behavior for mixtures, this again relies on `evaluate`, so there should be no problem, but I'm adding some tests which should check this

PerezHz:
Thanks for all your comments @lbenet! I actually had to tweak a bit some parts of `src/evaluate.jl` for the mixtures, but now everything's working smoothly. Tests are passing locally on julia 0.5.2, 0.6.0 and 0.7.0-DEV.1486, but let's see what travis thinks... An important thing that I would like to note is that the tweaks in `src/evaluate.jl` proposed here are also consistent with TaylorIntegration.jl tests on latest master for julia 0.5.2, 0.6.0 and 0.7.0-DEV

PerezHz:
@lbenet @dpsanders there was a weird error on travis [job#652.5](https://travis-ci.org/JuliaDiff/TaylorSeries.jl/jobs/270596125), could you please help me restart it? otherwise, tests are passing

dpsanders:
I restarted the job. 

Actually, with the syntax t(3) to evaluate, is there now any need for the function called evaluate?

PerezHz:
>I restarted the job.

Thanks, David!

> Actually, with the syntax t(3) to evaluate, is there now any need for the function called evaluate?

Well, currently `t(3)` is only an alternative syntax for `evaluate(t,3)` (is that what is called syntactic sugar?), i.e. `t(3) == evaluate(t,3)`, where t is a Taylor1... But I agree, if done carefully, I think it would be possible to get rid completely of the `evaluate` function, if we substituted the code for the evaluate methods in the definition of the `(p::Taylor1)(x)` methods, etc. On the other hand, probably there still would be a need for the in-place version of `evaluate` (i.e., the `evaluate!` function).

PerezHz:
The last commit addresses more mixture cases, such as evaluating a `TaylorN{Taylor1{Float64}}` or an `Array{TaylorN{Taylor1{Float64}}}` at an `Array{Taylor1{Float64}}`; I added the corresponding `evaluate` methods and tests... On travis, once again there was a job ([travis job #653.5](https://travis-ci.org/JuliaDiff/TaylorSeries.jl/jobs/270718728)) with some weird errors which had nothing to do with TaylorSeries.jl; after restarting everything should be working fine... Locally, tests are passing in julia 0.5.2, 0.6.0 and 0.7.0-DEV.1599

lbenet:
> Actually, with the syntax t(3) to evaluate, is there now any need for the function called evaluate?

Probably we should keep `evaluate` for a while, and eventually not exporting it or simply deprecate it.

I just restarted one test that hung.

PerezHz:
> Probably we should keep evaluate for a while, and eventually not exporting it or simply deprecate it.

I agree!

> I just restarted one test that hung.

Many thanks, @lbenet!

lbenet:
Tests are passing. Could you add something to the docs, and then we merge this? Thanks for all the work!

PerezHz:
> Could you add something to the docs, and then we merge this?

Sure! I added some general details about this new behavior in the User's Guide section of the docs. I was also thinking that I should add some docstrings, but I was thinking whether I should add those in the  `evaluate` docstrings, or in the docstrings for the constructors of `Taylor1`, `HomogeneousPolynomial` and `TaylorN` variables?

I'm asking this because, as I understood from JuliaDocs/Documenter.jl#228 and JuliaDocs/Documenter.jl#290, currently it is not possible to write docstrings for these kinds of methods, and the recommended approach mentioned there is to document within the type's docstring that objects of that type are callable and provide a usage example there. So, how would you recommend me to proceed?

PerezHz:
I've updated the docstrings for `evaluate` as well as the docstrings for the `Taylor1`, `HomogeneousPolynomial` and `TaylorN` constructors. That's the only change since last commit, so travis should give green lights. If there is anything that you think should be added, please let me know; otherwise, this should be ready to be merged...

PerezHz:
Job `#19.4` is failing on travis... While working with TaylorSeries.jl, I've seen this error is related to a test which fails from time to time (`test/onevariable.jl`, line 343) due to use of random variables, but if re-run, then it should pass

PerezHz:
Tests are now passing!

src/evaluate.jl
```diff
@@ -88,8 +97,25 @@ function evaluate{T<:Number}(a::Taylor1{T}, x::Taylor1{T})
     end
     suma
 end
+function evaluate{T<:NumberNotSeries}(a::Taylor1{Taylor1{T}}, x::Taylor1{T})
```
lbenet: `NumberNotSeries` here may be too restrictive; I'm thinking in cases such as `Taylor1{Taylor1{...{T}...}`. Would everything work if `T<:Number`?

src/evaluate.jl
```diff
@@ -100,6 +126,25 @@ function evaluate!{T<:Number}(x::Array{TaylorN{T},1}, δx::Array{T,1},
     end
     nothing
 end
+
```
lbenet: I think the two methods below can be written as one, imposing `T<:NumberNotSeriesN`.

src/evaluate.jl
```diff
@@ -190,10 +245,59 @@ function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
...
 evaluate{T<:Number}(a::TaylorN{T}) = a[1][1]
 
 function evaluate{T<:Number}(x::Array{TaylorN{T},1}, δx::Array{T,1})
     x0 = Array{T}( length(x) )
     evaluate!( x, δx, x0 )
     return x0
 end
+
+function evaluate{T<:NumberNotSeries}(x::Array{TaylorN{T},1}, δx::Array{Taylor1{T},1})
```
lbenet: Again, I think the two following methods can be written as one, using `T<:NumberNotSeriesN`.

src/evaluate.jl
PerezHz: Sure, it should work! Thanks!

src/evaluate.jl
PerezHz: Thanks @lbenet!

src/evaluate.jl
PerezHz: 👌Great! Thanks!

PerezHz:
@lbenet: just pushed the changes you suggested; tests are passing locally on 0.5.2, 0.6.0, and 0.7.0-DEV and are running on travis right now. Again, many thanks!

lbenet:
Thanks a lot for your contribution! I am merging it right now!

blas-ko:
I was just checking evaluation and it seems that broadcasting is pretty slow; check the following
```julia
julia> using TaylorSeries; import TaylorSeries: evaluate 

julia> function evaluate{T<:Number}(a::Taylor1{T},vals::Vector{T})
           ll = length(vals)
           sumas = zeros(T,ll)
           for i in 1:ll
               sumas[i] = evaluate(a,vals[i])
           end
           return sumas
       end
evaluate (generic function with 15 methods)

julia> function evaluate{T<:Number,S<:Number}(a::Taylor1{T},vals::Vector{S})
           ll = length(vals)
           R = promote_type(S,T)
           sumas = zeros(R,ll)
           for i in 1:ll
               sumas[i] = evaluate(a,vals[i])
           end
           return sumas
       end
evaluate (generic function with 16 methods)

julia> t = Taylor1(10); a = collect(0.:0.5:10);

julia> using BenchmarkTools

julia> @btime evaluate(t,a); @btime evaluate.(t,a);
  402.035 ns (1 allocation: 256 bytes)
  7.380 μs (26 allocations: 1.31 KiB)
```
I've not done this with the other cases, but in this particular case the difference is _yuuuge_. Maybe it's worth it to write the specific methods for evaluating in vectors or matrices.

lbenet:
@blas-ko I just merged this PR, seemingly at the same time you wrote your comment, though I was a bit faster ;-).

You are right that broadcasting (here) incurs in some allocations that cost time, roughly a factor 18. Note that
```julia
julia> @which evaluate.(t,a)
broadcast(f, A, Bs...) in Base.Broadcast at broadcast.jl:434
```
so my guess is that internally `broadcast` calls `evaluate(t,a[i])` with some internal evaluations,  which are not as efficient as the explicit method. In other words, if there is a method that does broadcasting, use it.

Issue #68, opened 2016-10-13
Implement an `expand` function? 


blas-ko:
I was thinking of implementing a function so it expands an independent variable of a given order around a function `f`? This could be something like

``` julia
function expand(f::Function,order::Int64=15)
    a = Taylor1(order)
    return f(a)
end
```

I don't know if it's very efficient because the function doesn't know a priori the output's type, but I think it's practical for `TaylorSeries` users.

PerezHz:
Sounds interesting!

```julia
julia> import Base.expand

julia> function expand(f::Function,order::Int64=15)
           a = Taylor1(order)
           return f(a)
       end
expand (generic function with 3 methods)

julia> expand(sin)
 1.0 t - 0.16666666666666666 t³ + 0.008333333333333333 t⁵ - 0.0001984126984126984 t⁷ + 2.7557319223985893e-6 t⁹ - 2.505210838544172e-8 t¹¹ + 1.6059043836821616e-10 t¹³ - 7.647163731819817e-13 t¹⁵ + 𝒪(t¹⁶)

julia> typeof(sin)
Base.#sin

julia> expand(λ->sin(cos(λ)))
 0.8414709848078965 - 0.2701511529340699 t² - 0.08267127702314792 t⁴ + 0.028036523686489446 t⁶ - 0.0019241418790800977 t⁸ - 0.0004838563431246893 t¹⁰ + 0.00013040017931079692 t¹² - 1.2102311792022246e-5 t¹⁴ + 𝒪(t¹⁶)

julia> x = Taylor1(15)
 1.0 t + 𝒪(t¹⁶)

julia> sin(cos(x))
 0.8414709848078965 - 0.2701511529340699 t² - 0.08267127702314792 t⁴ + 0.028036523686489446 t⁶ - 0.0019241418790800977 t⁸ - 0.0004838563431246893 t¹⁰ + 0.00013040017931079692 t¹² - 1.2102311792022246e-5 t¹⁴ + 𝒪(t¹⁶)

julia> sin(cos(x)) == expand(λ->sin(cos(λ)))
true
```

Perhaps it would be helpful to be able to choose the initial point of expansion? Otherwise, it will always expand around zero. I mean something like:

```julia
julia> using TaylorSeries

julia> import Base.expand

julia> function expand{T<:Number}(f::Function, x0::T, order::Int64=15)
           a = Taylor1([x0, one(x0)], order)
           return f(a)
       end
[...]

Issue #119, opened 2017-08-30
Add method for evaluating a TaylorN on a Vector{TaylorN}


PerezHz:
While working on PR #118 (see also discussion in #116), I was thinking that perhaps it'd be nice to have defined  `evaluate{T<:NumberNotSeries}(a::TaylorN{T}, vals::Array{TaylorN{T},1})`, or something similar, in order to be able to compose Taylor expansions of N-variable functions. There is already a similar method for `Taylor1`, which allows to evaluate a `Taylor1` on a `Taylor1`.

What I'm trying to say (using syntax from #118) is that whereas it is possible to compose two `Taylor1`s,

```julia
julia> using TaylorSeries

julia> t = Taylor1(25)
 1.0 t + 𝒪(t²⁶)

julia> p = sin(t)
 1.0 t - 0.16666666666666666 t³ + 0.008333333333333333 t⁵ - 0.0001984126984126984 t⁷ + 2.7557319223985893e-6 t⁹ - 2.505210838544172e-8 t¹¹ + 1.6059043836821616e-10 t¹³ - 7.647163731819817e-13 t¹⁵ + 2.811457254345521e-15 t¹⁷ - 8.220635246624331e-18 t¹⁹ + 1.9572941063391263e-20 t²¹ - 3.868170170630684e-23 t²³ + 6.446950284384474e-26 t²⁵ + 𝒪(t²⁶)

julia> q = cos(t)
 1.0 - 0.5 t² + 0.041666666666666664 t⁴ - 0.001388888888888889 t⁶ + 2.48015873015873e-5 t⁸ - 2.7557319223985894e-7 t¹⁰ + 2.08767569878681e-9 t¹² - 1.1470745597729726e-11 t¹⁴ + 4.779477332387386e-14 t¹⁶ - 1.5619206968586228e-16 t¹⁸ + 4.1103176233121653e-19 t²⁰ - 8.896791392450574e-22 t²² + 1.6117375710961184e-24 t²⁴ + 𝒪(t²⁶)

julia> p(q) #evaluate a Taylor1 on a Taylor1; compare with: sc = x->sin(cos(x)); sc(t)
[...]

Issue #80, opened 2016-11-22
update! functionality


BenLauwens:
Hi

Often when I work with TaylorSeries, I need to update the coefficients of a TaylorSeries around a next point.
A very convenient feature would be an update! function.
Example:
`x = Taylor1([1.0, 2.0, 3.0])`
`update!(x, 1.0)`
where the result would give for `x`
`6+8t+3t^2+O(t^3)`
As a workaround I am using for the moment
`t = Taylor1[1.0, 1.0]`
`x = evaluate(x, t)`
Looking at the code some optimization in update! is possible. A generalized Horner procedure can do this efficiently. If there is some interest, I will post tomorrow a possible implementation.

lbenet:
Hi,
first of all, let me apologize because I just noticed today your issue, which is already two weeks old. (Somehow, github notifications not always manage to get to my email.)

If I understand it correctly your idea, `update!(x, 1.0)` corresponds to `evaluate(x,1.0+Taylor1(1))`, so some kind of shift of the point were you are computing the Taylor expansion. (Note that this may not be an optimized version.) Please, do send a pull request with your implementation, with documentation, and I'll merge it. The only thing which may be a bit misleading is the name of the function.

lbenet:
This has been addressed and implemented in #121. I'm thus closing this.

PR #121, opened 2017-08-31
taylor_expand function for Taylor1 and TaylorN 

blas-ko:
This makes reference to #68. A `taylor_expand` function is added for `Taylor1` and `TaylorN` expansions around a `Number` or `Vector`, respectively.

_**Obs:** The way "common" functions are expanded for arrays such as `taylor_expand(cos.[0.0,1.0])` for julia 0.6 output a warning about a method deprecation for arrays. They suggest to use now the **dot** operator instead, but this generates some problems with user defined functions._

### DIFF
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -137,6 +137,15 @@ function evaluate!{T<:NumberNotSeriesN}(x::Array{TaylorN{T},1}, δx::Array{Taylo
     nothing
 end
 
+function evaluate!{T<:NumberNotSeriesN}(x::Array{TaylorN{T},1}, δx::Array{TaylorN{T},1},
+        x0::Array{TaylorN{T},1})
+    @assert length(x) == length(x0)
+    @inbounds for i in eachindex(x)
+        x0[i] = evaluate( x[i], δx )
+    end
+    nothing
+end
+
 function evaluate!{T<:Number}(x::Array{Taylor1{TaylorN{T}},1}, δt::T,
         x0::Array{TaylorN{T},1})
     @assert length(x) == length(x0)
@@ -224,7 +233,6 @@ function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
     suma = Taylor1(zeros(R, ord))
 
     for homPol in 1:length(a)
-        sun = zero(R)
         for (i,a_coeff) in enumerate(a.coeffs[homPol].coeffs)
             tmp = vals[1]^(ct[homPol][i][1])
             for n in 2:num_vars
@@ -248,7 +256,6 @@ function evaluate{T<:NumberNotSeries}(a::TaylorN{Taylor1{T}},
     suma = Taylor1(zeros(T, ord))
 
     for homPol in 1:length(a)
-        sun = zero(Taylor1{T})
         for (i,a_coeff) in enumerate(a.coeffs[homPol].coeffs)
             tmp = vals[1]^(ct[homPol][i][1])
             for n in 2:num_vars
@@ -260,6 +267,31 @@ function evaluate{T<:NumberNotSeries}(a::TaylorN{Taylor1{T}},
 
     return suma
 end
+
+function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
+        vals::Array{TaylorN{S},1})
+    @assert length(vals) == get_numvars()
+
+    num_vars = get_numvars()
+    ct = coeff_table
+    R = promote_type(T,eltype(S))
+    a_length = length(a)
+    ord = maximum( get_order.(vals) )
+    suma = zero(TaylorN{R})
+
+    for homPol in 1:length(a)
+        for (i,a_coeff) in enumerate(a.coeffs[homPol].coeffs)
+            tmp = vals[1]^(ct[homPol][i][1])
+            for n in 2:num_vars
+                tmp *= vals[n]^(ct[homPol][i][n])
+            end
+            suma += a_coeff * tmp
+        end
+    end
+
+    return suma
+end
+
 
 evaluate{T<:Number}(a::TaylorN{T}) = a[1][1]
 
@@ -275,6 +307,12 @@ function evaluate{T<:NumberNotSeriesN}(x::Array{TaylorN{T},1}, δx::Array{Taylor
     return x0
 end
 
+function evaluate{T<:NumberNotSeriesN}(x::Array{TaylorN{T},1}, δx::Array{TaylorN{T},1})
+    x0 = Array{TaylorN{T}}( length(x) )
+    evaluate!( x, δx, x0 )
+    return x0
+end
+
 evaluate{T<:Number}(x::Array{TaylorN{T},1}) = evaluate.(x)
 
 #function-like behavior for TaylorN
@@ -286,4 +324,3 @@ evaluate{T<:Number}(x::Array{TaylorN{T},1}) = evaluate.(x)
 (p::Array{TaylorN{T},1}){T<:Number}(x) = evaluate(p, x)
 
 (p::Array{TaylorN{T},1}){T<:Number}() = evaluate(p)
-
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -202,16 +202,15 @@ function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
     @assert length(vals) == get_numvars()
 
     num_vars = get_numvars()
-    ct = coeff_table
     R = promote_type(T,S)
     a_length = length(a)
     suma = zeros(R,a_length)
     for homPol in 1:length(a)
         sun = zero(R)
         for (i,a_coeff) in enumerate(a.coeffs[homPol].coeffs)
-            tmp = vals[1]^(ct[homPol][i][1])
+            tmp = vals[1]^(coeff_table[homPol][i][1])
             for n in 2:num_vars
-                tmp *= vals[n]^(ct[homPol][i][n])
+                tmp *= vals[n]^(coeff_table[homPol][i][n])
             end
             sun += a_coeff * tmp
         end
@@ -226,7 +225,6 @@ function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
     @assert length(vals) == get_numvars()
 
     num_vars = get_numvars()
-    ct = coeff_table
     R = promote_type(T,S)
     a_length = length(a)
     ord = maximum( get_order.(vals) )
@@ -234,9 +232,9 @@ function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
 
     for homPol in 1:length(a)
         for (i,a_coeff) in enumerate(a.coeffs[homPol].coeffs)
-            tmp = vals[1]^(ct[homPol][i][1])
+            tmp = vals[1]^(coeff_table[homPol][i][1])
             for n in 2:num_vars
-                tmp *= vals[n]^(ct[homPol][i][n])
+                tmp *= vals[n]^(coeff_table[homPol][i][n])
             end
             suma += a_coeff * tmp
         end
@@ -250,16 +248,15 @@ function evaluate{T<:NumberNotSeries}(a::TaylorN{Taylor1{T}},
     @assert length(vals) == get_numvars()
 
     num_vars = get_numvars()
-    ct = coeff_table
     a_length = length(a)
     ord = maximum( get_order.(vals) )
     suma = Taylor1(zeros(T, ord))
 
     for homPol in 1:length(a)
         for (i,a_coeff) in enumerate(a.coeffs[homPol].coeffs)
-            tmp = vals[1]^(ct[homPol][i][1])
+            tmp = vals[1]^(coeff_table[homPol][i][1])
             for n in 2:num_vars
-                tmp *= vals[n]^(ct[homPol][i][n])
+                tmp *= vals[n]^(coeff_table[homPol][i][n])
             end
             suma += a_coeff * tmp
         end
@@ -273,7 +270,6 @@ function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
     @assert length(vals) == get_numvars()
 
     num_vars = get_numvars()
-    ct = coeff_table
     R = promote_type(T,eltype(S))
     a_length = length(a)
     ord = maximum( get_order.(vals) )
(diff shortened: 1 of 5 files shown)


lbenet:
I like the idea of this PR, though I am still not sure about the name of the function (and perhaps other things), see below. 

My proposal here is to twist a little the essence, and address #80. That is, given `x::Taylor1` (for concreteness I'll focus on `Taylor1`s), the idea is to return `x` (probably update it in place), such that the new `x` is the polynomial evaluated at a different location. Then, `update{T}(x::Taylor1{T}, t0::T)` would be equivalent to `evaluate(x, Taylor1( [t0,one(t0)], x.order ))`. The tricky part (and I am not sure if this is possible or not to accomplish) is to *only* update `x`; maybe the following works:
```julia
function update!(x::Taylor1, t0::TaylorSeries.NumberNotSeries)
    x.coeffs .= evaluate(x, Taylor1( [t0,one(t0)], x.order ))[:]
    return x
end
```
If this works, I guess this could also be extended to `HomogeneousPolynomial` and `TaylorN`, but I am not sure.

Surely, the original idea of this PR is more general, in the sense that it returns the evaluated expansion of a given function `f` around a point. This could still be considered.

The distinction of returning a `Taylor1` or a `TaylorN` is by dispatch on `x0`. That's ok. Yet, I don't think that `taylor_evaluate` should change the number of variables nor the order for `TaylorN`; I think this should be done by the user explicitly, and `taylor_evaluate` should rely on `get_order()` and `get_numvars()`.

lbenet:
I think the idea above works

```julia
julia> using TaylorSeries

julia> function update!(x::Taylor1, t0::TaylorSeries.NumberNotSeries)
           x.coeffs .= evaluate(x, Taylor1( [t0,one(t0)], x.order ))[:]
           return nothing
       end
update! (generic function with 1 method)

julia> x = Taylor1([1.0,2.0,3.0])
 1.0 + 2.0 t + 3.0 t² + 𝒪(t³)

julia> update!(x, 1.0)

julia> x
 6.0 + 8.0 t + 3.0 t² + 𝒪(t³)

julia> update!(x, -1.0)

julia> x
 1.0 + 2.0 t + 3.0 t² + 𝒪(t³)
```

lbenet:
Incidentally, the fact that tests are not passing may be related to changing the order and the number of variables; see above.

blas-ko:
>  given x::Taylor1 (for concreteness I'll focus on Taylor1s), the idea is to return x (probably update it in place), such that the new x is the polynomial evaluated at a different location.

I think this is a great feature (that I think should be called something like `shift!` or `displace!`), but it's, in essence, different from what `taylor_expand` does. I can incorporate the `update!` feature with the name that we decide in this PR, but it will be defined via `evaluate` as you suggest and not via `taylor_expand`. What do you think @lbenet ?

Methods for evaluating `HomogeneousPolynomial` and `TaylorN` in `HomogenousPolynomial` and `TaylorN` are not yet defined. This should be constructed in order to implement the `update!` function accordingly. 

> Yet, I don't think that taylor_evaluate should change the number of variables nor the order for TaylorN; I think this should be done by the user explicitly, and taylor_evaluate should rely on get_order() and get_numvars().

The reason for this was that it took the dimension of the function `f` to expand it accordingly, but maybe it's a good practice for the user for him to change the `set_variables` environment a priori of the expansion. 

> Incidentally, the fact that tests are not passing may be related to changing the order and the number of variables; see above.

Output for travis (both Linux nightly & Mac nightly) is:
```
while loading /home/travis/.julia/v0.7/TaylorSeries/test/manyvariables.jl, in expression starting on line 7
Tests for HomogeneousPolynomial and TaylorN: Error During Test
  Test threw an exception of type MethodError
  Expression: taylor_expand(exp, [0, 0.0]) == exp.(set_variables("x", numvars=2))
  MethodError: no method matching exp(::Array{TaylorSeries.TaylorN{Float64},1})
  Closest candidates are:
    exp(!Matched::Float16) at math.jl:993
    exp(!Matched::Complex{Float16}) at math.jl:994
    exp(!Matched::BigFloat) at mpfr.jl:527
    ...
```
I think it has to do to broadcasting as `exp.(x)` still has some compatibility issues, but I'm not sure.

lbenet:
I agree that there is a difference between `update!` (or `displace!`, but not `shift!`, since it has a different meaning in Base) and `taylor_evaluate`; the former overwrittes the given `Taylor1`, while the latter creates a new one, but centered at some given value. In any case, implementing both is maybe the best.

For the tests, I think it's better to check the actual series constructed directly, either using `taylor_expand(f, x0, order)` or `as f(Taylor1([zero(x0),one(x0)],order))`; they should be identic.

Regarding the `TaylorN` case, we should perhaps leave it aside, to begin, and then latter address it.

blas-ko:
I've covered the case you're telling about @lbenet ! It works as
```julia
julia> using TaylorSeries

julia> xT, yT = set_variables("x y");

julia> f(x,y) = (x+y)^2

julia> f(xT,yT)
 1.0 x² + 2.0 x y + 1.0 y² + 𝒪(‖x‖⁷)

julia> taylor_expand(f, 0.0, 0.0)
1.0 x² + 2.0 x y + 1.0 y² + 𝒪(‖x‖⁷)

julia> @which taylor_expand(f, 0.0, 0.0)
taylor_expand(f::Function, x0...) at .../src/other_functions.jl:152
```
Additionally, `taylor_expand` for `TaylorN` throws a warning if the set variables are note equal to the required for the expansion.

blas-ko:
I edited my last comment, @lbenet. I'm not sure if Github notifies this... 
Cheers!

lbenet:
@blas-ko Can you rebase this branch to current master? This will also restart travis check and then I'll review it.

blas-ko:
Rebased! (Hope I did it correctly since I don't have experience rebasing yet).

src/evaluate.jl
```diff
@@ -137,6 +137,15 @@ function evaluate!{T<:NumberNotSeriesN}(x::Array{TaylorN{T},1}, δx::Array{Taylo
     nothing
 end
 
+function evaluate!{T<:NumberNotSeriesN}(x::Array{TaylorN{T},1}, δx::Array{TaylorN{T},1},
+        x0::Array{TaylorN{T},1})
+    @assert length(x) == length(x0)
+    @inbounds for i in eachindex(x)
+        x0[i] = evaluate( x[i], δx )
```
lbenet: Perhaps using here (as well in similar occurrences in other methods) `evaluate!(x[i], δx, x0[i])` instead, yields better performance.

src/evaluate.jl
```diff
@@ -261,6 +268,31 @@ function evaluate{T<:NumberNotSeries}(a::TaylorN{Taylor1{T}},
     return suma
 end
 
+function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
+        vals::Array{TaylorN{S},1})
+    @assert length(vals) == get_numvars()
+
+    num_vars = get_numvars()
+    ct = coeff_table
```
lbenet: This creates a copy of `coeff_table`. Perhaps using a `view` of it, or directly `coeff_table` saves this.

I didn't notice this before, sorry, but I think this should also be changed in other occurrences.

src/other_functions.jl
```diff
@@ -178,3 +179,70 @@ polynomials. For more details, see [`Base.isapprox`](@ref).
 function isapprox{T<:AbstractSeries,S<:AbstractSeries}(x::T, y::S; rtol::Real=rtoldefault(x,y), atol::Real=0, nans::Bool=false)
     x == y || (isfinite(x) && isfinite(y) && norm(x-y,1) <= atol + rtol*max(norm(x,1), norm(y,1))) || (nans && isnan(x) && isnan(y))
 end
+
+
+#taylor_expand function for Taylor1
+doc"""
+    taylor_expand(f,x0;order)
```
lbenet: Can you add white spaces? I mean `taylor_expand(f, x0; order)`.

As commented below, I am **not** in favor of changing the number of variables inside the function.

src/other_functions.jl
```diff
@@ -178,3 +179,70 @@ polynomials. For more details, see [`Base.isapprox`](@ref).
...
+
+Makes a Taylor expansion of the function `f` around the point `x0`. If x0 is a scalar,
+a `Taylor1` expansion will be done. If `x0` is a vector, a `TaylorN` expansion will be
+computed, changing the number of `TaylorN` variables
+according to the dimension of `f`.
+"""
+function taylor_expand(f::Function; order::Int64=15)
```
lbenet: Note that this will return a `Taylor1{Float64}` expansion. My point focuses in the `Float64` part.

src/other_functions.jl
```diff
@@ -178,3 +179,70 @@ polynomials. For more details, see [`Base.isapprox`](@ref).
...
+   return f(a)
+end
+
+function taylor_expand{T<:Number}(f::Function, x0::T; order::Int64=15)
+   a = Taylor1([x0, one(T)], order)
+   return f(a)
+end
+
+#taylor_expand function for TaylorN
+function taylor_expand{T<:Number}(f::Function, x0::Vector{T}; order::Int64=get_order()) #a Taylor expansion around x0
```
lbenet: I agree that, for `Taylor1` expansions it makes sense to allow the user to set the `order` of the expansion; I am not in favor of this for `TaylorN` expansions, at least as it is now. My objection is related to the use of `set_variables`, since it changes some internals, in particular, the *maximum* order of the `TaylorN` expansions. This costs time; also, the idea of the current implementation is to set the parameters related to TaylorN internals only once at the very beginning.

I think this method should assert that the the length of `x0` corresponds to `get_numvars()`, as you do it, and *only* when this is fulfilled evaluate `f(X .+ x0)` (notice the spaces). In this case, using `set_variables` will not change anything internally.

An alternative that avoids using `set_variables` is for example:
```julia
function taylor_expand{T<:Number}(f::Function, x0::Vector{T}; order::Int64=get_order())
    ll = length(x0)
    # check that x0 has the correct length and `order` does not exceed the maximum order of the expansions
    @assert ll == get_numvars() && order <= get_order()

    X = Array{TaylorN{T}}(length(x0)
    for i in eachindex(X)
        X[i] = TaylorN(T, i, order=order)
    end

    return f( X .+ x0 )
end
```

This alternative permits to have expansions of given `order`, smaller or equal to `get_order()`, e.g., `order=2` when `get_order()` is 6.

src/other_functions.jl
```diff
@@ -178,3 +179,70 @@ polynomials. For more details, see [`Base.isapprox`](@ref).
...
+        X = set_variables("x",order=order,numvars=ll)
+        warn("Changed number of TaylorN variables to $ll.")
+        end
+    return f(X.+x0)
+end
+
+function taylor_expand(f::Function, x0...; order::Int64=get_order()) #a Taylor expansion around x0
+    ll = length(x0)
+    ll == get_numvars() ? X = get_variables() : begin
+        X = set_variables("x",order=order,numvars=ll)
```
lbenet: Same comment as above.

src/other_functions.jl
```diff
@@ -178,3 +179,70 @@ polynomials. For more details, see [`Base.isapprox`](@ref).
...
+    return f(x0 .+ X...)
+end
+
+#taylor_expand! function for Taylor1
+doc"""
+    taylor_expand!(a,x0;order)
+
+Takes `a <: Union{Taylo1,TaylorN}` and expands it around the coordinate `x0`.
+"""
+function taylor_expand!(a::Taylor1)
```
lbenet: What about naming this function `update!` (as proposed in #80), and export it? Since this method is some sort of fallback, i think it can be written so `a::Union{Taylor1, TaylorN}`, which also covers another method below.

src/evaluate.jl
```diff
@@ -224,7 +233,6 @@ function evaluate{T<:Number,S<:NumberNotSeries}(a::TaylorN{T},
     suma = Taylor1(zeros(R, ord))
 
     for homPol in 1:length(a)
-        sun = zero(R)
```
lbenet: Good catch!

blas-ko:
Great! I've read all the changes and I'll make them as you propose. I wanted you to see the warning for `set_variables` in the `TaylorN` case, but I agree it's better to use an `@assert`. I'm gonna push this changes in a moment, so you can re-review.

src/evaluate.jl
blas-ko: Agreed for `evaluate` methods but not for `evaluate!` methods as `evaluate!`, as the docs say, just work for arrays.

lbenet: Good point!

lbenet:
Incidentally, I just noticed [this line](https://github.com/JuliaDiff/TaylorSeries.jl/pull/121/files#diff-06c3967d6023ae6130ae458e5fbe0cceL16), which seems to me unnecessary, since there is a method `real(T::Type)` (in Base at complex.jl:97) which does exactly this.

PerezHz:
> Incidentally, I just noticed this line, which seems to me unnecessary, since there is a method real(T::Type) (in Base at complex.jl:97) which does exactly this.

I added it in order to keep support for 0.5; once we drop 0.5 support, we can erase this line

src/other_functions.jl
blas-ko: Do you think `taylor_expand(T::Type,f::Function;order::Int64)` is a good idea? In this case,  should the same be done when passing an explicit `x0` to `taylor_expand`?

src/other_functions.jl
blas-ko: Really in favor of this. By the way, really good "type preserving strategy" doing the `X[i] = TaylorN(T, i, order)`.

blas-ko:
Seems that for Julia 0.5 broadcasting between `Array`s and `Tuple`s is not compatible... 
```
Tests for HomogeneousPolynomial and TaylorN: Error During Test
  Test threw an exception of type MethodError
  Expression: taylor_expand(f11,1.0,2.0) == taylor_expand(f22,[1,2.0])
  MethodError: no method matching .+(::Tuple{Float64,Float64}, ::Array{TaylorSeries.TaylorN{Float64},1})
```

This is, in particular, for line 216 of `src/other_functions.jl` for `function taylor_expand(f::Function, x0...; order::Int64=get_order()) ` method. 
What can be done here?

src/other_functions.jl
lbenet: Regarding the problem with travis in Julia 0.5 (and considering [your comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/121#issuecomment-334968066) about this), I think a possible solution is to have `X` as the shifted vector. I mean, define `X[i] = x0[i] + TaylorN(T, i, order=order)` *inside* the for-loop, and then only return the evaluation of `f( X )`. This could be used in each method of `taylor_expand`.

I think this could actually solve the problem with travis, and yield a tiny performance improvement, since only one loop would be needed.

src/other_functions.jl
```diff
@@ -178,3 +179,71 @@ polynomials. For more details, see [`Base.isapprox`](@ref).
...
+
+    for i in eachindex(X)
+        X[i] = TaylorN(T, i, order=order)
+    end
+
+    return f(X .+ x0)
+end
+
+function taylor_expand(f::Function, x0...; order::Int64=get_order()) #a Taylor expansion around x0
+    T = eltype(x0[1])
```
lbenet: Since `x0` is a tuple, each entry could have a different type. I think that `T` should be defined as `T = eltype( promote(x0...)[1] )`.

blas-ko:
Tests are passing!

lbenet:
Great! Thanks for all! I'm merging right away!

PerezHz:
Really cool stuff! Thanks for fixing #119 @blas-ko!

Issue #120, opened 2017-08-30
order is not correct when explicitly defining `Taylor1(Vector,order)`


blas-ko:
If one does
```julia
julia> using TaylorSeries;

julia> t = Taylor1([1.0,2,3,4,5],2)
 1.0 + 2.0 t + 3.0 t² + 4.0 t³ + 5.0 t⁴ + 𝒪(t³)

julia> length(t.coeffs), t.order
 (5,2)
```
notice the `𝒪(t³)` even when there are higher order terms in `t`. 

This could be treated in two ways:
+ Truncate the `Taylor1` object into the given `order`.
+ Extend the order of the `Taylor1` object into the lenght of the array of terms. 

I prefer the first solution because if you explicitly give the order to your polynomial, you would normally want to keep that order always. 

Feedback on this issue is welcomed; I can then do a PR based on what's discussed.

lbenet:
Thanks for reporting! I think this must have been introduced in 62d8184, or somewhere around, when the code was reorganized. I'll solve this.

lbenet:
See #122

lbenet:
Solved by #122. Closing.

PR #122, opened 2017-08-31
Solves an inconsistency when `order` is smaller than `length(coeffs)-1`

lbenet:
Solves an inconsistency when `order` is smaller than `length(coeffs)-1`.
Cc: @blas-ko, @dpsanders

### DIFF
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -11,10 +11,11 @@ is a basic polynomial algebraic manipulator in one or more
 variables; these two cases are treated separately.  Three new types are defined,
 [`Taylor1`](@ref), [`HomogeneousPolynomial`](@ref) and [`TaylorN`](@ref),
 which correspond to
-expansions in one independent variable, homogeneous polynomials of various variables,
-and the polynomial
+expansions in one independent variable, homogeneous polynomials of various
+variables, and the polynomial
 series in many independent variables, respectively. These types are subtypes
-of `AbstractSeries`, which in turn is a subtype of `Number`, and are defined parametrically.
+of `AbstractSeries`, which in turn is a subtype of `Number`, and are defined
+parametrically.
 
 The package is loaded as usual:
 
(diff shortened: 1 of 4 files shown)


lbenet:
Tests are passing; any comment @dpsanders ?

dpsanders:
I thought that adding an explicit `order` would force it to be of that order?

lbenet:
Originally, the maximum among the length of the vector of coefficients and the order was defining the actual order. Though never written, I think the rational was that if the user takes the time to provide "more coefficients", is because they are to be considered. At some point, when I rearranged the files, this behavior was kept, except that the `order` was not (internally) consistent with the length of the coefficients. This, then, looks odd when it is displayed; see #120.

dpsanders:
I guess that makes sense.
How do you truncate a `Taylor1` to a lower order then?

lbenet:
I think there is no method to do that, yet, though it would be easy to implement. For instance,
`truncate(x::Taylor1, order::Int) = Taylor1( x[1:order+1] )` would do the job.

lbenet:
Shall I include that in this PR?

blas-ko:
I think that if the user explicitly provides the coefficients, there's no need to provide the order of the polynomial, especially if the order is lower than the number of coefficients; that's why it exists a method `Taylor1(coeffs)` (wiithout `order` in it).

What I'm trying to say is that if you explicitly provide an order lower to the length of the coefficients, it is inferred that the user wants to truncate the polynomial, so I don't see the need to make an explicit `truncate` function if we follow this behaving.

dpsanders:
Yes, my feeling is that specifying the `order` explicitly should take preference. Then you can simply do something like
```
t = Taylor1([1,2,3,4,5])
Taylor1(t, order=2)
```

lbenet:
That's ok with me, and as @blas-ko  points out, this will actually behave as trunction, which means that the "added" coefficients should consistently be treated as zero.

Let me check how much would it break this change.

lbenet:
In particular, this may help (and perhaps also break) some components of TaylorIntegration...

lbenet:
It seems that this change does not break things, at least locally. Yet, even though it throws an error (related to `show`), the following `Taylor1([1,2,3], -1)` constructs a `Taylor1` object. An `@assert` could be used, but may slow things down a lot.

lbenet:
I've updated the docs and add few tests. I think this is ready to be merged. Waiting for the green lights in travis...

lbenet:
Tests are passing and I have updated the docs. I think this is ready to be merged.

lbenet:
@dpsanders I just realized that your suggestion included the keyword `order`. Should I revert the changes and implement your suggestion, or leave as it is now (which gives priority to `order` without using a keyword)?

dpsanders:
It's fine without the order keyword.

lbenet:
Thanks! I'm merging this...

Issue #127, opened 2017-09-26
Adding specific norm methods for polynomials


blas-ko:
I feel that `abs(Taylor)` has a different approach as `norm(Taylor,p)` in this Package. `abs`here is a function 
abs: P → P, where P is a Polynomial
where `abs(P) = P` if `constant_term(P)` > 0 and `abs(P) = -P` if `constant_term(P)` < 0 

I think `norm(Taylor)` should stricly return a non-negative value such that
$$ ||P||_p=(sum_(k=0)^n|a_k|^p)^(1/p) $$

where ||P||_ ∞ = max_k |a_k| 

Right now
```julia
using TaylorSeries

X,Y = set_variables("x y");
norm(X-1,2)
  1.0 - 1.0 x + 𝒪(‖x‖⁷)
norm(X-1,Inf)
  1.0 - 1.0 x + 𝒪(‖x‖⁷)
```
so I propose extending `norm` for Taylors in this way, in a sense we always get a non-negative number as a result.

PerezHz:
+1 for this idea! I think it is a good idea to be able to have a function, such as `norm`, which computes the [l-`p` norm of a polynomial](https://books.google.com.mx/books?id=wNTvBwAAQBAJ&lpg=PP1&pg=PA6#v=onepage&q&f=false). Actually, I have some work going in this direction; it relies on overloading `rtoldefault`, `isfinite`, and such functions from Julia `Base` for `Taylor1`, `HomogeneousPolynomial`, `TaylorN`, etc. This idea works at least on my machine, and would be happy to collaborate on a PR pursuing this.

Besides being related to #126, I think this kind of methods would make life much easier for comparing two polynomials, e.g., for testing in TaylorSeries and TaylorIntegration. 

Nevertheless, I understand there is a use case for `norm` as it is implemented now (@lbenet could clarify this), so we should think of being able to use both

blas-ko:
I've made `norm` work as proposed, but some of `Base` functions had to be overwritten. The whole thing goes like this (this is using the `nested_taylor` branch from #126 ): 
```julia
using TaylorSeries
import TaylorSeries: NumberNotSeries

x,y = set_nested_variables(2,get_order())

import Base: norm, abs2
norm{T<:Number}(a::Taylor1{T},p::Real) = norm(a.coeffs,p)
norm{T<:Number}(a::TaylorN{T},p::Real) = norm(a.coeffs,p)
[...]

PR #128, opened 2017-09-29
norm methods for AbstractSeries

blas-ko:
Follows discussion #127. `isapprox` is not yet added @PerezHz, but it may be easily included if we all agree with the `norm` methods here.


test/mixtures.jl
```diff
@@ -133,4 +133,10 @@ using Base.Test
     @test xx*δx + Taylor1(typeof(δx),5) == δx + δx^2 + Taylor1(typeof(δx),5)
     @test xx/(1+δx) == one(xx)
     @test typeof(xx+δx) == Taylor1{TaylorN{Float64}}
+
```
lbenet: Maybe you should convert `X` and/or `Y` to `Taylor1{TaylorN{Float64}}`s and include another test for that.

test/onevariable.jl
```diff
@@ -263,6 +263,17 @@ using Base.Test
     @test string(ta(0)^3-3) == " - 3 + 1 t³ + 𝒪(t¹⁶)"
     @test TaylorSeries.pretty_print(ta(3im)) == " ( 3 im )  + ( 1 ) t + 𝒪(t¹⁶)"
     @test string(Taylor1([1,2,3,4,5], 2)) == string(Taylor1([1,2,3]))
+
```
lbenet: My suggestion here it to test the `norm` of something like `t_C = complex(3.0,4.0) * t_a` against the `norm` of `complex(3.0,4.0) * a`; as defined below, `t_C` is only a constant Taylor series.

lbenet:
Do you understand why some tests fail in travis and others pass?

Incidentally, at some point you will have to rebase to current master.

lbenet:
I am restarting the tests; it may not be some glitch...

test/mixtures.jl
blas-ko: Great, I'll do this in a moment!

test/onevariable.jl
blas-ko: I'll change this too!

blas-ko:
> I am not sure if this implementation is enough to address this comment, without adding new methods for isapprox. Can you please check it?

Methods have to be extended for `AbstractSeries` as `isapprox` is defined specifically for each `Number` subtype. The way @PerezHz did it seems as the correct way to go. 

> Do you understand why some tests fail in travis and others pass?
No idea, honestly... 

> Incidentally, at some point you will have to rebase to current master.

Yes, I'll do your proposed changes and a rebase!

lbenet:
```julia
julia> methods(isapprox)
# 3 methods for generic function "isapprox":
isapprox(x::Number, y::Number; rtol, atol, nans) in Base at floatfuncs.jl:204
isapprox(x::AbstractArray, y::AbstractArray; rtol, atol, nans, norm) in Base.LinAlg at linalg/generic.jl:1280
isapprox(J1::UniformScaling{T}, J2::UniformScaling{S}; rtol, atol, nans) where {T<:Number, S<:Number} in Base.LinAlg at linalg/uniformscaling.jl:178
```

The reason I think we may get it for free (using the implementation of `norm`) is because of the first method above.

lbenet:
Tests are passing! It was a travis thing...

PerezHz:
> The reason I think we may get it for free (using the implementation of norm) is because of the first method above.

Julia 0.6 code for the relevant `isapprox` method that you mention reads (JuliaLang/julia/src/floatfuncs.jl, line 204):

```julia
function isapprox(x::Number, y::Number; rtol::Real=rtoldefault(x,y), atol::Real=0, nans::Bool=false)
    x == y || (isfinite(x) && isfinite(y) && abs(x-y) <= atol + rtol*max(abs(x), abs(y))) || (nans && isnan(x) && isnan(y))
end
```

Note the use of the `abs` function in the expression `abs(x-y) <= atol + rtol*max(abs(x), abs(y)))` above.

Using this method, if on this branch (`blas-ko:norm`) we do

```julia
using TaylorSeries

import Base: rtoldefault, isfinite

# rtoldefault ans isfinite overloading for TaylorSeries
rtoldefault(::Type{T}) where T<:AbstractSeries = sqrt(eps())
isfinite(x::AbstractSeries) = !isinf(x)

t = Taylor1(25)
p = sin(t)
q = sin(t+eps())
```

then `p ≈ p` returns `true`, `q ≈ q` returns `true`, but `q ≈ p` returns the following error:

```julia
ERROR: ArgumentError: The 0th order Taylor1 coefficient must be non-zero
(abs(x) is not differentiable at x=0).
Stacktrace:
 [1] abs(::TaylorSeries.Taylor1{Float64}) at /Users/Jorge/fork-bo/TaylorSeries.jl/src/other_functions.jl:86
 [2] #isapprox#399(::Float64, ::Int64, ::Bool, ::Function, ::TaylorSeries.Taylor1{Float64}, ::TaylorSeries.Taylor1{Float64}) at ./floatfuncs.jl:204
 [3] isapprox(::TaylorSeries.Taylor1{Float64}, ::TaylorSeries.Taylor1{Float64}) at ./floatfuncs.jl:204
```

Which is related to the fact that `abs` is not defined at zero for Taylor1 polynomials. So there are two solutions that I can think of:

1) overloading `isapprox` for TaylorSeries polynomials
2) re-defining `abs` in TaylorSeries

On another comment, I proposed an overloading of `isapprox` for TaylorSeries polynomials, which I print here for reference:

```julia
function isapprox{T<:AbstractSeries}(x::T, y::T; rtol::Real=rtoldefault(x,y), atol::Real=0, nans::Bool=false)
    x == y || (isfinite(x) && isfinite(y) && norm(x-y,1) <= atol + rtol*max(norm(x,1), norm(y,1))) || (nans && isnan(x) && isnan(y))
end
```
Note that this is essentially the same method as `Base`'s, except for the use of `norm(...,1)` instead of `abs(...)`. The reasoning behind this is that, `Base.isapprox` relies on `abs` returning a `<:Real` value. Further, in order for things to work properly, `abs` for TaylorSeries polynomials should fulfill the properties of a notion of distance between polynomials, which in turn should reduce to the notion of distance between real numbers for 0-th order polynomials. The current `abs` method in TaylorSeries doesn't fulfill these properties; but `norm(...,1)`, as is currently defined in this PR, does.

So, if on this branch we do:

```julia
using TaylorSeries

import Base: rtoldefault, isfinite, isapprox

# rtoldefault, isfinite and isapprox overloading for TaylorSeries
rtoldefault(::Type{T}) where T<:AbstractSeries = sqrt(eps())
isfinite(x::AbstractSeries) = !isinf(x)

function isapprox{T<:AbstractSeries}(x::T, y::T; rtol::Real=rtoldefault(x,y), atol::Real=0, nans::Bool=false)
    x == y || (isfinite(x) && isfinite(y) && norm(x-y,1) <= atol + rtol*max(norm(x,1), norm(y,1))) || (nans && isnan(x) && isnan(y))
end

t = Taylor1(25)
p = sin(t)
q = sin(t+eps())

```

Then `p ≈ p, q ≈ q, p ≈ q` returns `(true, true, true)`.

The second option, re-defining `abs` in TaylorSeries, is what would be closest to @lbenet suggestion (getting `isapprox` "for free") and would be viable as well as far as this PR goes, but also would represent giving `abs` a whole different meaning in TaylorSeries. Still, if we pursue this idea, removing the current `abs` methods in TaylorSeries, and then doing

```julia
using TaylorSeries

import Base: rtoldefault, isfinite, abs

# rtoldefault, isfinite and abs overloading for TaylorSeries
rtoldefault(::Type{T}) where T<:AbstractSeries = sqrt(eps())
isfinite(x::AbstractSeries) = !isinf(x)
abs{T<:AbstractSeries}(x::T) = norm(x,1)

t = Taylor1(25)
p = sin(t)
q = sin(t+eps())
```

then, again, `p ≈ p, q ≈ q, p ≈ q` returns `(true, true, true)`. I'd actually be more inclined towards this last option, but what do you guys think?

> isapprox is not yet added @PerezHz, but it may be easily included if we all agree with the norm methods here.

I'm happy with the `norm` methods in this PR!

src/other_functions.jl
```diff
@@ -126,3 +126,7 @@ function abs{T<:Real}(a::Taylor1{TaylorN{T}})
         (abs(x) is not differentiable at x=0)."""))
     end
 end
+
+#norm
+norm(x::AbstractSeries, p::Real=2) = norm( norm.(x.coeffs, p), p)
+norm{T<:NumberNotSeries}(x::Taylor1{T}, p::Real=2) = norm(x.coeffs, p)
```
PerezHz: Cool!👌

lbenet:
You are right @PerezHz, we do need specialized methods for `isapprox`. 

I would then suggest to add somewhere in the documentation the distinction: `abs` returns a polynomial, while `norm` a non-negative number.

lbenet:
Not sure what happened, but now no tests pass...

blas-ko:
I tried to add documentation for `abs` but I think I did it in the wrong place... Can you check this, @lbenet ?

blas-ko:
Note that I added @PerezHz `isapprox` methods from [here](https://github.com/blas-ko/TaylorSeries.jl/pull/1)  in.

lbenet:
I think the problem with travis is precisely related with the docstrings of `abs`: You should add the docstrings *before* the function (when you code it); alternatively, define the function and, later, you add the docstrings followed by the function name. See for instance the docstrings for [`exp!`](https://github.com/JuliaDiff/TaylorSeries.jl/blob/3b68fb3cff4354f7e2a4994216c0f3b4d8e7abe3/src/functions.jl#L364), or other functions there.

src/other_functions.jl
```diff
@@ -134,6 +129,13 @@ function abs{T<:Real}(a::Taylor1{TaylorN{T}})
         (abs(x) is not differentiable at x=0)."""))
     end
 end
+"""
+    abs(a::Union{Taylor1,TaylorN})
+
+Returns `a` if `constant_term(a) > 0` and `-a` if `constant_term(a) < 0`.
+Notice that `typeof(a) <: AbstractSeries`.
+"""
```
lbenet: You will still need to add `abs` after the (second) `"""`!

blas-ko: Oh! I'm sorry! I didn't remember that!

lbenet:
Perhaps it is better to add the docstrings of `abs` where `abs` is coded.

src/other_functions.jl
```diff
@@ -126,3 +129,50 @@ function abs{T<:Real}(a::Taylor1{TaylorN{T}})
         (abs(x) is not differentiable at x=0)."""))
     end
 end
+"""
+    abs(a)
+
+Returns `a` if `constant_term(a) > 0` and `-a` if `constant_term(a) < 0` for
+`a <:Union{Taylor1,TaylorN}`.
+Notice that `typeof(abs(a)) <: AbstractSeries`.
+"""abs
```
PerezHz: I think there's a missing space between the `"""` and `abs`

src/other_functions.jl
```diff
@@ -126,3 +129,50 @@ function abs{T<:Real}(a::Taylor1{TaylorN{T}})
...
+
+
+#norm
+"""
+    norm(x::AbstractSeries,p::Real)
+
+Computes the p-norm of an `AbstractSeries` defined by $P(\vec{x}) = \sum_k a_k \vec{x}^k$ as
```
PerezHz: Should the `$` signs be escaped?

PerezHz:
> Note that I added @PerezHz isapprox methods from here in.

Great! Thanks!

There seem to be still two errors which may be easily fixed, I've already commented on the relevant lines

src/other_functions.jl
PerezHz: What seems to work best, at least locally, is to use `doc` before the opening `"""` quotes, and then write the math expression as ``` ``P(\vec{x}) = \sum_k a_k \vec{x}^k`` ```

blas-ko:
Strange Travis error. Can you check this? 
Sorry for all the trouble with this! I promise I'll learn xD

PerezHz:
> Strange Travis error. Can you check this?

Try substituting occurrences of `$` for ``` `` ``` on line 146 of `src/other_functions.jl`

PerezHz:
> Anyways, let's see what happens and hope this is the solution.

Well, it seems that now most tests in travis are passing, except for one, which seems to be related to a breaking change in julia 0.7.0-DEV on OS X; namely, `Base.Test` is now a "standard library package", and must be called with the syntax `using Test` instead of `using Base.Test`; don't know if Compat.jl already handles this, there's another way to solve it, or simply accept for the time being failures on julia nightlies?

> Thanks!

Sure, no problem!

PerezHz:
> except for one, which seems to be related to a breaking change in julia 0.7.0-DEV on OS X

About this one failing test, I was able to reproduce locally this failure on julia 0.7.0-DEV.2036. Upon replacing `Base.Test` -> `Test`, the failure reported by travis disappeared, although new problems appear, unrelated to this PR, but rather related to changes in julia 0.7.0-DEV:

There are 4 errors like this one:

```julia
Tests for HomogeneousPolynomial and TaylorN: Error During Test
  Test threw an exception of type UnicodeError
  Expression: string(-xH) == " - 1 x₁"
  UnicodeError: invalid character index
  Stacktrace:
   [1] getindex(::String, ::UnitRange{Int64}) at ./strings/string.jl:292
   [2] homogPol2str(::TaylorSeries.HomogeneousPolynomial{Int64}) at /Users/Jorge/fork-bk/TaylorSeries.jl/src/printing.jl:135
   [3] pretty_print(::TaylorSeries.HomogeneousPolynomial{Int64}) at /Users/Jorge/fork-bk/TaylorSeries.jl/src/printing.jl:87
   [4] show at /Users/Jorge/fork-bk/TaylorSeries.jl/src/printing.jl:231 [inlined]
   [5] print(::Base.GenericIOBuffer{Array{UInt8,1}}, ::TaylorSeries.HomogeneousPolynomial{Int64}) at ./strings/io.jl:30
   [6] #print_to_string#247(::Void, ::Function, ::TaylorSeries.HomogeneousPolynomial{Int64}, ::Vararg{TaylorSeries.HomogeneousPolynomial{Int64},N} where N) at ./strings/io.jl:117
   [7] macro expansion at /Users/Jorge/fork-bk/TaylorSeries.jl/test/manyvariables.jl:366 [inlined]
   [8] macro expansion at /Users/Jorge/julia/usr/share/julia/site/v0.7/Test/src/Test.jl:942 [inlined]
   [9] anonymous at ./<missing>:?
   [10] include_relative(::Module, ::String) at ./loading.jl:526
   [11] include(::Module, ::String) at ./sysimg.jl:14
   [12] include(::String) at ./sysimg.jl:47
   [13] macro expansion at /Users/Jorge/fork-bk/TaylorSeries.jl/test/runtests.jl:14 [inlined]
   [14] anonymous at ./<missing>:?
   [15] include_relative(::Module, ::String) at ./loading.jl:526
   [16] include(::Module, ::String) at ./sysimg.jl:14
   [17] process_options(::Base.JLOptions) at ./client.jl:325
   [18] _start() at ./client.jl:391
```

PerezHz:
For reference, the change `Base.Test` -> `Test` was introduced in JuliaLang/julia#23876

lbenet:
I just restarted the failing travis job; what it surprises me is that Julia 0.7 tests passed in linux, but not in mac.

PerezHz:
> I just restarted the failing travis job;
Thanks! 

> what it surprises me is that Julia 0.7 tests passed in linux, but not in mac.

I think that's due to the fact that the julia version on linux is prior to the Base.Test->test thing, and the julia version for mac isn't

travis job on linux:
```
$ julia -e 'versioninfo()'
Julia Version 0.7.0-DEV.1775
Commit 55ec9fe (2017-09-13 21:22 UTC)
```

travis job on mac:
```
$ julia -e 'versioninfo()'
Julia Version 0.7.0-DEV.2036
Commit 6c4e30e367 (2017-10-02 16:24 UTC)
```

Just noticed that the travis job that you restarted, job \#675.6, failed again:

```
WARNING: Base.Test is deprecated, use Base.Test instead.
  likely near /Users/travis/.julia/v0.7/TaylorSeries/test/onevariable.jl:5
ERROR: LoadError: LoadError: LoadError: Base.Test.@testset has been moved to the standard library package Test.
Restart Julia and then run `using Test` to load it.
```

lbenet:
If the problem is related with `Base.Test` or `Test`, depending if we are in 0.6 or 0.7-dev, I suggest to add the following in `test/runtests.jl`:
```julia
if VERSION < v"0.7"
    using Base.Test
else
    using Test
end
```

PerezHz:
Sounds good! Using the `commit-name.sh` script, I was able to determine that the breaking change was introduced in `v"0.7.0-DEV.2004"`, so perhaps we could do:

```julia
if VERSION < v"0.7.0-DEV.2004"
    using Base.Test
else
    using Test
end
```

blas-ko:
This sounds good! Should I push this commit into this PR to see if Travis passes? Or should this be added into a new PR?

blas-ko:
Travis failed again. Nevertheless, this time error log shows something different.  There are tests that are not passing for Mac's `julia:nightly` as @PerezHz  noticed 2 days ago. Error log shows 
```
Tests for HomogeneousPolynomial and TaylorN: Error During Test
  Test threw an exception of type UnicodeError
  Expression: string(-xH) == " - 1 x₁"
  UnicodeError: invalid character index
```
and other 3 similar errors with `UnicodeError: invalid character index`.

blas-ko:
> My proposal is to merge this and address in another PR the compatibility issues related to mac.

I agree!

dpsanders:
I don't know what's going on. I only saw that there have been certain issues with Mac on Travis, but haven't been following the discussion, sorry. You might want to write ask on discourse.julialang.org.

PerezHz:
Just as a side note, since the last mac and linux builds were done with different julia nightlies, the difference may also be due to recent changes in julia 0.7.0-DEV:

travis linux julia version:
```
$ julia -e 'versioninfo()'
Julia Version 0.7.0-DEV.1775
Commit 55ec9fe (2017-09-13 21:22 UTC)
```

travis mac julia version:
```
$ julia -e 'versioninfo()'
Julia Version 0.7.0-DEV.2046
Commit 16139d054e (2017-10-04 06:00 UTC)
```

I agree to merge this and work out the nightlies failures for mac on a new PR!

Also, there's some discussion going on in JuliaLang/julia#23983 about the `Base.Test`->`Test` change; perhaps after that PR is merged and the travis julia versions are updated, in a couple of days  (weeks?) it'll be possible to do `using Base.Test` as deprecated syntax.

Meanwhile, I'll try to work out an MWE regarding the `UnicodeError` with strings and post a question to Discourse...

PerezHz:
Can confirm that, using julia version `0.7.0-DEV.1775`, on my machine (OS X/macOS 10.10.5 Yosemite), `TaylorSeries` tests do pass; but using julia version `0.7.0-DEV.2046`, they do not

test/manyvariables.jl
```diff
@@ -333,6 +337,36 @@ using Base.Test
     hessian!(hes2,g1(xT+1,yT-1)-g2(xT+1,yT-1))
     @test hes1 == hes2
 
+    a = 3x + 4y +6x^2 + 8x*y
+    @test typeof( norm(x) ) == Float64
+    @test norm(x) > 0
+    @test norm(a) == norm([3,4,6,8.0])
+    @test norm(a,4) == sum([3,4,6,8.0].^4)^(1/4.)
+    @test norm(a,Inf) == 8.
+    @test norm((3.+4im)*x) == abs(3.+4im)
```
lbenet: I am checking what is going on with travis for julia 0.7-dev on mac, and on doing this, I just noticed that julia 0.7 complains about `3.+4im` as ambiguous (due to parsing conflicts, essentially to know if you mean `3. + ...` or `3 .+ ...`). I suggest to write it as `3.0 + 4im`. Can you change it? See https://github.com/JuliaLang/julia/issues/19089.

lbenet:
I've been checking the problem with the subscripts in julia 0.7 in a mac, and it seems the problem is with printing things like `x₁`, which is done programmatically. I sent a comment on something small related with this PR to @blas-ko ; once it is addressed (no need to correct this later), I'll merge this and see if I can solve the problem in 0.7.

Still, I don't understand why it works in linux but it doesn't in mac...

PerezHz:
> Still, I don't understand why it works in linux but it doesn't in mac...

I think the different behavior may be due to different julia versions in linux vs mac on travis

lbenet:
I restarted the linux nighties job...

PerezHz:
I think I was able to get a MWE of the `UnicodeError` problem and a solution, thanks to @Keno!

Using julia version `0.7.0-DEV.2046` I tried:

```julia
julia> str1 = "a "
"a "

julia> str2 = "a"
"a"

julia> str2 == str1[1:end-1] # works fine
true

julia> str1 = "₁ "
"₁ "

julia> str2 = "₁"
"₁"

julia> str2 == str1[1:end-1] # Unicode character; throws error
ERROR: UnicodeError: invalid character index
Stacktrace:
 [1] getindex(::String, ::UnitRange{Int64}) at ./strings/string.jl:292
```

So, as @dpsanders suggested, I asked on Discourse if this was a known issue; turns out there's a breaking change introduced by JuliaLang/julia#22572, which corresponds to julia version `0.7.0-DEV.1864`. This change is actually documented in the [Breaking changes](https://github.com/JuliaLang/julia/blob/master/NEWS.md#breaking-changes) section of `NEWS.md`. As was pointed out by @Keno on [Discourse](https://discourse.julialang.org/t/problems-with-strings-which-contain-unicode-characters-on-julia-0-7-dev/6279), this problem can be easily fixed by doing:

```julia
julia> str1 = "₁ "
"₁ "

julia> str2 = "₁"
"₁"

julia> str2 == str1[1:prevind(str1, end)]
true
```

So the issue should be solved if, on line 135 of `src/printing.jl` we did `return strout[1:prevind(strout, end)]` instead of `return strout[1:end-1]`... Just tried that change locally (macOS 10.10.5, julia `0.7.0-DEV.2046`) and it works!

lbenet:
Thanks a lot for the fix @PerezHz ! I also found a less elegant but equivalent solution; yours is simply better. Can you push the fix, and also address [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/128#pullrequestreview-67543463) ?

PerezHz:
Sure! Just pushed the requested changes to @blas-ko's fork of TaylorSeries, where this PR is hosted

blas-ko:
Just merged @PerezHz PR which addresses `3.` -> `3.0` and string problem using `[1:prevind(str1, end)]`. Hope this finally solves everything.

lbenet:
Let's see how travis reacts!

PerezHz:
Green lights!

lbenet:
Merging! Thanks a lot to all for contributing to this!

PR #131, opened 2017-10-31
Declared new method for get_variables.

blas-ko:
This just add a method `get_variables(order::Int)` where `order <  get_order()` so lower order `TaylorN`s can be taken without changing internals.
If `order >  get_order()`, an `AssertionError` is thrown. 
```julia
julia> using TaylorSeries

julia> get_variables()
  1.0 x₁ + 𝒪(‖x‖⁷)
  1.0 x₂ + 𝒪(‖x‖⁷)

julia> get_variables(2)
  1.0 x₁ + 𝒪(‖x‖³)
  1.0 x₂ + 𝒪(‖x‖³)

julia> get_variables(10)
ERROR: AssertionError: order ≤ get_order() && lencoef ≤ num_coeffs
```

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -27,6 +27,7 @@ Taylor1([::Type{Float64}], [order::Int64=1])
 HomogeneousPolynomial{T<:Number}(::Type{T}, ::Int)
 TaylorN{T<:Number}(::Type{T}, nv::Int; [order::Int=get_order()])
 set_variables
+get_variables
 show_params_TaylorN
 get_coeff
 evaluate
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -211,6 +211,16 @@ variable name and the optional keyword argument `numvars`:
 set_variables("α", numvars=3)
 ```
 
+Alternatively to `set_variables`, [`get_variables`](@ref) can be used if one
+doesn't want to change internal dictionaries. `get_variables` returns a vector
+of independent variables  of a desired `order` (lesser than `get_order` so
+internals doesn't have to change) with the length and variable names defined
+by `set_variables` initially.
+
+```@repl userguide
+get_variables(order=2) #vector of independent variables of order 2
+```
+
 The function [`show_params_TaylorN`](@ref) displays the current values of the
 parameters, in an info block.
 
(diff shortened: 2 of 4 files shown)


lbenet:
I think I see the motivation for this :-) I do have one question and one comment, though.

- What do you get for the following:  `x, y = get_variables(2); x^3`? (Default order I think is 6.) I would expect that this yields zero, since you are over the order of `x`, but I am not sure.

- I would suggest to use keyword arguments here to specify the reduced order; that is, `get_variables()` yields the default (so `order=get_order()` is the default), and `get_variables(order=2)` returns the vector of `TaylorN`s, of order 2.

blas-ko:
> What do you get for the following: x, y = get_variables(2); x^3? (Default order I think is 6.) I would expect that this yields zero, since you are over the order of x, but I am not sure.

It is, indeed, zero:
```julia
julia> x,y = get_variables(2)
  1.0 x₁ + 𝒪(‖x‖³)
  1.0 x₂ + 𝒪(‖x‖³)

julia> x^3
 0.0 + 𝒪(‖x‖³)

julia> y^3
 0.0 + 𝒪(‖x‖³)
```

> I would suggest to use keyword arguments here to specify the reduced order; that is, get_variables() yields the default (so order=get_order() is the default), and get_variables(order=2) returns the vector of TaylorNs, of order 2.

Completely agree with this!! Originally that was my intention, but I didn't want to change the original `get_variables` method. If you agree, I'll make the `get_variables(;order::Integer=get_order())` method right away.

**NOTE:** Should I include any test? I didn't because `get_variables()` doesn't have any specific test, but please tell me if something should be added.

lbenet:
LGTM. 

Regarding the docs, I was thinking to mention the additions in `docs/userguide.md` somewhere around the description of `set_variables()`. This will be reflected in the "latest docs", and when we tag a new version, they will appear in the "stable docs".

Maybe @dpsanders has other comments or suggestions...

blas-ko:
I added some lines to `docs/src/userguide.md` please check this and tell me if you agree with syntax.

lbenet:
LGTM. Thanks! I'm merging.

Issue #133, opened 2017-11-03
Start indexing at 0


dpsanders:
The `Polynomials.jl` package starts indexing at `0`, so that `p[0]` is the 0th order term, `p[1]` is the first-order term, etc. (see below).

I suggest changing to this behaviour.

```jl
julia> using Polynomials

julia> p = Poly([1, 2, 3])
Poly(1 + 2*x + 3*x^2)

julia> p[0]
1

julia> p[1]
2

julia> p[2]
3

julia> p[3]
0
```

lbenet:
I think this is easy to implement (a simple shift in the definition), and may be it is good to stick to the same convention. But it is a breaking change, thinking simply on `TaylorIntegration.jl`.

One comment: currently, the equivalent of your example with `TaylorSeries.jl` (putting aside the issue on the 0-index or 1-index) yields the following:
```julia
julia> using TaylorSeries

julia> q = Taylor1([1,2,3])
 1 + 2 t + 3 t² + 𝒪(t³)

julia> q[1]
1

julia> q[4]
ERROR: BoundsError: attempt to access 3-element Array{Int64,1} at index [4]
Stacktrace:
 [1] getindex(::TaylorSeries.Taylor1{Int64}, ::Int64) at /Users/benet/.julia/v0.6/TaylorSeries/src/auxiliary.jl:77
```
The point I want to stress is the fact that `q[4]` throws an error, instead of being zero. I think we discussed this somewhere, and it was decided that way, but i don't recall the discussion.

Cc: @PerezHz

lbenet:
While it is trivially a shift, it matters a lot internally!

PerezHz:
I agree that this change would conform even more TaylorSeries syntax to standard mathematical notation, so I'd say let's go for it! Perhaps even more importantly, this change allows the user (or at least me) to "save" one mental step when dealing with TaylorSeries polynomials. So, even if that means changing a lot of internal stuff both in TaylorSeries and TaylorIntegration, I think it's worthwhile to implement this!

lbenet:
Solved by #135. Closing

PR #135, opened 2017-11-05
Change indexing to start from zero

lbenet:
Fixes issue #133, to start indexing from 0.

### DIFF
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -301,15 +301,18 @@ get_coeff(exy, [3,5])
 rationalize(ans)
 ```
 
-Similar to `Taylor1`, using vector notation can be used to request specific
-coefficients of `HomogeneousPolynomial` or `TaylorN` objects. In the case
-of `HomogeneousPolynomial` the index refers to the position of the hash
-table, and therefore, the counting starts at 1. For `TaylorN` objects it
-refers to the degree of the `HomogeneousPolynomial`.
+Similar to `Taylor1`, vector notation can be used to request specific
+coefficients of `HomogeneousPolynomial` or `TaylorN` objects. For `TaylorN`
+objects, the index refers to the degree of the `HomogeneousPolynomial`.
+In the case of `HomogeneousPolynomial` the index refers to the position
+of the hash table. The function [`show_monomials`](@ref) can be used to
+obtain the coefficient a specific monomial, given the degree of the
+`HomogeneousPolynomial`.
 
 ```@repl userguide
-exy[2] # get the 2th order term
-exy[2][1] # get the 1st coeff of the 2th order term
+exy[8] # get the 8th order term
+show_monomials(8)
+exy[8][6] # get the 1st coeff of the 2th order term
 ```
 
 Partial differentiation is also implemented for [`TaylorN`](@ref) objects,
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -31,7 +31,8 @@ order considered for the expansion (fieldname `order`). The
 coefficients are arranged in ascending order with respect to the degree of the
 monomial, so that
 `coeffs[1]` is the constant term, `coeffs[2]` gives the first order term (`t^1`),
-etc. This is a dense representation of the polynomial.
+etc. Yet, it is possible to have the natural ordering with respecto
+to the degree; see below. This is a dense representation of the polynomial.
 The order of the polynomial can be
 omitted in the constructor, which is then fixed by the length of the
 vector of coefficients. If the length of the vector does not correspond with
@@ -105,6 +106,16 @@ Again, errors are thrown whenever it is necessary.
 ```@repl userguide
 sqrt(t)
 log(t)
+```
+
+To obtain a specific coefficient, [`get_coeff`](@ref) can be used. Another
+alternative is to request the specific degree using the vector notation,
+where the index corresponds to the degree of the term.
+
+```@repl userguide
+expon = exp(t)
+get_coeff(expon, 0) == expon[0]
+rationalize(expon[3])
 ```
 
 Differentiating and integrating is straightforward for polynomial expansions in
@@ -288,6 +299,17 @@ the polynomial `exy` above, the coefficient of the monomial ``x^3 y^5`` is
 ```@repl userguide
 get_coeff(exy, [3,5])
 rationalize(ans)
+```
+
+Similar to `Taylor1`, using vector notation can be used to request specific
+coefficients of `HomogeneousPolynomial` or `TaylorN` objects. In the case
+of `HomogeneousPolynomial` the index refers to the position of the hash
+table, and therefore, the counting starts at 1. For `TaylorN` objects it
+refers to the degree of the `HomogeneousPolynomial`.
+
+```@repl userguide
+exy[2] # get the 2th order term
+exy[2][1] # get the 1st coeff of the 2th order term
 ```
 
 Partial differentiation is also implemented for [`TaylorN`](@ref) objects,
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -31,7 +31,7 @@ order considered for the expansion (fieldname `order`). The
 coefficients are arranged in ascending order with respect to the degree of the
 monomial, so that
 `coeffs[1]` is the constant term, `coeffs[2]` gives the first order term (`t^1`),
-etc. Yet, it is possible to have the natural ordering with respecto
+etc. Yet, it is possible to have the natural ordering with respect
 to the degree; see below. This is a dense representation of the polynomial.
 The order of the polynomial can be
 omitted in the constructor, which is then fixed by the length of the
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -108,13 +108,13 @@ sqrt(t)
 log(t)
 ```
 
-To obtain a specific coefficient, [`get_coeff`](@ref) can be used. Another
+To obtain a specific coefficient, [`getcoeff`](@ref) can be used. Another
 alternative is to request the specific degree using the vector notation,
 where the index corresponds to the degree of the term.
 
 ```@repl userguide
 expon = exp(t)
-get_coeff(expon, 0) == expon[0]
+getcoeff(expon, 0) == expon[0]
 rationalize(expon[3])
 ```
 
@@ -291,13 +291,13 @@ x, y = set_variables("x y", order=10);
 exy = exp(x+y)
 ```
 
-The function [`get_coeff`](@ref)
+The function [`getcoeff`](@ref)
 gives the normalized coefficient of the polynomial that corresponds to the
 monomial specified by a vector `v` containing the powers. For instance, for
 the polynomial `exy` above, the coefficient of the monomial ``x^3 y^5`` is
 
 ```@repl userguide
-get_coeff(exy, [3,5])
+getcoeff(exy, [3,5])
 rationalize(ans)
 ```
 
@@ -312,7 +312,7 @@ obtain the coefficient a specific monomial, given the degree of the
 ```@repl userguide
 exy[8] # get the 8th order term
 show_monomials(8)
-exy[8][6] # get the 1st coeff of the 2th order term
+exy[8][6] # get the 6th coeff of the 8th order term
 ```
 
 Partial differentiation is also implemented for [`TaylorN`](@ref) objects,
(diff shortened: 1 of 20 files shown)


lbenet:
This PR addresses #133; so far, only Taylor1 has been addressed.

lbenet:
Rebased to current master...

By some reason, tests pass locally but they don't with travis. Let's see what happens now.

lbenet:
Travis is driving me crazy...

Anyway, I just added `show_monomials(ord)`, which displays the indices of the `HomogeneousPolynomial`s and relate them to the actual monomial.
```julia
julia> show_monomials(3)
 1  -->  x³
 2  -->  x² y
 3  -->  x y²
 4  -->  y³
```

So know, if `p` is a `TaylorN` polynomial in two variables, then `p[3][2]` is the coefficient of the monomial x² y.

dpsanders:
It would be useful to have a function that returns the variables and indices, not just displays them.

lbenet:
I'm not sure I understand you. What do you mean by variables and indices? Is `get_variables()` what you are looking for?

The indices being displayed are precisely the internal indices used for each monomial. So the coefficient of the monomial x y² can be obtained by `get_coeff( a, [1,2] )`, or as `a[3][3]`.

lbenet:
Despite my complains, I must be thankful to travis and the "--check-bounds=yes" flag, that spotted two very subtle problems... I hope tests pass now.

lbenet:
Tests are passing...

docs/src/userguide.md
```diff
@@ -31,7 +31,8 @@ order considered for the expansion (fieldname `order`). The
 coefficients are arranged in ascending order with respect to the degree of the
 monomial, so that
 `coeffs[1]` is the constant term, `coeffs[2]` gives the first order term (`t^1`),
-etc. This is a dense representation of the polynomial.
+etc. Yet, it is possible to have the natural ordering with respecto
```
PerezHz: with respect~o~

lbenet:
Good catch @PerezHz! I'll correct it!

@dpsanders Shall I change in some way `show_monomials()`? This is ready for review...

lbenet:
@dpsanders Any comments on this?

docs/src/userguide.md
```diff
@@ -290,6 +301,20 @@ get_coeff(exy, [3,5])
...
+objects, the index refers to the degree of the `HomogeneousPolynomial`.
+In the case of `HomogeneousPolynomial` the index refers to the position
+of the hash table. The function [`show_monomials`](@ref) can be used to
+obtain the coefficient a specific monomial, given the degree of the
+`HomogeneousPolynomial`.
+
+```@repl userguide
+exy[8] # get the 8th order term
+show_monomials(8)
+exy[8][6] # get the 1st coeff of the 2th order term
```
dpsanders: What about `exy[8, 6]`?

src/arithmetic.jl
```diff
@@ -23,16 +23,18 @@ for T in (:Taylor1, :TaylorN)
     end
 end
 
-==(a::HomogeneousPolynomial, b::HomogeneousPolynomial) = a.coeffs == b.coeffs
-
+function ==(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
+    a.order == b.order && return a.coeffs == b.coeffs
+    return (a.coeffs == zeros(a.coeffs)) && (b.coeffs == zeros(b.coeffs))
```
dpsanders: Use `iszero`?

dpsanders: Why is this change necessary?

src/arithmetic.jl
```diff
@@ -225,7 +227,7 @@ for T in (:HomogeneousPolynomial, :TaylorN)
         function *(a::Taylor1{T}, b::$T{Taylor1{S}}) where
                 {T<:NumberNotSeries, S<:NumberNotSeries}
 
-            @inbounds aux = a * b[1]
+            @inbounds aux = a * b.coeffs[1]
```
dpsanders: `b[0]`?

src/arithmetic.jl
```diff
@@ -382,7 +384,7 @@ end
 for T in (:HomogeneousPolynomial, :TaylorN)
     @eval function /(b::$T{Taylor1{S}}, a::Taylor1{T}) where
             {T<:NumberNotSeries,S<:NumberNotSeries}
-        @inbounds aux = b[1] / a
+        @inbounds aux = b.coeffs[1] / a
```
dpsanders: `b[0]`?

src/arithmetic.jl
```diff
@@ -500,14 +502,14 @@ div!(v::Taylor1, b::NumberNotSeries, a::Taylor1, k::Int) =
 
 @inline function div!(c::TaylorN, a::TaylorN, b::TaylorN, k::Int)
     if k==0
-        @inbounds c[1] = a[1] / constant_term(b)
+        @inbounds c[0] = a[0] / constant_term(b)
```
dpsanders: Is `constant_term(b)` not just `b[0]`?

src/auxiliary.jl
```diff
@@ -121,55 +138,82 @@ Return the coefficient of `a::TaylorN`, specified by
 function get_coeff(a::TaylorN, v::Array{Int,1})
```
dpsanders: Rename to `getcoeff` and/or remove?

dpsanders: Since this is basically the same as `getindex`?

src/auxiliary.jl
```diff
@@ -187,9 +231,9 @@ end
 For `a::Taylor1` returns `zero(a[1])` while for `a::TaylorN` returns
 a zero of a k-th order `HomogeneousPolynomial` of proper type.
 """
-zero_korder(a::Taylor1, ::Int) = zero(a[1])
+zero_korder(a::Taylor1, ::Int) = zero(a[0])
 
-zero_korder(a::TaylorN, k::Int) = HomogeneousPolynomial(zero(a[1][1]), k)
+zero_korder(a::TaylorN, k::Int) = HomogeneousPolynomial(zero(a[0][0]), k)
```
dpsanders: Should this be `a [0][1]`?

src/evaluate.jl
```diff
@@ -189,7 +189,10 @@ function evaluate(a::HomogeneousPolynomial{T}, vals::Array{S,1} ) where
     return suma
 end
 
-evaluate(a::HomogeneousPolynomial) = zero(a[1])
+function evaluate(a::HomogeneousPolynomial)
+    a.order == 0 && return a[1]
+    zero(a[1])
```
dpsanders: `a[0]`?

src/other_functions.jl
```diff
@@ -108,9 +108,9 @@ function mod2pi(a::Taylor1{TaylorN{T}}) where {T<:Real}
 end
 
 function abs(a::TaylorN{Taylor1{T}}) where {T<:Real}
-    if constant_term(a)[1] > zero(T)
+    if constant_term(a)[0] > zero(T)
```
dpsanders: `> 0`

src/other_functions.jl
```diff
@@ -120,9 +120,9 @@ function abs(a::TaylorN{Taylor1{T}}) where {T<:Real}
 end
 
 function abs(a::Taylor1{TaylorN{T}}) where {T<:Real}
-    if constant_term(a[1]) > zero(T)
+    if constant_term(a[0]) > zero(T)
```
dpsanders: `> 0`

src/other_functions.jl
```diff
@@ -182,7 +183,7 @@ Inexact equality comparison between polynomials: returns `true` if
 `norm(x-y,1) <= atol + rtol*max(norm(x,1), norm(y,1))`, where `x` and `y` are
 polynomials. For more details, see [`Base.isapprox`](@ref).
 """
-function isapprox(x::T, y::S; rtol::Real=rtoldefault(x,y), atol::Real=0,
+function isapprox(x::T, y::S; rtol::Real=rtoldefault(x,y), atol::Real=0.0,
```
dpsanders: I don't see the need for this change (from `0` to `0.0`).

src/power.jl
```diff
@@ -317,11 +317,11 @@ function sqrt(a::Taylor1)
 
     # The last l0nz coefficients are set to zero.
     lnull = div(l0nz, 2)
-    @inbounds aux = sqrt( a[l0nz+1] )
+    @inbounds aux = sqrt( a[l0nz] )
     T = typeof(aux)
 
-    c = Taylor1( zeros(T, a.order+1) )
-    @inbounds c[lnull+1] = aux
+    c = Taylor1( zero(T), a.order )
```
dpsanders: `Taylor1{T}(0, a.order)`?

test/manyvariables.jl
```diff
@@ -368,36 +418,38 @@ end
     @test typeof( norm(x) ) == Float64
     @test norm(x) > 0
     @test norm(a) == norm([3,4,6,8.0])
-    @test norm(a,4) == sum([3,4,6,8.0].^4)^(1/4.)
-    @test norm(a,Inf) == 8.
-    @test norm((3.0 + 4im)*x) == abs(3.0 + 4im)
+    @test norm(a, 4) == sum([3,4,6,8.0].^4)^(1/4.)
+    @test norm(a, Inf) == 8.0
+    @test norm((3.0 + 4im)*x) == abs(complex(3.0,4.0))
```
dpsanders: It's more readable as `3 + 4im`.

docs/src/userguide.md
lbenet: I guess this could be possible, though I don't know how, since `TaylorN` are vectors. This currently throws a `BoundsError`; leaving it aside for the moment.

Incidentlally, your suggestion made me notice that the comment on that line is wrong; it should read "# get the 6th coeff of the 8th order term". I have corrected it.

src/arithmetic.jl
lbenet: Thanks for the suggestion, it surely makes everything more readable. 

I don't recall exactly where, but at some point I encountered a case where I was comparing two vanishing `HomogeneousPolynomials` of different order, and getting `false`, which is not correct. Vanishing `HomogeneousPolynomials` are kind of special...

src/arithmetic.jl
lbenet: I think `b.coeffs[1]` here is correct, though it doesn't look so nice, since it covers both `b::HomogeneousPolynomial` (b[1]) and `b::TaylorN` (b[0]).

src/arithmetic.jl
lbenet: Same as above; `b` may be an HomogeneousPolynomial...

src/arithmetic.jl
lbenet: The difference is subtle; `b[0]` is a HomogeneousPolynomial of order 0, while `constant_term(b)` is just the (numeric) coefficient; using `b[0]` throws an error because dividing by `HomogeneousPolynomials`s is not defined.

src/auxiliary.jl
lbenet: I have renamed it as proposed; sometimes it is more practical than using the vector notation.

src/auxiliary.jl
lbenet: Good catch! Changed to `constant_term(a)`. Thanks!

src/evaluate.jl
lbenet: `a[1]` is correct, since `a` is a `HomogeneousPolynomial`

src/other_functions.jl
lbenet: Changed as suggested, including another similar case (above).

src/other_functions.jl
lbenet: Done! Thanks!

src/other_functions.jl
lbenet: `atol` is usually some sort of Float; I though like this I avoided a conversion. I am leaving it as it is, buty it may be changed eventually.

src/power.jl
lbenet: This throws a MethodError...

test/manyvariables.jl
lbenet: Reverted; thanks!

lbenet:
Thanks a lot for the review, which was as painful as the PR itself. 

I'm pushing the changes and waiting for travis to do its work!

dpsanders:
Is this ready to merge?

lbenet:
I think so.

dpsanders:
Thanks a lot!

lbenet:
Thanks to you too for your review!

Can you take a look in #135? We should tag soon a new version...

lbenet:
Sorry, I meant #136...

PR #136, opened 2017-11-11
Integration of TaylorN series

lbenet:
Solves #130.

PR #137, opened 2017-11-18
RFC: Return n-th derivative of a Taylor1 as a Taylor1

PerezHz:
Current implementation returns the derivative of a `Taylor1` as a `Taylor1`, but higher-order derivatives of a `Taylor1` are returned as single values. This PR is intended to instead return higher-order derivatives of a `Taylor1` polynomial as a `Taylor1` polynomial. In order to do this, this PR adds a new method to `derivative(n::Int, a::Taylor1)` and renames the existing method for the value of the n-th derivative of a Taylor1 polynomial as `derivativeval` (name suggestions are more than welcome 😛!). This was done in order to get the whole n-th derivative polynomial when it's needed, but the existing method for the value of the n-th derivative is faster when only the numerical value is needed, so it was kept for performance reasons.

### DIFF
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -32,7 +32,7 @@ import Base: zero, one, zeros, ones, isinf, isnan, iszero,
 
 export Taylor1, TaylorN, HomogeneousPolynomial, AbstractSeries
 
-export getcoeff, derivative, derivativeval, integrate,
+export getcoeff, derivative, integrate,
     evaluate, evaluate!, inverse,
     show_params_TaylorN, show_monomials,
     get_order, get_numvars,
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -32,7 +32,7 @@ import Base: zero, one, zeros, ones, isinf, isnan, iszero,
 
 export Taylor1, TaylorN, HomogeneousPolynomial, AbstractSeries
 
-export getcoeff, derivative, integrate,
+export getcoeff, derivative, derivativeval, integrate,
     evaluate, evaluate!, inverse,
     show_params_TaylorN, show_monomials,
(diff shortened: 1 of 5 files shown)


lbenet:
I like the idea of adding such method. 

Currently, we have `derivative(a)` which returns the `Taylor1` series of the derivative, and `derivative(n, a)` which returns the value of the n-th derivative of the polynomial (evaluated at zero). What about adding `derivative(a, n)` with the behavior you propose? The advantage is that such a change would break any current code.

Any opinion on this @dpsanders?

dpsanders:
I think having `derivative(a, n)` be different from `derivative(n, a)` is confusing, but I don't really have a better suggestion.

lbenet:
Or use keyword variables like `derivative(a, ord=n)` to return what is proposed here...

PerezHz:
Just updated `derivative` with the latest agreed convention: `derivative(n,a)` returns just the value of the n-th derivative evaluated at the original point of expansion; `derivative(a,n)` returns the whole n-th derivative polynomial

lbenet:
Thanks a lot! 

I'll wait to the green lights of travis before merging. 

Maybe also @dpsanders has other comments.

PerezHz:
> First, since #136 is merged you have to rebase to current master.

Thanks; done! 😄 

> Second, there seems to be a problem with (at least) one test in travis; see this; I suggest to correct the failing test (Julia v0.7-dev only), and wait to see if there are other issues with travis.

I just corrected this (at least locally; let's see if travis agrees), and also added some tests for the new `derivative!` function and `jacobian` for mixtures as well to keep coverage up!

PerezHz:
Thank you for all your comments and suggestions!

lbenet:
Thanks a lot! I'm merging this!

PR #143, opened 2017-12-04
Avoid irrelevant checks in getindex and setindex!

lbenet:
This PR erases some `@assert`s and other bound-checking instructions which I introduced while implementing #135. They induce a general slowdown and are not necessary.


lbenet:
Rebased to current master; once travis finishes, this is ready for being merged.

PR #145, opened 2018-02-16
Correct the returned series when factorizing, with tests

### LINKED ISSUE #58: sqrt gives misleading O() term
```
julia> t = Taylor1(10)
 1.0 t + 𝒪(t¹¹)

julia> sqrt(t^2 + t^3)
 1.0 t + 0.5 t² - 0.125 t³ + 0.0625 t⁴ - 0.0390625 t⁵ + 0.02734375 t⁶ - 0.0205078125 t⁷ + 0.01611328125 t⁸ + 𝒪(t¹¹)
```

This seems to imply that the t^9 and t^10 terms have zero coefficients, which is incorrect.

lbenet:
This fixes #58 and other similar cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ julia:
 notifications:
     email: false
 
+matrix:
+  allow_failures:
+  - julia: nightly
+
 script:
  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
  - julia --check-bounds=yes -e 'Pkg.clone(pwd()); Pkg.build("TaylorSeries"); Pkg.test("TaylorSeries"; coverage=true)'
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -486,9 +486,14 @@ term of the denominator.
     end
 
     @inbounds for i = 0:k-1
+        k+ordfact-i > b.order && continue
         c[k] += c[i] * b[k+ordfact-i]
     end
-    @inbounds c[k] = (a[k+ordfact]-c[k]) / b[ordfact]
+    if k+ordfact ≤ b.order
+        @inbounds c[k] = (a[k+ordfact]-c[k]) / b[ordfact]
+    else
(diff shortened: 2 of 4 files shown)


lbenet:
I'll add a summary here: Currently, in master, we get the following:

```julia
julia> t = Taylor1(5)
 1.0 t + 𝒪(t⁶)

julia> 1/(1+t)
 1.0 - 1.0 t + 1.0 t² - 1.0 t³ + 1.0 t⁴ - 1.0 t⁵ + 𝒪(t⁶)

julia> t/(t+t^2) # should be identical to the result above
 1.0 - 1.0 t + 1.0 t² - 1.0 t³ + 1.0 t⁴ + 𝒪(t⁶)

julia> t*sqrt(1+t)
 1.0 t + 0.5 t² - 0.125 t³ + 0.0625 t⁴ - 0.0390625 t⁵ + 𝒪(t⁶)

julia> sqrt(t^2+t^3) # should be identical to the result above
 1.0 t + 0.5 t² - 0.125 t³ + 𝒪(t⁶)
```

This PR corrects this.

lbenet:
Tests should pass now...

lbenet:
@dpsanders I think this is ready!

dpsanders:
I don't follow the calculations, but if the new tests work, LGTM!

lbenet:
The basic idea is to mimic an "extension" of the order of the independent variable, by using zeros appropriately; this is used to compute the coefficients (which before were simply set to zero) to the correct order.

Considering the example of `sqrt(t^2+t^3)` with `t=Taylor1(5)`, the idea is to treat the independent variable of the expression as being of order 7 (instead of order 5), because `t^2` can be factorized (which are the two orders lost in the returned expression). So the idea is to mimic the behavior of `sqrt( `Taylor1(7)^2+Taylor1(7)^3 )` (using `t` and zeros for order 6 and 7) and return the result up to order 5.

I hope this short explanation makes some sense.

Issue #148, opened 2018-02-20
Allow differentiation and integration with respect to symbolic variable names


dpsanders:
```jl
t, x = set_variables("x y")
integrate(t, :t)
integrate(x, :t)
```

PR #150, opened 2018-02-20
Add symbols for variables to params

dpsanders:
Add variable symbols to `_params_TaylorN_`.

### DIFF
--- a/src/parameters.jl
+++ b/src/parameters.jl
@@ -11,28 +11,35 @@ DataType holding the current parameters for `TaylorN` and
 
 **Fields:**
 
-- `order          :: Int`  Order (degree) of the polynomials
-- `num_vars       :: Int`  Number of variables
-- `variable_names :: Array{String,1}` Name of the variables
+- `order            :: Int`  Order (degree) of the polynomials
+- `num_vars         :: Int`  Number of variables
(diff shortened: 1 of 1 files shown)


lbenet:
One test is failing; the problem seems to be [this one](https://github.com/JuliaDiff/TaylorSeries.jl/blob/c6b45d29dd19eafc5491970d555684087a1ec6bc/test/manyvariables.jl#L37).
The problem is that `set_variable_names` now returns a vector of symbols, while before it was returning a vector of strings. 

Probably it is better that it returns `nothing`, or print some info, and for the tests we could have
```julia
@test TaylorSeries.set_variable_names(["x","y"]) == nothing
@test TaylorSeries._params_TaylorN_.variable_names == ["x","y"]
@test TaylorSeries._params_TaylorN_.variable_symbols == [:x, :y]
@test get_variable_symbols() == [:x, :y]
```

Note finally that `get_variable_symbols` is not exported.
Also, it would be nice to have

lbenet:
@dpsanders In order to address #151 and #148, this PR should be merged. See `lb/variable_symbols` branch for a commit that fixes the 0.6 tests and extends a little bit the functionality. Feel free to cherry-pick that commit.

PR #152, opened 2018-02-22
Fix issues related to symbols

lbenet:


lbenet:
@dpsanders This fixes the tests of #150, and is required to allow integration, derivation and evaluation using symbols.

This is ready to be merged.

PR #157, opened 2018-02-24
Deprecations related to Julia0.7

lbenet:


lbenet:
This is to pave the way to Julia 0.7. There are still things to do...

lbenet:
Essentially all warnings due to deprecations have been solved.

This is ready for review and merge. We should afterwards tag a new minor version.

dpsanders:
LGTM, thanks!

PR #160, opened 2018-03-10
Solve deprecations that break tests in Julia 0.7

lbenet:


lbenet:
I think this is ready.

PR #162, opened 2018-03-25
Type-stable mutating functions

lbenet:


lbenet:
Just to provide some context, currently master with `Traceur.jl` yields the following:
```julia
julia> using TaylorSeries, Traceur

julia> ti = Taylor1(Int, 5)
 1 t + 𝒪(t⁶)

julia> @trace sqrt(1+ti);
(sqrt)(::TaylorSeries.Taylor1{Int64}) at /Users/benet/.julia/v0.6/TaylorSeries/src/power.jl:310
  returns Union{TaylorSeries.Taylor1{Float64}, TaylorSeries.Taylor1{Int64}}
 1.0 + 0.5 t - 0.125 t² + 0.0625 t³ - 0.0390625 t⁴ + 0.02734375 t⁵ + 𝒪(t⁶)

julia> @trace (1+ti)^2.0
(^)(::TaylorSeries.Taylor1{Int64}, ::Float64) at /Users/benet/.julia/v0.6/TaylorSeries/src/power.jl:73
  returns Union{TaylorSeries.Taylor1{Float64}, TaylorSeries.Taylor1{Int64}}
 1 + 2 t + 1 t² + 𝒪(t⁶)
```

This PR solves that.

This is related to some odd behavior found in https://github.com/PerezHz/TaylorIntegration.jl/pull/31.

lbenet:
I have pushed three more commits. As stated above, the issue that motivated this PR was related with some the lack of type-stability for mutating functions. In the new commits, I've changed the behavior of `a^n` and `a^r` (a::Taylor1{T}, n::Integer, r::Real, and T not an Integer), so for example `a^3` and `a^3.0` yield the same result (as it was before) but using `pow!` instead of `power_by_squaring`. The reasons for the latter are accuracy, mimics what happens in Base and the equivalence when using mutating functions (e.g., for TaylorIntegration).

@dpsanders While I want to amend the docstrings a little bit, I think this is ready for reviewing.

lbenet:
@dpsanders I think this is ready to be merged, and I think we should tag a new patch version. 

The changes related to the docs, while numerous, are simply to get the API-page to display correctly (which currently is not).

PR #165, opened 2018-04-17
Separate methods for ^ to use power_by_squaring or pow!

lbenet:


lbenet:
This implements a suggestion of @dpsanders (off-line discussion)

lbenet:
This makes that `(interval(-1,1) + Taylor1(5)^4` yields the expected result.

dpsanders:
Maybe add some tests with intervals? That would require adding `IntervalArithmetic.jl` to `test/REQUIRE`.

lbenet:
Good point; I'll implement some tests for intervals.

lbenet:
Due to the use of `power_by_squaring`, some intervals arising from `(t-a)^n` (`a` is an `Interval`) are actually wider than the explicit formula, which evaluates `a^k` with appropriate `k`. This is simply the dependency problem.

lbenet:
I think this is ready for review.

lbenet:
Tests are passing and https://github.com/JuliaDiff/TaylorSeries.jl/pull/165/commits/4d5ac348a807345e81c2c83d37ef8264657a57ec provided *narrower* intervals.

PR #169, opened 2018-06-27
Allow using Base.show_default instead of custom show

lbenet:
This allows to use the default function in Base for `show` instead of the customized one. This is important, e.g., when saving series to a file to load them later.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -48,6 +48,7 @@ norm
 isapprox
 isfinite
 displayBigO
+use_show_default
 ```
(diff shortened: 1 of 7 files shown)


lbenet:
Tests in Julia 0.7 are not passing because they rely in `IntervalArithmetics` v0.13.0. Once a new version of is tagged, tests will pass (they pass locally using current master). I propose to merge this, and wait for the necessary tags to release a new version.

@dpsanders Do you have any comments?

PR #171, opened 2018-07-07
Skip interval tests in Julia0.7

lbenet:
In Julia0.7 they require current master of IntervalArithmetics


lbenet:
Merging, to tag a version that will be the final one that supports Julia 0.6.

PR #172, opened 2018-07-09
Support Julia 0.7 only

lbenet:


lbenet:
Rebased to current master

lbenet:
@dpsanders This seems to be ready...

lbenet:
@dpsanders This is essentially ready to be merged, and then drop support for Julia v 0.6.x. Aside from the deprecation warnings, I have also updated the documentation (solving some errors that appeared), the README.md and LICENSE.md files.

For some tests, IntervalArithmetic.jl is currently needed. I am not sure if this is handled automagically by Pkg, or if we actually need to tag a new version.

Can you take a look on this?

dpsanders:
Very nice! It seems to work fine on 1.0 except for the following:
```
WARNING: could not import LinearAlgebra.gradient into TaylorSeries
WARNING: could not import Base.ctranspose into TaylorSeries
WARNING: could not import Base.endof into TaylorSeries
```

I guess the IntervalArithmetic tests will work once IntervalArithmetic works...

dpsanders:
So shall I merge?

lbenet:
I am in favor of merging, but then we also need to tag a new version (v0.8.0, I guess), since we drop Julia 0.6.x support.

Did you check the changes in README.md? (The idea is to prepare things to eventually submit what is in lb/joss to JOSS.)

lbenet:
@dpsanders Should we merge and then tag a new minor version of this? What about my proposal about JOSS?

lbenet:
I think it is worth to get full compatibility with Julia 1.0 in this PR. I am dealing with this now.

lbenet:
Mmmm... Seems that Julia 0.7 imposes explicitly to import Base.gradient, while Julia 1.0 does not accept that. I will revert that change, and in another PR address Julia 1.0.

lbenet:
Tests pass. Some warnings are due to use IntervalArithmetic v0.14, and not the version compatible with Julia 0.7. There is another warning in `test/common.jl` that I will check later. Merging this.

PR #176, opened 2018-08-14
Fix a method of `evaluate` for HomogeneousPolynomials

### LINKED ISSUE #175: Allow evaluation of a HomogeneousPolynomial on an IntervalBox
The following hangs / gives a stack overflow?

```jl
julia> t, a = set_variables("t a", order=3)
  1.0 t + 𝒪(‖x‖⁴)
  1.0 a + 𝒪(‖x‖⁴)

julia> typeof(t[1])
TaylorSeries.HomogeneousPolynomial{Float64}

julia> using IntervalArithmetic

julia> X = (0..1)×(3..4)

julia> t[1](X)

```

lbenet:
Fixes #175

### DIFF
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -219,6 +219,8 @@ evaluate(a::HomogeneousPolynomial{T}, vals::Array{S,1} ) where
 
 evaluate(a::HomogeneousPolynomial, v, vals...) = evaluate(a, (v, vals...,))
 
+evaluate(a::HomogeneousPolynomial, v) = evaluate(a, v...)
+
 function evaluate(a::HomogeneousPolynomial)
     a.order == 0 && return a[1]
     zero(a[1])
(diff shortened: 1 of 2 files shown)


lbenet:
Tests are passing in Julia 0.6; the problem in Julia 0.7 is related to the new `Pkg`, and is solved in #172 . Also, all fixes proposed by femtocleaner are already implemented in #172. I propose to merge this as it is. 

@dpsanders Any comments?

lbenet:
Tests are broken here, due to some differences in `Pkg` in Julia 0.6 and 0.7, which are relevant for travis. They pass if those changes are included in travis (after rebasing to lb/julia07, #172). 

The question is if this should be merged first, or simply merge #172, tag a new version (which is only compatible with Julia 0.7) and then merge this.

lbenet:
@dpsanders Any thoughts?

PR #177, opened 2018-08-22
Iterators

lbenet:


lbenet:
I just compared the performance (`perf/fateman.jl`) in v0.7.4 (julia0.6) with current master and this PR: master is faster than v0.7.4 about a 30%, and this PR yields a further 2-3% of speed up.

lbenet:
The only thing we need now to tag a version 0.7 compatible only, is the new tag of `IntervalArithmetic.jl`

PR #178, opened 2018-08-22
Promotion rule for nested Taylor1's

lbenet:
This PR adds a promotion rule involving `Type{Taylor1{Taylor1{T}}` and `Type{Taylor1{T}}`; the present situation leads to an ambiguity which causes an error in Julia 0.7.


lbenet:
Merging.

PR #181, opened 2018-09-20
JOSS paper

lbenet:


lbenet:
Thanks for reviewing this. I agree and have implemented all suggestions.

lbenet:
I've just pushed another commit with your last suggestion. Thanks a lot!

lbenet:
Merging, to allow JOSS review go on.

PR #188, opened 2018-12-13
Address observations to paper

lbenet:
This PR addresses the [observations](https://github.com/openjournals/joss-reviews/issues/1043#issuecomment-443262789) by @tobydriscoll on our submission to JOSS. 

Concretely:
- it adds some pointers to other automatic differentiation packages in the docs; 
- it removes exporting `gradient`, `jacobian` and `hessian`; and
- adds `differentiate` as a synonym of `derivative` (including proper refs in the docs and few tests).

@dpsanders, Do you have any comments/additions/corrections on this?

### DIFF
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -49,9 +49,8 @@ export getcoeff, derivative, integrate, differentiate,
     get_order, get_numvars,
     set_variables, get_variables,
     get_variable_names, get_variable_symbols,
-    # jacobian, hessian,
-    ∇, jacobian!, hessian!,
-    taylor_expand, update!, constant_term
+    # jacobian, hessian, jacobian!, hessian!,
+    ∇, taylor_expand, update!, constant_term
 
 include("parameters.jl")
 include("hash_tables.jl")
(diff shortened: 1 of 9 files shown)


src/TaylorSeries.jl
```diff
@@ -49,7 +49,8 @@ export getcoeff, derivative, integrate,
     get_order, get_numvars,
     set_variables, get_variables,
     get_variable_names, get_variable_symbols,
-    ∇, jacobian, jacobian!, hessian, hessian!,
+    # jacobian, hessian,
+    ∇, jacobian!, hessian!,
```
dpsanders: Why are these still exported?

dpsanders:
LGTM apart from the comment on exporting ` jacobian!`.

src/TaylorSeries.jl
lbenet: They are in-place versions of `jacobian` and `hessian`; they are still exported since I think there is no name-conflict risk.

dpsanders:
I meant that it seems a bit inconsistent to export these but not the non-in place versions.

lbenet:
Ok. I'll ex-export those functions as well... I'll push a new PR with those changes later.

lbenet:
I just pushed a new commit that unexports the functions as suggested; actually, I did check if those functions exist in ForwarDiff.jl, and they do.

lbenet:
Merging.

Issue #193, opened 2019-02-04
Improve `evaluate` for `Interval`s


lbenet:
The naive way in which `evaluate` is implemented for `Interval`s may be a limitation, e.g., in relation to https://github.com/JuliaIntervals/TaylorModels.jl/pull/18. For reference, see Sect 2.3 [here](https://easychair.org/publications/open/9Tz3). I think we should add an alternative method, e.g., normalizing the `Interval`s, and take the intersection.

mforets:
Alright, so IIUC you'd like to write an alternative to [this method](https://github.com/JuliaDiff/TaylorSeries.jl/blob/master/src/evaluate.jl#L17) (for the univariate case, say), among one of the 3 methods proposed in that paper. For our case `T::Interval{Float64}` so `suma = suma*dx + a[k]` uses interval arithmetic, and this corresponds to the method "interval arithmetic - no normalized". 

Let me propose what i think is a fourth one - using the bounding property of [Bernstein polynomials](https://en.wikipedia.org/wiki/Bernstein_polynomial), see e.g. [here](https://ac.els-cdn.com/0377042793E0270V/1-s2.0-0377042793E0270V-main.pdf?_tid=a1c3b342-c7ad-4cb0-b97e-195bb9fd368f&acdnat=1549254851_18887d3bbe5e0357249a329946fa8ef1), or [Smith's thesis](https://d-nb.info/1028327854/34). The code in [JuliaReach/BernsteinExpansions.jl](https://github.com/JuliaReach/BernsteinExpansions.jl) can be adapted for this purpose (that we wrote originally with @roccaa). I ignore how does it compare to the other methods, though.

I just experimented with Example 1:

```julia
julia> using TaylorSeries, IntervalArithmetic

julia> x = Taylor1(Interval{Float64}, 6)
 [1, 1] t + 𝒪(t⁷)

julia> R = Interval(0, 6.)
[0, 6]

julia> evaluate(0.1*x^3 - 0.5*x^2 + 1, R) 
[-17, 4.60001]   # this is [-17, 22.6] in Example 1 ... why?? answer below -->

julia> 0.1 * R^3  - 0.5 * R^2 + 1  # Horner's rule is giving better bounds than naive substitution
[-17, 22.6001]
```

I tried the normalisation too; note that:

```python
sage: f = 0.1*x^3 - 0.5*x^2 + 1

[...]

PR #194, opened 2019-02-04
Use promotion in Taylor1's evaluate

mforets:
A small change to avoid the code duplicate in the `evaluate` function for `Taylor1`.

### DIFF
--- a/src/evaluate.jl
+++ b/src/evaluate.jl
@@ -21,13 +21,12 @@ function evaluate(a::Taylor1{T}, dx::T) where {T<:Number}
     end
     suma
 end
+
 function evaluate(a::Taylor1{T}, dx::S) where {T<:Number, S<:Number}
-    suma = promote(a[end], dx)[1]
-    @inbounds for k in a.order-1:-1:0
-        suma = suma*dx + a[k]
-    end
-    suma
+    R = promote_type(T, S)
+    return evaluate(convert(Taylor1{R}, a), convert(R, dx))
 end
+
 evaluate(a::Taylor1{T}) where {T<:Number} = a[0]
 
 """


lbenet:
I suspect that the proposed implementation leads to more allocations (and perhaps some penalties in timing) against the current one. Have you checked this?

mforets:
Nope, i didn't time it, but i understand that your concern is to avoid an eventual copy behind `convert(Taylor1{R}, a)`, defined [in this line](https://github.com/JuliaDiff/TaylorSeries.jl/blob/master/src/conversion.jl#L10).

Let me note by passing that if the `Taylor1` is already of the target type (`R` in our case), it will create a new object, but to avoid this why not introducing a no-op `convert(::Type{Taylor1{T}}, a::Taylor1{T}) where {T<:Number} = a`.

Another approach to this PR without affecting performance, but more lengthy, is to use a helper function like:

```julia
function evaluate(a::Taylor1{T}, dx::T) where {T<:Number}
    @inbounds suma = a[end]
    _evaluate_horner!(suma, a, dx)
end

function evaluate(a::Taylor1{T}, dx::S) where {T<:Number, S<:Number}
    suma = promote(a[end], dx)[1]
    _evaluate_horner!(suma, a, dx)
end

function _evaluate_horner!(suma, a, dx)
    @inbounds for k in a.order-1:-1:0
        suma = suma*dx + a[k]
    end
    suma
end
```

mforets:
BTW the *motivation* for this PR is adding new options to evaluate, such as `horner = true / false` and `normalise = true / false` discussed in #193, so you want the code to be shared. The signature of that function can be:

```julia
function evaluate(a::Taylor1{<:Interval{T}}, dx::Interval{T};
                              horner=true, normalize=false) where {T<:Number}
...
```

lbenet:
I am currently facing some issues which are somewhat related this PR. The current implementation, or the current in `master`, throw an error due to promotion limitations. 

The example that breaks is evaluates `a::TaylorModel1{TaylorModelN{2,Float64,Float64},Float64}` at `dx::Interval{Float64}`. Clearly, we should bring better support when using `Interval`s. This is probably better address at `TaylorModels.jl`.

mforets:
Seems that this PR is no longer needed (at least, the motivation concerning to apply or not normalization is handled with functions in https://github.com/JuliaDiff/TaylorSeries.jl/blob/master/src/intervals.jl).

PR #195, opened 2019-02-05
Add improved evaluation for intervals

mforets:
This PR implements the improved evaluation of taylor series in one dimension, for intervals, using normalization discussed in https://github.com/JuliaDiff/TaylorSeries.jl/issues/193. The specialized `evaluate` for intervals is added as an optional dependency through Requires. The default options are such that the current ones on master are used, although we may want to change this.

With this code we can reproduce the observations from [the example](https://github.com/JuliaDiff/TaylorSeries.jl/issues/193#issuecomment-460358474), see below, where i also added some timing results. The `intersection` option is much slower but also substantially more precise.

```julia
julia> using TaylorSeries, IntervalArithmetic

julia> t = (0..0) + Taylor1(3)
 [1, 1] t + 𝒪(t⁴)

julia> f(x) = 0.1 * x^3 - 0.5*x^2+1

julia> g = f(t)
 [1, 1] + [-0.5, -0.5] t² + [0.1, 0.100001] t³ + 𝒪(t⁴)

julia> R = 0..6
[0, 6]

julia> # default options (corresponds to current on master)
       evaluate(g, R)
[-17, 4.60001]

julia> evaluate(g, R, horner=true, normalize=true)
[-8.00001, 6.40001]

julia> evaluate(g, R, horner=true, normalize=false)
[-17, 4.60001]

julia> evaluate(g, R, horner=false, normalize=true)
[-4.40001, 6.40001]

julia> evaluate(g, R, horner=false, normalize=false)
[-17, 22.6001]

julia> # best: intersect four combinations
       evaluate(g, R, intersection=true)
[-4.40001, 4.60001]

julia> using BenchmarkTools

julia> for h in [true, false]
           for n in [true, false]
               println("\n horner=$h, normalize=$n\n")
               @btime evaluate($g, $R, horner=$h, normalize=$n, intersection=false)
           end
       end

 horner=true, normalize=true

  1.573 μs (40 allocations: 3.02 KiB)

 horner=true, normalize=false

  87.891 ns (0 allocations: 0 bytes)

 horner=false, normalize=true

  5.235 μs (126 allocations: 6.78 KiB)

 horner=false, normalize=false

  4.508 μs (81 allocations: 3.22 KiB)

# intersecting 4 combinations
julia> @btime evaluate($g, $R, intersection=true)
  10.918 μs (217 allocations: 10.70 KiB)
[-4.40001, 4.60001]
```

Global minimum and maximum:

```julia
julia> using IntervalOptimisation

julia> global_min, minimisers = minimise(x ->  0.1 * x^3 - 0.5*x^2+1, 0..6);

julia> global_min
[-0.85439, -0.851851]

julia> global_min, minimisers = minimise(x ->  -0.1 * x^3 + 0.5*x^2 - 1, 0..6);

julia> -global_min
[4.59805, 4.60487]
```
So the maximum is relatively tight but the minimum is not.


lbenet:
What about using methods that exploit multiple dispatch through `Val{true}` and `Val{false}` for the keyword parameters? I am not sure if that would be faster or not, but certainly it avoids the `if...end` blocks by dispatch, or at least most of them. Another suggestion to avoid allocations when `intersect=true` is to compute the intersection by pairs, which are stored in the returned variable.

mforets:
> What about using methods that exploit multiple dispatch through `Val{true}` and `Val{false}` for the keyword parameters?

In julia v1.0 using keywords arguments does not induce a performance penalty, as it was in previous versions (don't ask me to explain why, i've read it somewhere and also verified it with my own experiments).

The 3 `if` clauses should add little extra overhead (don't they just add 1 more machine instruction? so eating just a couple of ns?). I don't know if dispatching on values is more efficient than using kwargs in this particular case, or in general, i haven't tried. I usually use kwargs because they are natural and usually the slow part hides somewhere else.

> Another suggestion to avoid allocations when `intersect=true` is to compute the intersection by pairs, which are stored in the returned variable.

Good. It does seem like one can do the intersection in one shot. Let's try! (let me know if this is what you had in mind or it was sth else)

```julia
function intersect_four_naive(I1::Interval{T}, I2::Interval{T}, I3::Interval{T}, I4::Interval{T}) where T
    return I1 ∩ I2 ∩ I3 ∩ I4
end

# for a fair comparison, i add the emptiness check below
function intersect_four_fused(I1::Interval{T}, I2::Interval{T}, I3::Interval{T}, I4::Interval{T}) where T
    low = max(I1.lo, I2.lo, I3.lo, I4.lo)
    high = min(I1.hi, I2.hi, I3.hi, I4.hi)
    result = Interval(low, high)
    isempty(result) && return emptyinterval(T)
    return result
end
```

Test:

```julia
julia> (I1, I2, I3, I4) = [Interval(rand()-1, rand()) for _ in 1:4]

julia> @btime intersect_four_naive($I1, $I2, $I3, $I4)
26.602 ns (0 allocations: 0 bytes)

[-0.0110472, 0.06795]

julia> @btime intersect_four_fused($I1, $I2, $I3, $I4)
  10.938 ns (0 allocations: 0 bytes)

[-0.0110472, 0.06795]
```

That's a nice ~2x speedup 👍

dpsanders:
The logic for the empty case is wrong:
```
julia> intersect_four_fused(0..1, 3..4, 0..1, 0..1)
[3, 1]
```
You could make this into a method of `intersect` and accept an arbitrary number of arguments.

mforets:
Good catch, thanks. I didn't know about the subtle difference between `Interval` and `interval`. I propose this fix:

```julia
import IntervalArithmetic: is_valid_interval

function intersect_four_fused(I1::Interval{T}, I2::Interval{T}, I3::Interval{T}, I4::Interval{T}) where T
    low = max(I1.lo, I2.lo, I3.lo, I4.lo)
    high = min(I1.hi, I2.hi, I3.hi, I4.hi)
    if !is_valid_interval(low, high) 
        return emptyinterval(T)
    end
    return Interval(low, high)
end
```

FWIW the time is now ~14ns.

> You could make this into a method of `intersect` and accept an arbitrary number of arguments.

I've opened https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/256 for the general case. In the meantime, i'll push the new `intersect_four_fused` code to this branch.

mforets:
I think that the normalization is wrong:

```julia
julia> using TaylorSeries, IntervalArithmetic

julia> f(t) = -0.8 + -0.9t + 3.6t^2 + 2.7t^3

julia> t = (0..0) + Taylor1(3)
 [1, 1] t + 𝒪(t⁴)

julia> f(t)
 [-0.800001, -0.8] + [-0.900001, -0.9] t + [3.6, 3.60001] t² + [2.7, 2.70001] t³ + 𝒪(t⁴)

julia> evaluate(f(t), (-1..1), normalize=false)
[-8.00001, 6.40001]

julia> evaluate(f(t), (-1..1), normalize=true) # ??
[4.59999, 4.60001]
```

lbenet:
From [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/issues/193#issuecomment-462654861):

> (Please feel free to commit on top of https://github.com/JuliaDiff/TaylorSeries.jl/pull/195/files by dispatching on values instead of kwargs, or to change it in another way, if you prefer.)

The different normalizations are dealt by dispatch; changes in `evaluate` are coming...

Sorry, not a `git` guru; so I have pushed to the branch `mforets/normalize_evaluate`. Please let me know to what branch should I push so this PR reflects the proposed changes.

mforets:
Great, I saw that you also added the normalization for the n-variate case as well, thanks 👍👍 

The shortcoming that i see using a single boolean for dispatch is that one cannot choose to not normalize (assuming that there is a shared function for evaluation, i.e. merging `evaluate` with `normalize_taylor`). This is the reason that i used a string for the outer function signature and a bool for the inner signature of `_normalize`. That said, perhaps we just don't care about "none", because it always helps to normalize; I wanted to keep all possible combinations in the experimental phase :)

> Please let me know to what branch should I push so this PR reflects the proposed changes.

Since the checkbox "Allow edits from maintainers. " is selected in my PR, and you are a maintainer, i think that [you can directly push to this branch](https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/).

So maybe this works:

```
$ git checkout mforets/normalize_evaluate_Taylor1
$ git cherry-pick 6fb2f0c && git cherry-pick f9d3a9d
$ git push origin mforets/normalize_evaluate_Taylor1
```

lbenet:
> The shortcoming that i see using a single boolean for dispatch is that one cannot choose to not normalize (assuming that there is a shared function for evaluation, i.e. merging `normalize_taylor` into `evaluate`). This is the reason that i used a string for the outer function signature and a bool for the inner signature of `_normalize`. That said, perhaps we just don't care about "none", because it always helps to normalize; I wanted to keep all possible combinations in the experimental phase :)

You are right with the boolean limitation; the point is that it paves the way to generalize this. For example, instead of using `Val(true)` it is possible to use `Val(:symmetric)`; other symbols should then be added individually.

My idea is to use `_normalize` from `evaluate`, or perhaps another function, which would allow defining alternative methods of implementing evaluate. I am currently not implementing the most general case, but one which would make `evaluate(a)` faster than it is now in https://github.com/JuliaIntervals/TaylorModels.jl/pull/18 .

lbenet:
@mforets Sorry for the delay; I just pushed some commits which allow to normalize the Taylor polynomials and evaluate with in a bit better way, which avoids some allocations. 

The `evaluate` methods you where defining are still there, but commented. Feel free to go ahead to improve this, along the way you had in mind.

lbenet:
@mforets I just pushed a new commit, which simplifies a bit the code in `src/evaluate.jl`.

src/evaluate.jl
```diff
@@ -194,26 +194,33 @@ it's evaluated at zero. Note that the syntax `a(vals)` is equivalent to
...
 
     for (i,a_coeff) in enumerate(a.coeffs)
         iszero(a_coeff) && continue
-        tmp = vals[1]^(ct[i][1])
+        @inbounds tmp = vals[1]^(ct[i][1])
         for n in 2:N
-            tmp *= vals[n]^(ct[i][n])
+            @inbounds tmp *= vals[n]^(ct[i][n])
         end
         suma += a_coeff * tmp
```
dpsanders: I think you need `abs(a_coeff)` here?

lbenet: Why? The actual coefficient is `a_coeff`...

dpsanders: Yes, sorry, you're right -- we're just trying to bound the polynomial.

lbenet:
> Why don't you just iterate over the elements of `ct` and then over the elements of `ct[i]`?

Your proposal is worth thinking, which I haven't done yet. Anyway, as a fast answer to your question, the hash table `coeff_table` is an `Array{Array{Int,1},1}`, which contains the powers of each variable contained in each monomial of the `HomogeneousPolynomial`; the needed vector is stored in `ct`. The implementation iterates over those monomials, and the evaluation requires to know which power goes with which variable; this is done in the inner-most loop.

lbenet:
@mforets Thanks for cleaning up the commented code I left there. Do you think there is something more to be done here?

cc: @dpsanders

src/intervals.jl
```diff
@@ -0,0 +1,159 @@
...
+    ct = coeff_table[a.order+1]
+    @inbounds suma = a[1]*Interval{T}(0,0)
+
+    Ieven = Interval{T}(0,1)
+    for (i,a_coeff) in enumerate(a.coeffs)
+        iszero(a_coeff) && continue
+        @inbounds tmp = iseven(ct[i][1]) ? Ieven : dx[1]
+        for n in 2:N
+            @inbounds vv = iseven(ct[i][n]) ? Ieven : dx[1]
+            tmp *= vv
```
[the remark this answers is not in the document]
dpsanders: Isn't the answer for the product always `[-1, 1]`?

src/evaluate.jl
```diff
@@ -194,26 +194,33 @@ it's evaluated at zero. Note that the syntax `a(vals)` is equivalent to
...
     R = promote_type(T,S)
     suma = zero(R)
 
     for (i,a_coeff) in enumerate(a.coeffs)
         iszero(a_coeff) && continue
-        tmp = vals[1]^(ct[i][1])
+        @inbounds tmp = vals[1]^(ct[i][1])
         for n in 2:N
-            tmp *= vals[n]^(ct[i][n])
+            @inbounds tmp *= vals[n]^(ct[i][n])
```
dpsanders: This loop could be written equivalently as `prod(vals .^ ct[i])`

src/evaluate.jl
```diff
@@ -249,17 +256,8 @@ function evaluate(a::TaylorN{T}, vals::NTuple{N,S}) where
...
-                tmp *= vals[n]^(coeff_table[homPol][i][n])
-            end
-            sun += a_coeff * tmp
-        end
-        suma[homPol] = sun
+    @inbounds for homPol in length(a):-1:1
+        suma[homPol] = evaluate(a.coeffs[homPol], vals)
     end
 
     return sum( sort!(suma, by=abs2) )
```
dpsanders: What is the need for the sort here?

src/evaluate.jl
```diff
@@ -301,19 +291,11 @@ function evaluate(a::TaylorN{Taylor1{T}}, vals::NTuple{N, Taylor1{T}}) where
...
-        for (i, a_coeff) in enumerate(a.coeffs[homPol].coeffs)
-            iszero(a_coeff) && continue
-            tmp = vals[1]^(coeff_table[homPol][i][1])
-            for n in 2:N
-                tmp *= vals[n]^(coeff_table[homPol][i][n])
-            end
-            suma += a_coeff * tmp
-        end
+    for homPol in length(a):-1:1
+        suma += evaluate(a.coeffs[homPol], vals)
```
dpsanders: There seem to be several functions with the same content. Could they be unified?

src/intervals.jl
```diff
@@ -0,0 +1,159 @@
...
+            for k in a.order-2:-2:1
+                suma1 += a[k]
+            end
+            suma2 = a[end-1]
+            for k in a.order-3:-2:2
+                suma2 += a[k]
+            end
+        end
+    end
+    return a[0] + suma1*dx + suma2 * (one(dx) * (0..1))
```
dpsanders: I think the last term should be written as Interval{T} (0, 1) for type stability. The one should not be necessary.

src/intervals.jl
```diff
@@ -0,0 +1,159 @@
+using .IntervalArithmetic
+
+function evaluate(a::Taylor1, dx::Interval)
+    dx == (-1..1) && return _evaluate(a, dx, Val(true))
```
dpsanders: It seems like it is inefficient to be checking this equality each time.

src/intervals.jl
```diff
@@ -0,0 +1,159 @@
...
+        suma += evaluate(a.coeffs[homPol], dx)
+    end
+
+    return suma
+end
+
+function evaluate(a::HomogeneousPolynomial, dx::IntervalBox{N,T}) where {T, N}
+
+    @assert N == get_numvars()
+    dx == IntervalBox(-1..1, Val(N)) && return _evaluate(a, dx, Val(true))
```
dpsanders: This check is probably quite slow. Can we record somewhere if the TaylorSeries or TaylorModel is normalized and call the specialised version?

src/evaluate.jl
lbenet: Thanks for the suggestion. I will change it as suggested.

src/evaluate.jl
lbenet: The short answer is to keep precision. There is no canonical Horner rule for multivariable polynomials, as far as I know, so what @blas-ko did was to save the result of each monomial evaluation in a vector, which is sorted before adding its elements. In that way, the sum is computed from the smallest number (wrt `abs2`) to the largest. This is certainly not the fastest way of doing it, but it warranties to maintain precision.

src/evaluate.jl
lbenet: You are right, and that is something I have already considered. I propose to open a new issue on that, and leave it for another PR.

src/intervals.jl
lbenet: I fully agree. Do you have any suggestion which avoids calling directly `_evaluate(a, dx, Val(true))` or `_evaluate(a, dx, Val(false))`? I think this was also mentioned by @mforets, but I haven't come with any better solution...

src/intervals.jl
lbenet: The `one(dx)` was introduced to assure type stability :-) 

I will change it as suggested, thanks!

src/intervals.jl
lbenet: I agree with the observation; yet, I see no other way of branching to the specialized method. 

Regarding your question, for a `TaylorModelN` this is already in the struct; for a "general" `TaylorN` (or `HomogeneousPolynomial`) I see no point to include it.

src/intervals.jl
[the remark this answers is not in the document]
lbenet: > Isn't the answer for the product always `[-1, 1]`?

No, though at some point in the history of this PR I though exactly the same. Consider `x^3 * y` and `x^2 * y^2`, with both `x` and `y` evaluated in `-1 ..1`. Both terms are of order 4, so the same parity. Yet, the range of the first one is `-1 .. 1` (`dx[1]`), whereas the second one is `0 .. 1`.

Thinking about this now, perhaps only for even orders the check needs to be performed; for odd orders the result seems always to be `dx[1]`...

src/evaluate.jl
lbenet: Done in [03e448e](/JuliaDiff/TaylorSeries.jl/pull/195/commits/03e448edd524aedfbfa8dc844621dbccd0d8260d).

src/intervals.jl
lbenet: Implemented in [03e448e](/JuliaDiff/TaylorSeries.jl/pull/195/commits/03e448edd524aedfbfa8dc844621dbccd0d8260d).

lbenet:
@dpsanders Thanks for the comments and the review. Some of your suggestions have been implemented in [03e448e](/JuliaDiff/TaylorSeries.jl/pull/195/commits/03e448edd524aedfbfa8dc844621dbccd0d8260d); I've left open the others...

dpsanders:
What I meant was that a TaylorModel should have a field `normalized::Bool`, and a specialized evaluation method could be defined for TaylorModels (as opposed to TaylorSeries), where you only check this Bool, rather than checking if the the whole IntervalBox is equal to IntervalBox(-1..1, n).

mforets:
```julia
julia> t = (0..0) + Taylor1(3)
julia> f(x) = 0.1 * x^3 - 0.5*x^2+1

julia> evaluate(f(t), 0..6) # same as in master, unchanged in this branch (*)
[-17, 4.60001]

julia> evaluate(normalize_taylor(f(t), 0..6, true), -1..1) 
[-2.60001, 4.60001]

julia> evaluate(normalize_taylor(f(t), 0..6, false), 0..1) # wrong?
[1, 4.60001]
```

I think that the evaluation on the interval `0..1` is wrong, because the polynomial has a minimum at `-0.8531`. I couldn't figure out where does the problem come from.

((*) I made a point earlier that from a usage point of view it's nice if `evaluate(...)`, just gives a good interval without having to call `normalize`, but ok.)

lbenet:
@mforets Thanks for the catch! 

I just found the problem! It has to do with the distributive property... I just pushed a new commit that should solve it.

mforets:
Very good. From my side nothing to add. Thanks for the hard work 🏆

lbenet:
I noticed a subtle error which was very hidden, so I added more tests. Now the code is a bit more simplified. In the case of `Taylor1` evaluated at -1..1 it yields a sharper bounds using essentially Horner's rule for the even and odd powers separately. I also noticed that for `Taylor1` evaluated at 0..1 the sum is identical to the usual Horner rule, so I deleted the `if`.

lbenet:
@dpsanders @mforets I think this is ready for review.

src/intervals.jl
lbenet: For the time being I'll leave it like this. In another PR we could address this point.

lbenet:
Shall I go ahead and merge this?

dpsanders:
Yes please go ahead!

lbenet:
Done! Thanks!

mforets:
Would you mind tagging a new release of `TaylorSeries` that contains the new code in this branch?

lbenet:
I'll tag a minor version!

lbenet:
I tagged v0.9.1, and it has already being merged in METADATA.

mforets:
Great, thanks!

PR #206, opened 2019-04-11
Add linear_polynomial

lbenet:
This adds `linear_polynomial` to extract the linear part of the polynomial (returning a polynomial of the same type) for `Taylor1` and `TaylorN`, including tests.


lbenet:
@dpsanders This is ready.

lbenet:
Merging.

Issue #199, opened 2019-03-06
Rounding Not Defined on TaylorSeries


jessebett:
Rounding on `Taylor1{T}` could be defined to return a new `Taylor1` whose coefficients are rounded to the appropriate type.
i.e.
```
t = Taylor1(1.,3) #type Taylor1{Float64}
Float32(t) # Method Error
# expect == Taylor1(Float32(1.),3)
```

lbenet:
Thanks for reporting. 

While the following is not the nicest way, I think it does what you are expecting:
```julia
julia> t = Taylor1(1.,3)
 1.0 + 𝒪(t⁴)

julia> Taylor1(Float32.(t[:]))
 1.0 + 𝒪(t⁴)

julia> typeof(ans)
Taylor1{Float32}
```

Another option is to work from the beginning with coefficients that are `Float32`, simply through:
```julia
julia> t = Taylor1(Float32, 3)
 1.0 t + 𝒪(t⁴)

julia> typeof(ans)
Taylor1{Float32}
```

dpsanders:
It would certainly be nice to be able to do `Float32.(t)` (note the `.`; I don't think it makes sense to define `Float32(t)` ). 

For this I think it's necessary to tap into the broadcasting machinery itself by defining `broadcasted`; see e.g. [here](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/9a38d80f5eb67da5d8d721fcc0060ec162c55164/src/multidim/arithmetic.jl#L26) for how it's done for `IntervalBox`.

jessebett:
I'm actually still just chasing down bugs trying to get these types to work through broadcasting of Flux's tracked functions. This was the next bug I ran into, that `Float32` is being called during the evaluation of the `Dense` layer. So while it probably makes sense to have this require `Float32.(t)` it would have the unfortunate effect that I don't think it would play nice with dropping `Taylor1`s where others have other number types.

lbenet:
As @dpsanders mentions, this could be solved by changing the behavior of `eltype` to fix broadcasting; see #198. 

I am currently a bit overloaded, so I hope to find some time to look this up next week; I'm truly sorry. In any case, if you have some progress, let us know.

lbenet:
@jessebett See #200 (branch is `lb/broadcasted`); it fixes the specific case that you described above. Can you play with that to see if it works as expected?

[...]

PR #207, opened 2019-04-20
Broadcasting

lbenet:
This implements broadcasting following the [Julia documentation](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration-1), motivated by [this comment](https://github.com/JuliaIntervals/TaylorModels.jl/issues/39#issuecomment-484923751).


test/broadcasting.jl
```diff
@@ -0,0 +1,18 @@
...
+@testset "Broadcasting for Taylor1 expansions" begin
+    t = Taylor1(Int, 5)
+
+    @test 1.0 .+ t == 1.0 + t
+    @test typeof(1.0 .+ t) == Taylor1{Float64}
+    @test 1.0 .+ [t] == [1.0 + t]
+    @test typeof(1.0 .+ [t]) == Vector{Taylor1{Float64}}
+    @test 1.0 .+ [t, 2t] == [1.0 + t, 1.0 + 2t]
+    @test [1.0,2.0] .+ [t 2t] == [1.0+t 1.0+2t; 2.0+t 2.0+2t]
+    @test [1.0] .+ t == 1.0 + t
```
dpsanders: This seems to be different from the behaviour with numbers:
```
julia> [1.0] .+ 1

1-element Array{Float64,1}:
 2.0
```
By analogy it seems that the result should be `[1.0 + t]`.

lbenet: Thanks for checking this! Indeed, you are right. I am trying to find my way in this broadcasting, and it is tricky, so it will go slowly. Hold a bit longer... :-)

jessebett: So to clarify that test `@test [1.0] .+ t == 1.0 + t` should be `@test [1.0] .+ t == [1.0 + t]`

lbenet: Thanks for insisting on this. I agree with both of you that the it should return a `Vector`. For the moment I am trying first to make things work as they did before, so there is no problem with packages that depend on `TaylorSeries.jl`, and then get into this subtleties. But the process is a bit rough for the moment...

lbenet:
Thanks @dpsanders and @jessebett for your comments. 

Somehow, I got to a dead end in my previous attempt, so I've just redesigned everything. Now, all tests pass locally (but not in travis) as well as the case you noted above (included in the tests), i.e., `[1.0] .+ t` returns the *vector*  `[1.0 .+ t]`; that was really tricky. Something that is still not working is broadcast assignment (`.=`).

With regards to #199, `Float32(t)` is still throwing the same error. I could define specific `broadcasted` rules in these cases, so things like `Float32.(t)` would return a `Taylor1{Float32}`. Another approach could be simply extend `Float32(::Taylor1{T})`. Please let me know what's your opinion on this.

lbenet:
Things are slowly converging here, though things like `Float32(t)` is not yet solved. 

@dpsanders Can you take another look into this?

lbenet:
@dpsanders @jessebett Do you have any comments on this PR?

lbenet:
@dpsanders I added the possibility to have `Float32.(t)` return what it should (so `Taylor1(Float32.(t.coeffs))`, using `broadcasted`, as suggested in #199. Other functions could be added as well following similar patterns. Yet, I would be reluctant to allow too much, simply because `sin.(t)` yields the same as `sin(t)`, which I think is fine.

I think this is ready for review.

cc @jessebett

lbenet:
@dpsanders What shall we do about this?

dpsanders:
I think we should merge it. Shall I go ahead?

lbenet:
Thanks! Yes, please, go ahead! I am not sure if #214 will require a rebase.

lbenet:
Sorry, I just noticed that #214 is already merged. Then, simply go ahead!

dpsanders:
Thanks a lot!

lbenet:
@dpsanders Maybe this is a good time to release v0.10.0; do you agree?

lbenet:
I will prepare things to support Julia 1.0 and beyond.

dpsanders:
Yes definitely!

lbenet:
See #216

PR #210, opened 2019-05-30
Add quick examples to README.md

MasonProtter:
Personally, when I visit a package repo I want to see at least some simple code examples right away. Currently, one has to first navigate to the docs, and then from the main docs page to the user guide before they ever see a code example.

This PR adds some flashy demos (taken from the docs) to the REAME so that potential users can see what an awesome package this is right from the start.


lbenet:
You indeed have a point! Thanks for taking the time to improve this!

lbenet:
Thanks a lot again! I'm merging this right away, since tests should simply pass.

MasonProtter:
My pleasure!

PR #216, opened 2019-06-03
Add Project.toml, remove REQUIRE and test/REQUIRE

lbenet:
Support Julia >= 1


lbenet:
@dpsanders Test coverage has decreased, I guess due to some changes in Coverage itself. I'll leave this to other time, so we can release a new version asap. So, I'll merge.

PR #222, opened 2019-09-05
Test (travis) using Julia 1.2 and 1.3 too

lbenet:


.travis.yml
```diff
@@ -7,6 +7,8 @@ os:
 julia:
     - 1.0
     - 1.1
```
ararslan: We could probably replace 1.1 with 1.2 since 1.1 has been effectively superseded by 1.2.

lbenet: Thanks for taking a look on this. So, do you suggest to only test Julia for 1.2, 1.3 and nightly?

ararslan: I'd do 1.0 (long term support release), 1.2 (current stable), 1.3 (current unstable), and nightly

lbenet: Thanks for the suggestion; I'll push a new commit following it.

lbenet:
Thanks @ararslan for your review! 

Merging.

PR #226, opened 2019-09-13
Mathematically consistent big O notation

### LINKED ISSUE #225: Confused by behaviour of addition
```julia
julia> Taylor1(1)
 1.0 t + 𝒪(t²)

julia> Taylor1(Float64,10)^100
 0.0 + 𝒪(t¹¹)

julia> Taylor1(1) + (Taylor1(Float64,10)^100)
 1.0 t + 𝒪(t¹¹)
```

I would have expected `(t + O(t^2)) + O(t^11) == t + O(t^2)`.

lbenet:
Fixes #225

### DIFF
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -47,7 +47,9 @@ t = shift_taylor(0.0) # Independent variable `t`
 ```
 
 Note that the information about the maximum order considered is displayed
-using a big-𝒪 notation. In some cases, it is desirable to not display
+using a big-𝒪 notation. The convention followed when different orders are
+combined is consistent with the mathematics and the big-𝒪 notation, that is,
+to propagate the lowest order. In some cases, it is desirable to not display
 the big-𝒪 notation. The function [`displayBigO`](@ref) allows to
 control whether it is displayed or not.
 ```@repl userguide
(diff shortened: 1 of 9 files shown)


lbenet:
I am checking now the consequences of the changes here in `TaylorIntegration.jl` and `TaylorModels.jl`...

dlfivefifty:
Looks good to me

lbenet:
Thanks @dlfivefifty for taking a look on this. 

I may delay a bit merging this into master because I'm checking that everything is still working in TaylorIntegration.jl and TaylorModels.jl. I'm reviewing the conversion and promotion methods, to check a bit more thoroughly that mixtures of Taylor1 and TaylorN as well as nested Taylor1's behave as they should. These, strange beasts (which if I recall were the reason to make the structs subtypes of `Number`, are important for us to apply "jet transport" techniques.

lbenet:
@dpsanders I think this should be ready now.

dpsanders:
Sorry for the delay. LGTM, thanks!

lbenet:
Merging...

lbenet:
@dpsanders Shall we tag a new patch version, now this was merged?

dpsanders:
Probably this needs a new minor version since it's a pretty important change.

lbenet:
This has been done, we are in v0.10.1 :-)

ebelnikola:
Hi!

I think the use of big-O notation is still a little bit inconsistent. This patch makes it so that a constant `Taylor1{Float64}(c)` behaves as a series of the form $c + 0t + 0t^2 + \ldots$, in the sense that `Taylor1{Float64}(c) + Taylor1(n)` is of order `n` and not zero. On the other hand, the function `getcoeff` still treats `Taylor1{Float64}(c)` as a finite series, and `getcoeff(Taylor1{Float64}(c), 1)` throws an error. And this is reasonable, as `getcoeff` cannot tell whether `order = 0` because it is a constant or due to any other reason. Wouldn't it be better, for full mathematical consistency, to define a special case for constants with the order parameter equal to infinity?

PR #227, opened 2019-09-29
Initialize coefficients of mutated array in mutating functions

PerezHz:
On current master, arrays which store the results of in-place methods such as `mul!`, require those arrays to be initialized to zero in order to work properly:

```julia
julia> using TaylorSeries

julia> order = 2
2

julia> a = Taylor1(rand(order+1))
 0.3901307590706329 + 0.9302337261932201 t + 0.6610315422490114 t² + 𝒪(t³)

julia> b = Taylor1(rand(order+1))
 0.2612832916798675 + 0.00877818753697479 t + 0.6191465524314967 t² + 𝒪(t³)

julia> c = Taylor1(rand(order+1)) # this array stores the results of mul!
 0.5731001108245433 + 0.21748961305056236 t + 0.08638238016455047 t² + 𝒪(t³)

julia> d = a*b # this is the out-of-place product of a and b
 0.1019346489155403 + 0.2464791709784575 t + 0.42243037784084925 t² + 𝒪(t³)

julia> for i = 0:order
           TaylorSeries.mul!(c, a, b, i)
       end

julia> c == d
false
```
Instead, if we initialize the coefficients of `c` to zero, then things work as expected:
```julia
julia> using TaylorSeries

julia> order = 2
2

julia> a = Taylor1(rand(order+1))
 0.36955349820946815 + 0.6054725203327813 t + 0.996916828899657 t² + 𝒪(t³)

julia> b = Taylor1(rand(order+1))
 0.7519229599490191 + 0.9713889338097239 t + 0.23777103622352547 t² + 𝒪(t³)

julia> c = Taylor1(zeros(order+1)) # this array stores the results of mul!
 0.0 + 𝒪(t³)

julia> d = a*b # this is the out-of-place product of a and b
 0.27787576023317784 + 0.8142488682677665 t + 1.4256230769956608 t² + 𝒪(t³)

julia> for i = 0:order
           TaylorSeries.mul!(c, a, b, i)
       end

julia> c == d
true
```

This PR changes the internals of mutating functions `mul!`, `div!`, `pow!`, `sqrt!`, so that the storing array `c` is not required to be initialized to zero:

```julia
julia> using TaylorSeries

julia> order = 2
2

julia> a = Taylor1(rand(order+1))
 0.5223686459210033 + 0.590802594088681 t + 0.762529408631156 t² + 𝒪(t³)

julia> b = Taylor1(rand(order+1))
 0.15529320411399827 + 0.8192967906969988 t + 0.22329571472836762 t² + 𝒪(t³)

julia> c = Taylor1(rand(order+1))
 0.4577969424667405 + 0.6157155512378607 t + 0.44432155422776276 t² + 𝒪(t³)

julia> d = a*b
 0.08112030075376325 + 0.5197225829987081 t + 0.7191009845124225 t² + 𝒪(t³)

julia> for i = 0:order
           TaylorSeries.mul!(c, a, b, i)
       end

julia> c == d
true
```

I tried to change things as little as possible without adding too much overhead, so performance shouldn't be changed too much, and offline benchmarks with `TaylorIntegration.@taylorize` seem to confirm this, i.e., performance is essentially maintained. This PR was indeed inspired by working with `TaylorIntegration.@taylorize`, where these methods are used intensively, and should simplify life working with it.

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -291,7 +291,12 @@ for T in (:Taylor1, :TaylorN)
     @eval @inline function mul!(c::$T{T}, a::$T{T}, b::$T{T}, k::Int) where {T}
 
         # c[k] = zero( a[k] )
-        @inbounds for i = 0:k
+        if $T == Taylor1
+            c[k] = a[0] * b[k]
+        else
+            mul!(c[k], a[0], b[k])
+        end
+        @inbounds for i = 1:k
             if $T == Taylor1
                 c[k] += a[i] * b[k-i]
             else
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -495,7 +495,7 @@ term of the denominator.
     end
 
     imin = max(0, k+ordfact-b.order)
-    c[k] = c[imin] * b[k+ordfact-imin]
+    @inbounds c[k] = c[imin] * b[k+ordfact-imin]
     @inbounds for i = imin+1:k-1
         c[k] += c[i] * b[k+ordfact-i]
     end
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -291,10 +291,10 @@ for T in (:Taylor1, :TaylorN)
     @eval @inline function mul!(c::$T{T}, a::$T{T}, b::$T{T}, k::Int) where {T}
 
         # c[k] = zero( a[k] )
-        @inbounds if $T == Taylor1
-            c[k] = a[0] * b[k]
+        if $T == Taylor1
+            @inbounds c[k] = a[0] * b[k]
         else
-            mul!(c[k], a[0], b[k])
+            @inbounds mul!(c[k], a[0], b[k])
         end
         @inbounds for i = 1:k
             if $T == Taylor1
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -290,7 +290,6 @@ end
 for T in (:Taylor1, :TaylorN)
     @eval @inline function mul!(c::$T{T}, a::$T{T}, b::$T{T}, k::Int) where {T}
 
-        # c[k] = zero( a[k] )
         if $T == Taylor1
             @inbounds c[k] = a[0] * b[k]
         else
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -291,7 +291,7 @@ for T in (:Taylor1, :TaylorN)
     @eval @inline function mul!(c::$T{T}, a::$T{T}, b::$T{T}, k::Int) where {T}
 
         # c[k] = zero( a[k] )
-        if $T == Taylor1
+        @inbounds if $T == Taylor1
             c[k] = a[0] * b[k]
         else
             mul!(c[k], a[0], b[k])
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -494,8 +494,9 @@ term of the denominator.
         return nothing
     end
 
-    @inbounds for i = 0:k-1
-        k+ordfact-i > b.order && continue
+    imin = max(0, k+ordfact-b.order)
+    c[k] = c[imin] * b[k+ordfact-imin]
+    @inbounds for i = imin+1:k-1
         c[k] += c[i] * b[k+ordfact-i]
     end
     if k+ordfact ≤ b.order
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -191,8 +191,12 @@ for T in (:Taylor1, :TaylorN)
                 @inbounds c[0] = exp(constant_term(a))
                 return nothing
             end
-
-            @inbounds for i = 0:k-1
+            if $T == Taylor1
+                @inbounds c[k] = k * a[k] * c[0]
+            else
+                @inbounds mul!(c[k], k * a[k], c[0])
+            end
+            @inbounds for i = 1:k-1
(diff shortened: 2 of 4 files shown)


lbenet:
I think this was made like this with the rational that `c` is "initialized properly". In the examples, you are initializing `c` as `c = Taylor1(rand(order+1))`. If, however, you do it as `c = zero(b)`, everything is consistent:
```julia
julia> a = Taylor1(rand(order+1))
 0.3941964028550935 + 0.5567245605299305 t + 0.8295906090479761 t² + 𝒪(t³)

julia> b = Taylor1(rand(order+1))
 0.19410008277968682 + 0.49260704676615075 t + 0.12788959707098568 t² + 𝒪(t³)

julia> c = zero(b)
 0.0 + 𝒪(t³)

julia> d = a*b
 0.07651355442562842 + 0.3022442091406317 t + 0.48568366664226503 t² + 𝒪(t³)

julia> for i = 0:order
                TaylorSeries.mul!(c, a, b, i)
          end

julia> c == d
true
```

PerezHz:
Exactly, but if then the coefficients of `a` and/or `b` are changed, and we wish to compute the product `a*b` again, then with the current implementation we should either re-initialize `c` to zero, or allocate a new Taylor1 initialized to zero, in order to compute the new product. My proposal is precisely to first set `c[k]` inside `mul!` as
```julia
c[k] = a[0]*b[k]
```
and then do
```julia
for i in 1:k
    c[k] += a[i]*b[k-i]
end
```
(and equivalent things for `div!`, etc.). That way, `c` is not required to be initialized to zero, so we may re-use it as it is to compute other products, while avoiding setting `c.coeffs` to zero or allocating a new Taylor1 variable.

lbenet:
Is this ready for review?

PerezHz:
> Is this ready for review?

Yes, it is now ready for review, thanks!

PerezHz:
I went on and made corresponding changes in other mutating functions such as `exp!`, `log!`, `sincos!`, etc.

lbenet:
Thanks a lot for your contribution! I'll take a look on this now.

src/arithmetic.jl
```diff
@@ -489,8 +493,9 @@ term of the denominator.
         return nothing
     end
 
-    @inbounds for i = 0:k-1
-        k+ordfact-i > b.order && continue
+    imin = max(0, k+ordfact-b.order)
+    @inbounds c[k] = c[imin] * b[k+ordfact-imin]
+    @inbounds for i = imin+1:k-1
```
lbenet: Very nice!

src/functions.jl
```diff
@@ -191,34 +191,44 @@ for T in (:Taylor1, :TaylorN)
...
             return nothing
         end
 
         @inline function log!(c::$T{T}, a::$T{T}, k::Int) where {T}
             if k == 0
                 @inbounds c[0] = log(constant_term(a))
                 return nothing
+            elseif k == 1
+                @inbounds c[1] = a[1] / constant_term(a)
+                return nothing
```
lbenet: I think this is not really needed (the same result is obtained from the rest of the code below), though it may actually speed things up a little. Have you done some benchmarks on its usefulness?

src/power.jl
```diff
@@ -200,15 +200,10 @@ coefficient of `a`.
         return nothing
     end
 
-    # Relevant for positive integer r, to avoid round-off errors
-    if isinteger(r) && (k-l0 > r*findlast(a) > 0)
-        @inbounds c[k-l0] = zero(c[0])
-        return nothing
-    end
-
```
lbenet: I can't remember if these lines are important or not?

src/arithmetic.jl
PerezHz: Thanks! I actually took inspiration from `pow!`, where similar things were already implemented to optimize a similar loop

src/functions.jl
PerezHz: I'll do some benchmarks and report back!

src/power.jl
PerezHz: Good catch! Sorry yeah, they do look important for integer powers; I'll put them back

src/functions.jl
PerezHz: I just did some benchmarks and performance looks essentially the same for `Taylor1{Float64}`. On the other hand, for `Taylor1{Taylor1{Float64}}` performance is a bit better with the `elseif k == 1 ...` block

lbenet: Thanks for reporting. Using nested Taylor1's is a good test for benchmarking. Then, I think it is worth leaving it there.

PerezHz:
@lbenet thank you for your review! I just pushed some commits addressing the points you brought up, although there seems to be an issue on travis (see the relevant line in the build log [here](https://travis-ci.org/JuliaDiff/TaylorSeries.jl/jobs/592877851#L270)) with julia nightly (Linux) related to matrix multiplications...

src/power.jl
lbenet: Thanks!

lbenet:
I just restarted travis, so let's wait and see. Those tests, from time to time, do not pass; yhey use some randomly generated matrices and it is difficult to see if this is a rounding-off problem or what...

lbenet:
Tests are passing now. :-)

lbenet:
Thanks again! I'm merging!

PerezHz:
Thanks for merging!

PR #229, opened 2019-10-16
Custom printing for Taylor1

### LINKED ISSUE #211: Custom printing for Taylor1
Currently, with `TaylorN` you can choose to use whatever string you want for the power series variable printing, however it seems that `Taylor1`s are stuck with the letter `t` for their printing. It'd be nice if this was customizable.

lbenet:
Fixes #211

... by adding a new global variable.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -50,6 +50,7 @@ isapprox
 isfinite
 displayBigO
 use_show_default
+set_taylor1_varname
 ```
 
 ## Internals
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -25,8 +25,6 @@ using Requires
 using LinearAlgebra: norm, mul!,
     lu, lu!, LinearAlgebra.lutype, LinearAlgebra.copy_oftype,
     LinearAlgebra.issuccess
-    # istriu, triu!, istril, tril!, UpperTriangular, LowerTriangular,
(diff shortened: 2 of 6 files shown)


dpsanders:
So you can't do a Taylor1 of a Taylor1 with different variable names? It seems to me like we could just store the variable name inside the Taylor1 object.

lbenet:
Thanks for the remarks!

Indeed, you can't have different symbols which still makes those objects pretty confusing. There is some work on nested `Taylor1`s (see #126) though it has some problems; I have some (old) implementation which I haven't pushed, which would allow to distinguish those variables nicely. Perhaps I should also push that for evaluation.

I also thought about adding a fieldname in the struct as you propose. The question is what would we do when we have an operation (say addition) involving two `Taylor1` objects with different display symbols? Independently of what the answer to that question will be, we would need to check the field of the displayed symbols, which would impact negatively the performance. (Perhaps, the solution is to promote the result to a `Taylor1{Taylor1{T}}` since we are actually talking about two distinct independent Taylor1 variables, but this is also debatable.) 

So I simply decided not to get into those issues.

lbenet:
@dpsanders Do you agree to go ahead and merge this?

lbenet:
I'll go ahead and merge this.

dpsanders:
Sorry for the delay, thanks for merging!

PR #236, opened 2020-02-27
Reimplement pow! and div! for Taylor1s

lbenet:
The first non-zero element is computed in pow!, instead of being passed as an argument.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
-version = "0.10.1"
+version = "0.10.2"
 
 [deps]
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-IntervalArithmetic = "0.16"
+IntervalArithmetic = "0.15, 0.16"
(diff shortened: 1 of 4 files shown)


lbenet:
In TaylorIntegration.jl, we observed that for some equations involving `^`, the use of `pow!` (through the macro `@taylorize`) was problematic. An example is the Duffing oscillator, which involves a power 3, using as initial condition one component being zero. The problem arose
because the macro assumed a default option (`l0=0` in the previous implementation), which was incorrect. This PR solves that. I think it is ready to be merged.

cc @PerezHz @dpsanders

lbenet:
See https://github.com/PerezHz/TaylorIntegration.jl/issues/99 for an explanation of the problem.

dpsanders:
I don't think I quite understand the problem or the solution but feel free to go ahead and merge!

lbenet:
The problem (for `TaylorSeries.jl`) can be summarized as follows: for the computation of `c=a^r` (`a` and `c` being `Taylor1`s) we need to compute the first non-zero term of `a`. This was done withing the method `^(a::Taylor1, r)`, which calls `pow!` and uses/passes that information. In `TaylorIntegration.jl` we call `pow!` directly (by some function created by the macro `@taylorize`, which does improve performance), but had no way on knowing the first non-zero term of `a`, so we used some default value. For a `Taylor1` object whose first entry (constant term) was zero, this lead to some `NaN`s. The solution is to change a little bit `^(a::Taylor1, r)` and `pow!`, so the first non-zero term is computed inside `pow!`. This is what this PR implements.

I hope this explanation makes a little more sense.

lbenet:
@dpsanders What do you suggest, do we release a patch or a minor version?

dpsanders:
Good catch!

As far as I understand you are not changing pre-existing functionality, only adding something extra? In that case, a patch version.

lbenet:
I think everything is now ready.

lbenet:
I'll go ahead and merge this.

lbenet:
Mmm, I just noticed something odd in travis... I'll check this tomorrow.

Issue #240, opened 2020-04-03
Supporting a static Taylor1 or TaylorN type?


mewilhel:
I was just wondering if anyone has thoughts around possibly supporting statically typed versions of Taylor series in this package. I've thrown together a sample implementation and it seems like there is a reasonable speed up to be had.

```julia
addition speedup: 63.0x
subtraction speedup: 63.0x
scalar multiplication speedup: 14.0x
multiplication speed up: 10.0x
```

Sample Snippet:

```julia
import Base: +, -, *

struct STaylor1{N,T}
    coeffs::SVector{N,T}
end

+(x::STaylor1{N,T}, y::STaylor1{N,T}) where {T<:Number,N} = STaylor1{N,T}(x.coeffs + y.coeffs)
-(x::STaylor1{N,T}, y::STaylor1{N,T}) where {T<:Number,N} = STaylor1{N,T}(x.coeffs - y.coeffs)
*(x::STaylor1{N,T}, y::T) where {T<:Number,N} = STaylor1{N,T}(y*x.coeffs)
*(y::T, x::STaylor1{N,T}) where {T<:Number,N} = STaylor1{N,T}(y*x.coeffs)

 @generated function *(x::STaylor1{N,T}, y::STaylor1{N,T}) where {T<:Number,N}
    ex_calc = quote end
    for j = 1:N
        ex_line = :(x.coeffs[1]*y.coeffs[$j])
        for k = 2:j
            ex_line = :($ex_line + x.coeffs[$k]*y.coeffs[$(j-k+1)])
        end
        sym = Symbol("c$j")
        ex_line = :($sym = $ex_line)
        push!(ex_calc.args, ex_line)
    end
    exout = :((c1,))
    for i = 2:N
        sym = Symbol("c$i")
        push!(exout.args, sym)
    end
    return quote
            $ex_calc
            tup = $exout
            svec = SVector{N,T}(tup)
            return STaylor1{N,T}(svec)
         end
end
```

Benchmarking script:

```julia
using StaticArrays, BenchmarkTools, TaylorSeries

order = 5
a = 1.1

s1 = @SVector rand(order)
s2 = @SVector rand(order)

st1 = STaylor1{order,Float64}(s1)
st2 = STaylor1{order,Float64}(s2)

out1 = Taylor1(Float64[s1[i] for i in 1:length(s1)])
t1 = Taylor1(Float64[s1[i] for i in 1:length(s1)])
[...]

PR #241, opened 2020-04-10
Preliminary work on static analogs for Taylor1

mewilhel:
#240 

- Includes the definition of `STaylor1` structure and constructors.
- Adds the following functions for the `STaylor1` type: `+`, `-`. `*`, `exp`, `zero`, `one`, `evaluate`, `==`, functionality to iterate and get indices.

I've added a small benchmarking library for individual operators. This was mostly to provide a means to confirm that none of the operators with `STaylor1` objects are allocating.

This doesn't include promotion for different types such as `+(STaylor{N,T}, S)` or a number of other operators.  I figured it was probably better to get some feedback before going too much further.


benchmark/benchmarks.jl
```diff
@@ -0,0 +1,67 @@
...
+
+S = SUITE["arithmetic"] = BenchmarkGroup()
+for i in dims
+    r = rand(i)
+    x = Taylor1(r)
+    x2 = Taylor1(r)
+    q = STaylor1(r)
+    q2 = STaylor1(r)
+    y = rand()
+    S["Taylor1{$i,Float64} + Float64"] = @benchmarkable f(+, $x, $y, $n)
```
dpsanders, line 38: You should be able to use metaprogramming to script these tests.

src/arithmetic.jl
```diff
@@ -31,11 +33,23 @@ end
 for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
     @eval iszero(a::$T) = iszero(a.coeffs)
 end
+function iszero(a::STaylor1)
```
dpsanders, line 36: `all(iszero, a.coeffs)`  perhaps?

src/arithmetic.jl
```diff
@@ -205,6 +219,48 @@ end
 -(a::Taylor1{T}, b::TaylorN{S}) where {T<:NumberNotSeries,S<:NumberNotSeries} =
     -(promote(a,b)...)
 
+@inline +(a::STaylor1{N,T}, b::STaylor1{N,T}) where {N, T<:Number} = STaylor1(add_tuples(a.coeffs, b.coeffs))
```
dpsanders, line 222: Can you just use `STaylor1(`a.coeffs .+ b.coeffs`)`?

src/arithmetic.jl
```diff
@@ -225,6 +281,38 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
...
+     return quote
+                Base.@_inline_meta
+                $ex_calc
+                return STaylor1{N,T}($exout)
+             end
+end
+
+
+function *(a::STaylor1{N,T}, b::T) where {N, T<:Number}
+    STaylor1{N,T}(scale_tuple(a.coeffs, b))
```
dpsanders, line 310: Again, should just be `a.coeffs .* b`

src/auxiliary.jl
```diff
@@ -317,3 +346,46 @@ linear_polynomial(a::TaylorN) = TaylorN([a[1]])
 linear_polynomial(a::Vector{T}) where {T<:Number} = linear_polynomial.(a)
 
 linear_polynomial(a::Number) = a
+
+
+#=
+Tuple operations taken from ForwardDiff.jl
```
dpsanders, line 352: I think all of these can be replaced by simple broadcasts these days.

src/evaluate.jl
```diff
@@ -30,18 +31,41 @@ function evaluate(a::Taylor1{T}, dx::S) where {T<:Number, S<:Number}
 end
 evaluate(a::Taylor1{T}) where {T<:Number} = a[0]
 
+function evaluate(a::STaylor1{N,T}, dx::T) where {N, T<:Number}
```
dpsanders, line 34: Maybe some functions like this can be defined only once for both `Taylor1` and `STaylor1`. 
Is there an `AbstractTaylor1` type?

dpsanders:
Check here for an alternative approach to many of these functions:

https://github.com/dpsanders/StaticTaylorSeries.jl/blob/203f79543688a4c32e29434c0da945aea9ce4614/src/StaticTaylorSeries.jl#L70

using the `ntuple` function.
If this is as performant then I think it's preferable.

src/arithmetic.jl
mewilhel: Done!

src/arithmetic.jl
mewilhel: Done!

src/arithmetic.jl
mewilhel: Done!

src/evaluate.jl
mewilhel: I've looked into this a little bit. I haven't added an AbstractTaylor1 type yet. 

We could do this for the evaluate(a::Taylor1{T}) function (just have both functions call get_order) but I haven't been able to figure out a way to avoid this for most other functions.

dpsanders:
Sorry, this fell under the radar.

I personally think that the separate `StaticTaylorSeries.jl` package would be a more suitable place for this.

Are you still working in this direction?

mewilhel:
@dpsanders No worries. I had to divert my attention elsewhere for a quite while. The separate package idea works for me. 

I'm currently throwing a separate package together and I should have the preliminary work done by the end of this week.

dpsanders:
@mewilhel I was meaning this package that I wrote quite a while ago that has a lot of overlapping functionality:

https://github.com/dpsanders/StaticTaylorSeries.jl

mewilhel:
@dpsanders Ah! Sorry, I haven't looked at this in a while. That sounds good too. I'll fork a copy of that repository and transfer some of the content there.

mewilhel:
@dpsanders I've created a pull request for StaticTaylorSeries.jl: https://github.com/dpsanders/StaticTaylorSeries.jl/pull/2.

PR #243, opened 2020-05-21
Fix update! for TaylorN

UzielLinares:
The current implementation of this function gives `DimensionMismatch` error when the order of the updating polynomial doesn't match the order specified with the `set_variables` function, e.g.

```julia
set_variables([:x, :y], order=6)

xT = TaylorN(Float64, 1, order=3)
yT = TaylorN(Float64, 2, order=3)

update!(xT, [1., 0.]) # gives error
```
This PR fix this behavior by using the same order of the updating polynomial in the `update!` function internals.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
-version = "0.10.3"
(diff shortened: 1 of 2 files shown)


lbenet:
Good catch!

lbenet:
Can you update Project.toml, so we can release release a new patch version?

lbenet:
Let's wait for the tests, and then I merge and tag a new version. Thanks again!

lbenet:
Merging! Thanks for contributing to our package!

PR #245, opened 2020-05-24
Add two-arg `atan` method and corresponding tests

PerezHz:
This PR adds two-argument methods for `atan` and tests, which hopefully is useful to have in `TaylorSeries.jl`. Things seem to be working for `Taylor1` and `TaylorN`, but I'm not sure if a new in-place `atan!` method is needed as well?

### DIFF
--- a/src/functions.jl
+++ b/src/functions.jl
@@ -119,6 +119,12 @@ for T in (:Taylor1, :TaylorN)
             return c
         end
 
+        function atan(a::$T, b::$T)
+            c = atan(a/b)
+            c[0] = atan(constant_term(a), constant_term(b))
+            return c
+        end
+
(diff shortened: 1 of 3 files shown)


lbenet:
Thanks a lot for the PR! Great idea to have such implementation.

About  your question: `atan!` will be needed if you want to use the two-arg function together with `@taylorize`. This can certainly be addressed in a separate PR, or here, as you prefer.

Let me know when  this PR is ready for review.

PerezHz:
Thanks @lbenet! I mean, if it's not strictly necessary to have `atan!(y,x)` implemented for `atan(y,x)` to work, then perhaps we can do it in another PR; so this is ready for review

PR #249, opened 2020-09-15
Improve evaluation (Horner  methods) of Taylor1s with intervals

lbenet:


lbenet:
Current behavior in master is the following:

```julia
julia> using  TaylorSeries, IntervalArithmetic

julia> t = Taylor1(2)
 1.0 t + 𝒪(t³)

julia> p = 1+t^2
 1.0 + 1.0 t² + 𝒪(t³)

julia> p(-1 .. 2)
[-1, 5]

julia> p(-2 .. 2)
[-3, 5]

julia> q = (1+t)^2
 1.0 + 2.0 t + 1.0 t² + 𝒪(t³)

julia> q(-1..2)
[-3, 9]

julia> q(-2..2)
[-7, 9]
```

With this PR we get:
```julia
julia> p(-1 .. 2)  # tightest result!
[1, 5]

julia> p(-2 .. 2)  # tightest result!
[1, 5]

julia> q(-1..2) # tightest result is [0,9]
[-1, 9]

julia> q(-2..2) # tightest result is [0,9]
[-3, 9]
```

While the proposed implementation does not yield the tightest evaluation always (due to the dependency problem in Horner's evaluation), it does improve w.r.t. the current behavior. This is specially important for TaylorModels.

cc @UzielLinares

UzielLinares:
Hi, these changes break some tests on `TaylorModels.jl`, eg,

Screenshot from 2020-09-17 10-43-46

lbenet:
Thanks for reporting! Regarding the concrete test you point out, I just checked it and it actually "improves" the remainder that is defined in that test. I guess the proper change is to check the containment of remainder of the computed TaylorModel with respect to what it is given (in the test). There are other tests that are also broken, some of them related to the display, so it is not tragic to amend them.

I'll try to look into this later this afternoon...

lbenet:
I will go ahead and merge this, and check the tests in TaylorModels separately.

PR #250, opened 2020-10-01
For ^(a::Taylor1, n::Integer) fall back to a^float(n)

lbenet:


lbenet:
This solves a subtle type unstability, and a problem with TaylorIntegration.jl tests (in Julia 1.5)

dpsanders:
Aren't integer powers more efficient than float powers? Or are they the same in the end?

lbenet:
Thanks for asking, I also thought that too. 

Let me begin saying that in Julia base (v1.4 and 1.5) something like that is also done, in the sense that (in general) it relies on `ccall("llvm.pow.f64", llvmcall, Float64, (Float64, Float64), x, Float64(y))`. To answer your question, the code I am replacing (which was type unstable), for n>3 also computed `a^float(n)`. The idea was to internally use `pow!`. Recently, we noticed that some test in `TaylorIntegration.jl` were not passing (in Julia 1.5), and the culprit was related to this.

lbenet:
@dpsanders Shall I update Project.toml for a new patch version? (I don't think it is worth a minor version, since it fixes a type unstability.)

lbenet:
I'll go ahead and merge...

PR #253, opened 2020-12-02
Fix zero for TaylorN{Taylor1}

PerezHz:
When `t` is a `TaylorN{Taylor1}}`, `zero(t)` does not preserve the order of the coefficients of `t`, which are themselves `Taylor1`s. This PR attempts to fix this and adds corresponding tests. Example:

```julia
t = Taylor1(3)
xHt = HomogeneousPolynomial([one(t), zero(t)])
yHt = HomogeneousPolynomial([zero(t), t])
tN1 = TaylorN([HomogeneousPolynomial([t]), xHt, yHt^2])
zero(tN1).coeffs # the order of each coefficient (Taylor1) should be 3, but it's 1
```

@lbenet: do you think similar changes should be made for `one`?

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -33,8 +33,9 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
 end
 
 ## zero and one ##
-for T in (:Taylor1, :TaylorN), f in (:zero, :one)
-    @eval ($f)(a::$T) = $T(($f)(a[0]), a.order)
+for T in (:Taylor1, :TaylorN)
+    @eval one(a::$T) = $T(one(a[0]), a.order)
+    @eval zero(a::$T) = $T(zero.(a.coeffs))
 end
 
 function zero(a::HomogeneousPolynomial{T}) where {T<:Number}
@@ -42,11 +43,11 @@ function zero(a::HomogeneousPolynomial{T}) where {T<:Number}
     return HomogeneousPolynomial(v, a.order)
 end
 
-function zeros(::HomogeneousPolynomial{T}, order::Int) where {T<:Number}
-    order == 0 && return [HomogeneousPolynomial([zero(T)], 0)]
+function zeros(a::HomogeneousPolynomial{T}, order::Int) where {T<:Number}
+    order == 0 && return [HomogeneousPolynomial([zero(a[1])], 0)]
     v = Array{HomogeneousPolynomial{T}}(undef, order+1)
     @simd for ord in eachindex(v)
-        @inbounds v[ord] = HomogeneousPolynomial([zero(T)], ord-1)
+        @inbounds v[ord] = HomogeneousPolynomial([zero(a[1])], ord-1)
     end
     return v
 end
@@ -278,9 +279,9 @@ function *(a::HomogeneousPolynomial{T}, b::HomogeneousPolynomial{T}) where
 
     order = a.order + b.order
 
-    order > get_order() && return HomogeneousPolynomial(zero(T), get_order())
+    order > get_order() && return HomogeneousPolynomial(zero(a[1]), get_order())
 
-    res = HomogeneousPolynomial(zero(T), order)
+    res = HomogeneousPolynomial(zero(a[1]), order)
     mul!(res, a, b)
     return res
 end
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -51,7 +51,7 @@ function zeros(a::HomogeneousPolynomial{T}, order::Int) where {T<:Number}
     order == 0 && return [HomogeneousPolynomial([zero(a[1])], 0)]
     v = Array{HomogeneousPolynomial{T}}(undef, order+1)
     @simd for ord in eachindex(v)
-        @inbounds v[ord] = HomogeneousPolynomial([zero(a[1])], ord-1)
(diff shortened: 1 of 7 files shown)


lbenet:
Thanks for spotting this  out! Indeed,  this is a quite subtle and intricate bug. I would naively say that yes, this should also be implemented for `one`.

Have you checked all occurrences of things like `zero(T)` or `one(T)`, which are used to create a `Taylor1`, `TaylorN`, or a `HomogeneousPolynomial`. My guess is that the problem you spotted is somewhat linked to this. I don't yet know, but perhaps we need certain specialized methods for mixtures...

lbenet:
By  some reason, the tests (travis) are not triggered. Any thoughts?

PerezHz:
Tried to access travis via the build badge, and it says that the travis builds have been migrated to travis-ci.com, so maybe the build triggers have to be activated there? Not sure if that's the solution, though...

lbenet:
Thanks for checking this; I'll try  to look if I can see what to do. 

Note that Travis does run in TaylorIntegration and there is no huge difference in the yml files...

lbenet:
Thanks for rebasing! That has triggered the tests!

lbenet:
@PerezHz Can you rebase to current master (again)? There are conflicts... Is this  ready?

lbenet:
@PerezHz Ups, it seems I forgot about this PR... Could you rebase to master to test again before merging?

PerezHz:
> @PerezHz Ups, it seems I forgot about this PR... Could you rebase to master to test again before merging?

Rebased to current master; let's see how CI reacts!

src/arithmetic.jl
```diff
@@ -33,20 +33,25 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
...
 end
 
-function zeros(::HomogeneousPolynomial{T}, order::Int) where {T<:Number}
-    order == 0 && return [HomogeneousPolynomial([zero(T)], 0)]
+function zeros(a::HomogeneousPolynomial{T}, order::Int) where {T<:Number}
+    order == 0 && return [HomogeneousPolynomial([zero(a[1])], 0)]
     v = Array{HomogeneousPolynomial{T}}(undef, order+1)
     @simd for ord in eachindex(v)
-        @inbounds v[ord] = HomogeneousPolynomial([zero(T)], ord-1)
+        @inbounds v[ord] = HomogeneousPolynomial([zero(a[1])], ord-1)
```
lbenet, line 54: In this case you can avoid creating  the vector `[zero(a[1])]`

src/calculus.jl
```diff
@@ -124,8 +124,9 @@ to the `r`-th variable.
 """
 function derivative(a::HomogeneousPolynomial, r::Int)
     @assert 1 ≤ r ≤ get_numvars()
+    # @show zero(a[1]) zero(eltype(a))
     T = eltype(a)
-    a.order == 0 && return HomogeneousPolynomial([zero(T)], 0)
+    a.order == 0 && return HomogeneousPolynomial([zero(a[1])], 0)
```
lbenet, line 129: You can avoid creating the vector here.

src/evaluate.jl
```diff
@@ -247,7 +246,7 @@ function _evaluate(a::TaylorN{T}, vals::NTuple, ::Val{false}) where {T<:Number}
 end
 
 function evaluate(a::TaylorN{T}, vals::NTuple{N,Taylor1{S}}) where
-        {T<:Number, S<:NumberNotSeries, N}
+        {T<:NumberNotSeries, S<:NumberNotSeries, N}
```
lbenet, line 249: I think `Number` is correct here, or am I missing something?

src/power.jl
```diff
@@ -250,12 +250,10 @@ for T in (:Taylor1, :TaylorN)
 end
 
 function square(a::HomogeneousPolynomial)
-    T = eltype(a)
-
     order = 2*a.order
-    order > get_order() && return HomogeneousPolynomial([zero(T)], get_order())
+    order > get_order() && return HomogeneousPolynomial([zero(a[1])], get_order())
```
lbenet, line 254: You can avoid creating the vector here.

src/power.jl
```diff
@@ -250,12 +250,10 @@ for T in (:Taylor1, :TaylorN)
...
 
-    res = HomogeneousPolynomial([zero(T)], order)
+    res = HomogeneousPolynomial([zero(a[1])], order)
```
lbenet, line 256: ... and also here.

lbenet (commented):
Thanks a lot! Sending few minor comments...

src/evaluate.jl
PerezHz: Thanks, yes I think this should be `Number`

src/arithmetic.jl
PerezHz: Changed, thanks!

src/calculus.jl
PerezHz: Changed, thanks!

src/power.jl
PerezHz: Changed, thanks!

src/power.jl
PerezHz: Done

lbenet (approved):

lbenet:
Thanks a lot! LGTM, so I'm merging!

PR #257, opened 2020-12-08
CI and documentation using GitHub actions

lbenet:


lbenet:
While the docs seems to have been built properly, it seems that coveralls is causing problems...  So I'll comment those lines (and remove travis).

lbenet:
Grrrr.... windows is causing headaches.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,53 @@
...
+      - master
+  pull_request:
+    branches:
+      - master
+    tags: '*'
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
```
PerezHz, line 18: To allow failures on nightly the line `continue-on-error: ${{ matrix.version == 'nightly' }}` could be added here

lbenet: Good catch! Thanks a lot!

lbenet:
There are still some issues related with `continue-on-error`, but it seems that this is the way this keyword [currently behaves](https://github.com/actions/toolkit/issues/399). The current implementation marks the nightly version as passing (green), despite nightly jobs are failing; see https://github.com/JuliaDiff/TaylorSeries.jl/actions/runs/413834591, in the annotations section, to note that the nightly versions do not pass. Details can  be found by opening the specific runs.

I think this is ready.

Issue #261, opened 2021-02-04
Support `asinh, acosh, atanh`.


Jollywatt:
The three inverse hyperbolic functions `asinh, acosh, atanh` are the only trig functions yet to be implemented.

Once these are defined on series, Julia’s full suite of standard trig functions

    sin, cos, tan, asin, acos, atan, sinh, cosh, tanh, asinh, acosh, atanh, csc, sec, cot, acsc, asec, acot, csch, sech, coth, acsch, asech, acoth

will be supported, thanks to the default definitions like `acsch(x::Number) = inv(asinh(x))` in `Base`.

This should be an easy fix. (I would make a PR myself, but the actual code in `functions.jl` is a bit obscure…)

lbenet:
Thanks for opening this issue! You are right, we should add those functions. It would be great if you can make the PR!

At the moment I can't guide you through what exactly one has to do, but I'll try to get back on this on Monday. Let me only say that you should first obtain the *recurrence relation* for the specific elementary function, such as those provided [in the docs](http://www.juliadiff.org/TaylorSeries.jl/stable/background/#Elementary-functions-of-polynomials). 

How do you do that? You first propose an ordinary differential equation (initial value problem) whose solution is the function you look for, e.g., `asinh(g(t))`, where `g(t)` is a (Taylor) polynomial. You propose a polynomial solution, which you solve *order by order*; that precisely yields the recurrence relation. I suggest that you try it first for the `exp`  function, because it  is simpler and it clarifies how to do it (and the answer is there), and then get to the more difficult problems.

Regarding the obscurity of the code (truly sorry!), this is related to the possibility of calling the functions that compute the recurrence relations directly. This is important for us in [TaylorIntegration.jl](https://github.com/PerezHz/TaylorIntegration.jl) to optimize the performance.

mforets:
> This should be an easy fix. (I would make a PR myself, but the actual code in functions.jl is a bit obscure…)

I would encourage that you don't get intimidated with the code and ask direct questions about the package internals and such -- so you can contribute and learn thins along the way ;) The code patterns being used in that file are not uncommon: (i) implement an in-place version of the function for efficiency, but expose the out-of-place version; (ii) use macros to reduce the LOC (for the latter you may find the section on [code generation](https://docs.julialang.org/en/v1/manual/metaprogramming/#Code-Generation) useful).

Jollywatt:
[...]

PR #262, opened 2021-02-05
Added inverse hyperbolic functions (#261)

Jollywatt:
* Added `asinh`, `acosh`, `atanh` with tests. This completes support for
Julia's suite of standard trig functions.

* Commented-out a seemingly moot test asserting that the number
of `_dict_binary_ops` entries be `22`.

* Minor typo in docs.

### DIFF
--- a/src/dictmutfunct.jl
+++ b/src/dictmutfunct.jl
@@ -115,11 +115,11 @@ const _dict_unary_ops = Dict(
         :(_aux = sinh(_arg1))],
     :tanh => [:tanh!, (:_res, :_arg1, :_aux, :_k), :(_res = tanh(_arg1)),
         :(_aux = tanh(_arg1)^2)],
-    :asinh => [:asin!, (:_res, :_arg1, :_aux, :_k), :(_res = asinh(_arg1)),
+    :asinh => [:asinh!, (:_res, :_arg1, :_aux, :_k), :(_res = asinh(_arg1)),
         :(_aux = sqrt(_arg1^2 + 1))],
-    :acosh => [:acos!, (:_res, :_arg1, :_aux, :_k), :(_res = acosh(_arg1)),
+    :acosh => [:acosh!, (:_res, :_arg1, :_aux, :_k), :(_res = acosh(_arg1)),
         :(_aux = sqrt(_arg1^2 - 1))],
-    :atanh => [:atan!, (:_res, :_arg1, :_aux, :_k), :(_res = atanh(_arg1)),
+    :atanh => [:atanh!, (:_res, :_arg1, :_aux, :_k), :(_res = atanh(_arg1)),
         :(_aux = 1 - _arg1^2)],
 );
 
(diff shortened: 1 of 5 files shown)


Jollywatt:
(I see "Some checks were not successful," but not sure the failed check is related to this PR. When I view the details, it reports that the `documenter/deploy` job succeeded in 3m 20s.)

src/dictmutfunct.jl
```diff
@@ -115,6 +115,12 @@ const _dict_unary_ops = Dict(
         :(_aux = sinh(_arg1))],
     :tanh => [:tanh!, (:_res, :_arg1, :_aux, :_k), :(_res = tanh(_arg1)),
         :(_aux = tanh(_arg1)^2)],
+    :asinh => [:asin!, (:_res, :_arg1, :_aux, :_k), :(_res = asinh(_arg1)),
+        :(_aux = sqrt(_arg1^2 + 1))],
+    :acosh => [:acos!, (:_res, :_arg1, :_aux, :_k), :(_res = acosh(_arg1)),
+        :(_aux = sqrt(_arg1^2 - 1))],
+    :atanh => [:atan!, (:_res, :_arg1, :_aux, :_k), :(_res = atanh(_arg1)),
+        :(_aux = 1 - _arg1^2)],
```
lbenet, line 123: The first argument in these functions is the name (as a symbol) of the (internal) function that deals with them. So they should be `:asinh!`, `:acosh!` and `:atanh!`, respectively. (My apologies for the lack of proper docs for this dictionaries.)

lbenet (commented):
Thanks for this PR, and sorry for the long delay to get into it. I only spotted one simple typo (in `src/dictmutfunct.jl`) which should be very simple to correct. It would also be nice if you could add come docstrings (at the end of `src/functions.jl`). Aside from that, LGTM!

lbenet:
> (I see "Some checks were not successful," but not sure the failed check is related to this PR. When I view the details, it reports that the documenter/deploy job succeeded in 3m 20s.)

I think that error is related to the deployment of the docs, which somehow does not work when the PR comes from a fork.

Jollywatt:
@lbenet In addition to the typos and docs you suggested, I made the `ArgumentError`s thrown by the inverse trig functions more informative (see last commit). E.g.:
```julia
julia> asin(Taylor1([1, 0]))
ERROR: ArgumentError: Series expansion of asin(x) diverges at x = ±1
```

(Should these be `DomainError`s instead, mirroring the way Julia handles, e.g.,
```julia
julia> asin(2)
ERROR: DomainError with 2.0:
asin(x) is not defined for |x|>1.
```
? The rest of `TaylorSeries` uses `ArgumentError`s.)

lbenet:
Thanks for considering my suggestions for this PR! LGTM!

> Should these be DomainErrors instead, mirroring the way Julia handles

Good point! Indeed, we throw `ArgumentError`s when we should be throwing `DomainError`s. Would you like to change that in this PR, or would you prefer doing it in a new one? (This issue also applies to other functions, such as `sqrt` or `log`.)

Jollywatt:
I think I better open a new PR to do that, after this one is merged!

Jo

lbenet (approved):
LGTM! I'll go ahead and merge this! 

Thanks a lot

PR #267, opened 2021-03-10
Use DomainError instead of ArgumentError with several functions

lbenet:


lbenet:
Following [this suggestion](https://github.com/Jollywatt).

PR #271, opened 2021-04-09
Replace a == zero(a) with iszero(a) in pretty printing

matthieubulte:
Printing was crashing when using `TaylorSeries` with `Symbolics` coefficients because if `x` is a symbolic expression, the Julia expression `x == zero(x)` translates to the symbolic expression `x == zero(x)` instead of a boolean. Using `iszero` instead of `x == zero(x)` does the trick.


lbenet:
Thanks a lot for addressing this! LGTM! Merging!

Issue #276, opened 2021-05-29
More clear documentation for basic users


KronosTheLate:
As a less competent user, both in terms of programming and math, I found the documentation pretty confusing and unhelpful. The first time the function I was looking for (`taylor_expand()`) was mentioned, was in the _end_  of the  "User guide/One independent variable" section. I feel like a basic example could be given much earlier.

Then there is the other way of doing things, where one defines a variable and uses that to compute taylor polynomials. This way of thinking and doing things only became clear to me in @dpsanders 's reply in [this discourse thread](https://discourse.julialang.org/t/find-taylor-series-of-a-function/51847/7), and not by the technically heavy documentation. For what it is worth, it seems like whoever created that discourse-post would agree with me that basic examples of how to expand e.g. `exp(x)` are not readily available in the current docs. A simple example like that given in the discourse discussion would be extremely appreciated at an early point.

The conclusion is that I found the documentation very techincal and heavy on math (like proving some identity I have never seen as the first example), and I would like to help remedy the situation. So I would like to write a section along the lines of "Just give me a Taylor series", with some basic usage examples. The problem is that I really do not see how such a section would fit in right now, other than perhaps shoehorned in before "One independent variable". I might give that a go, but the optimal solution would clearly be a more comprehensive rewrite, with basic examples baked into it.

lbenet:
Thanks for open this issue and for proposing having a "Just give me a Taylor series" section in the documentation. I agree with you that the documentation needs to be improved, but I haven't found the time to do it. So your help is truly welcome and appreciated.

KronosTheLate:
Good to hear! I am drawing great value for using all these amazing packages, so it truly is my pleasure to give somehing back.

I have added [this PR](https://github.com/JuliaDiff/TaylorSeries.jl/pull/277), which adds more or less what I was missing. I actually think that if these examples are added, the problem would almost be fixed by a reference to it. Something like `If you just want to compute a simple taylor-series, see the first examples under the section "Examples"` early in the `User guide`.

lbenet:
Closed by #277

PR #277, opened 2021-05-29
Added two basic examples

KronosTheLate:
As described in [this issue](https://github.com/JuliaDiff/TaylorSeries.jl/issues/276), I think the documentation is too advanced for basic users simply wanting some basic usage.  So I have added two basic examples, which would alleviate the issue a lot.

I would certainly have appreciated them greatly when I found this (very well implemented, good job!) package.

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -14,6 +14,7 @@ julia> using TaylorSeries
 julia> taylor_expand(exp, 0, order=3)
  1.0 + 1.0 t + 0.5 t² + 0.16666666666666666 t³ + 𝒪(t⁴)
 ```
+
 If the final `𝒪(t⁴)` information about the error is not interesting, you can make it not print with the `displayBigO` function:
 ```
 julia> displayBigO(false)
@@ -37,6 +38,8 @@ julia> x = Taylor1(3)
 julia> exp(x)
  1.0 + 1.0 t + 0.5 t² + 0.16666666666666666 t³ + 𝒪(t⁴)
 ```
+
+### Changing printing variable
 Even though we are expanding `exp(x)`, the default variable for printing is `t`. This can be set with the function `set_taylor1_varname()`
 
 ```@repl Taylor1_variable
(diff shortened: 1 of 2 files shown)


lbenet:
Thanks a lot for addressing #276 here! The added examples are really clear and nice. And also thanks for your help in improving our package!

If you consider it worth, I would add an example of evaluating the Taylor expansion at a given point, which was the question in the [discourse thread](https://discourse.julialang.org/t/find-taylor-series-of-a-function/51847/7) you quoted. Something like:
```julia
julia> x = Taylor1(3)
 1.0 t + 𝒪(t⁴)

julia> p = exp(1+x)
 2.718281828459045 + 2.718281828459045 t + 1.3591409142295225 t² + 0.45304697140984085 t³ + 𝒪(t⁴)

julia> p(0.01)
2.74560101388203

julia> exp(1.01)
2.7456010150169163
```

lbenet:
One final suggestion, inspired by [your comment](https://github.com/JuliaDiff/TaylorSeries.jl/issues/276#issuecomment-850826456): maybe adding the sentence you suggest in the Readme.md, and linking directly to the section examples there, is worth.

KronosTheLate:
> ```julia
> julia> x = Taylor1(3)
>  1.0 t + 𝒪(t⁴)
> 
> julia> p = exp(1+x)
>  2.718281828459045 + 2.718281828459045 t + 1.3591409142295225 t² + 0.45304697140984085 t³ + 𝒪(t⁴)
> 
> julia> p(0.01)
> 2.74560101388203
> 
> julia> exp(1.01)
> 2.7456010150169163
> ```

I like this example a lot. It should definitely be added!


>One final suggestion, inspired by your comment: maybe adding the sentence you suggest in the Readme.md, and linking directly to the section examples there, is worth.

If you agree with the idea, let's do it. You are more than welcome to edit my PR directly - I will not make your suggested changes right now, as I am busy. If you don't add your suggestions yourself, I'll do it at a later point ^_^

lbenet:
At the moment I'm busy too... we can wait and you complete the example !

KronosTheLate:
> At the moment I'm busy too... we can wait and you complete the example !

Alrighty, I manged to squeeze it in tonight! I felt rude placing the reference to the examples before the nice intro in the `User guide` section, but because I felt that the intro was even a little techincal, I wanted to make sure that the disinterested reader would not fall off before finding the link to the examples. Hope you agree with it, and feel free to make changes!

lbenet:
LGTM. Let us wait that tests pass, and check how the docs are displayed. Then i merge. Thanks a lot!

KronosTheLate:
I see that the PR is merged, but the output is off - it says julia is not defined:
image

I must have messed something up. I don't know how autogenerated docs work, as I have never edited them before. Any idea on what went wrong?

lbenet:
Sorry... I couldn't check the docs because the PR comes from a fork... i think that if you erase the "julia> " in every line of the code, that would avoid the problems; see the `example.md` file further below.

lbenet:
I'm addressing the problem in #278

PR #278, opened 2021-05-29
Fix docs

lbenet:


KronosTheLate:
Looks great! Thanks for the help is making the result good ^_^

I noticed in the [current latest docs](https://juliadiff.org/TaylorSeries.jl/latest/examples/#Examples), the custom function example does not seem to be working, even though the code in the actual .md file looks good to me:
image

If you do some editing to make that work, then would you mind changing the definition of `f(x)` also? It should probably use another variable, because `x` is already the `Taylor1` object.

KronosTheLate:
I just created [this PR](https://github.com/JuliaDiff/TaylorSeries.jl/pull/279) which adresses the too wide output, and the f(x)->f(y) to avoid possible confusion. The printing of the custum function example is not changed, because I have no idea what makes it only print the first line.

PR #279, opened 2021-05-30
Remove printing of BigO

KronosTheLate:
This PR removes the printing of the BigO, because including it makes the output wider than the screen, resulting in a horizontal scrolling bar, which looks pretty bad:
image

It also changes the f(x) definition to f(y), because x is already defined, which might cause confusion.

The custom function example still only prints the first line. I don't know what goes wrong there, but feel free to fix it in this PR it you know how.

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -10,17 +10,14 @@ A keyword argument `order` determines which order to expand to:
 ```@repl taylor_expand
 using TaylorSeries
 
-taylor_expand(exp, 0, order=3)
- 1.0 + 1.0 t + 0.5 t² + 0.16666666666666666 t³ + 𝒪(t⁴)
+taylor_expand(exp, 0, order=2)
 ```
 
-If the final `𝒪(t⁴)` information about the error is not interesting, you can make it not print with the [`displayBigO`](@ref) function:
+If the final BigO-notation error is not interesting, you can make it not print with the [`displayBigO`](@ref) function:
 ```
 displayBigO(false)
-false
 
 taylor_expand(exp, 0, order=3)
- 1.0 + 1.0 x + 0.5 x² + 0.16666666666666666 x³
 ```
 
 ## Expanding exp(x) with a symbolic object
@@ -31,27 +28,22 @@ to expand to:
 ```@repl Taylor1_variable
 using TaylorSeries
 
-x = Taylor1(3)
- 1.0 t + 𝒪(t⁴)
+x = Taylor1(2)
 
 exp(x)
- 1.0 + 1.0 t + 0.5 t² + 0.16666666666666666 t³ + 𝒪(t⁴)
 ```
 
 Let's also get rid of the printed error for the next few examples:
 ```@repl Taylor1_variable
 displayBigO(false)
-false
 ```
 ### Changing printing variable
 Even though we are expanding `exp(x)`, the default variable for printing is `t`. This can be set with the function [`set_taylor1_varname()`](@ref)
 
 ```@repl Taylor1_variable
 set_taylor1_varname("x")
-"x"
 
 exp(x)
- 1.0 + 1.0 x + 0.5 x² + 0.16666666666666666 x³
 ```
 
 ### Changing point to expand about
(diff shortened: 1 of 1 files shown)


KronosTheLate:
Is it possible that the `f (generic function with 1 method)` line messes things up? In the same way that the `julia>` messed things up before. If every line inside the example is run, then that line after `f` is defined will error:

```julia

julia> f(x)=2x

julia> f (generic function with 1 method)
ERROR: syntax: space before "(" not allowed in "f (" at REPL[20]:1
Stacktrace:
 [1] top-level scope
   @ none:1
```

lbenet:
This answers part of [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/278#issuecomment-850963778):

> I noticed in the current latest docs, the custom function example does not seem to be working, even though the code in the actual .md file looks good to me

I guess you mean the fact that there is no "julia> " prompt in the second code block, right? That is easy to remedy; we forgot to add `@repl taylor_expand` after the opening triple "```" for the code block, in line 19. Can fix this here too?

lbenet:
> Is it possible that the f (generic function with 1 method) line messes things up?

Yes! Actually, the in the `@repl` mode for the documentation, there is no need to include the julia output of the commands. If you check the docs further below, in the example of the four square identity (lines 117-124 in the .md file, the first code block of [this section](https://juliadiff.org/TaylorSeries.jl/latest/examples/#Four-square-identity)), only the julia instructions are included, which are then processed and the results are included (`Documenter.jl` takes care of this). 

So my suggestion is to get rid of all the julia output, and let us see how it comes.

KronosTheLate:
There, output removed. Did that fix things?

Is it possible for me to see what the docs would look like if this PR got comitted somehow?

lbenet:
Thanks a lot! I am actually checking the docs locally right now. 

What I am doing is the following. From the TaylorSeries.jl directory (where the Project.toml is), I run a julia repl session:
```julia
julia> cd("docs/")

# Get into the pkg manager
(@v1.6) pkg> activate .

(docs) pkg> instantiate
### (lots of output)

julia> include("make.jl")
[ Info: Precompiling Documenter [e30172f5-a6a5-5a46-863b-614d45cd2de4]
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: ExpandTemplates: expanding markdown templates.
[ Info: CrossReferences: building cross-references.
[ Info: CheckDocument: running document checks.
┌ Warning: 1 docstring not included in the manual:
│ 
│     TaylorSeries.nonlinear_polynomial :: Tuple{AbstractSeries}
│ 
│ 
│ These are docstrings in the checked modules (configured with the modules keyword)
│ that are not included in @docs or @autodocs blocks.
└ @ Documenter.DocChecks ~/.julia/packages/Documenter/bFHi4/src/DocChecks.jl:69
[ Info: Populate: populating indices.
[ Info: RenderDocument: rendering document.
[ Info: HTMLWriter: rendering HTML pages.
┌ Warning: Documenter could not auto-detect the building environment Skipping deployment.
└ @ Documenter ~/.julia/packages/Documenter/bFHi4/src/deployconfig.jl:75
```
Then, you can open in a browser the (local) file: `(TaylorSeries.jl-dir)/docs/build/examples.html `

docs/src/examples.md
```diff
@@ -10,17 +10,14 @@ A keyword argument `order` determines which order to expand to:
...
 using TaylorSeries
 
-taylor_expand(exp, 0, order=3)
- 1.0 + 1.0 t + 0.5 t² + 0.16666666666666666 t³ + 𝒪(t⁴)
+taylor_expand(exp, 0, order=2)
 ```
 
-If the final `𝒪(t⁴)` information about the error is not interesting, you can make it not print with the [`displayBigO`](@ref) function:
+If the final BigO-notation error is not interesting, you can make it not print with the [`displayBigO`](@ref) function:
 ```
```
lbenet, line 17: The next line should read:
```julia
```@repl taylor_expand
```
similar to line 10.

KronosTheLate:
>Thanks a lot! I am actually checking the docs locally right now.

Thanks for the detailed response. I think that is a bit too involved for me, so I will just hope that we finally have something working now ^_^

docs/src/examples.md
lbenet: Thanks!

lbenet (approved):
LGTM! I'll merge this! Thanks a lot!

PR #280, opened 2021-06-01
Smaller fixes

KronosTheLate:
After creating something like a 5th PR on the example documentation, I think I finally have gotten the hang of how the documenter functions work.

This PR clears up some inconsistency (2nd and 3rd order expansion), and fixes the fact that the displayBigO function was set to false, before settinging it to false in the second example. This PR also removed the redundant definition of x halfway through the second example.

A seccond commit will be added that makes the first paragraph in the "User guide" more professional.

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -7,25 +7,21 @@ CurrentModule = TaylorSeries
 ## Expanding exp(x) with `taylor_expand()`
 The [`taylor_expand`](@ref) function takes the function to expand as it's first argument, and the point to expand about as the second argument.
 A keyword argument `order` determines which order to expand to:
-```@repl taylor_expand
+```@repl 1
 using TaylorSeries
 
 taylor_expand(exp, 0, order=2)
 ```
 
-If the final BigO-notation error is not interesting, you can make it not print with the [`displayBigO`](@ref) function:
-```@repl taylor_expand
-displayBigO(false)
-
-taylor_expand(exp, 0, order=3)
+And voìla! It really is that simple to calculate a simple taylor polynomial. The next example is slightly more complicated.
 ```
 
 ## Expanding exp(x) with a symbolic object
 An alternative way to compute the single-variable taylor expansion for a function is by defining a variable of type `Taylor1`,
-and simply using it in the function you wish to expand. The argument given to the [`Taylor1`](@ref) constructor is the order
+and using it in the function you wish to expand. The argument given to the [`Taylor1`](@ref) constructor is the order
 to expand to:
 
-```@repl Taylor1_variable
+```@repl 2
 using TaylorSeries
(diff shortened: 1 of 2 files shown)


KronosTheLate:
I am sorry about creating yet another PR about this documentation, and I hope you are you getting tired of indulging my learning experience with documentation. But because there were several issues that bugged me, I found it necessary to fix them. I think the result is better for it, and I hope you agree ^_^

lbenet:
> I am sorry about creating yet another PR about this documentation, and I hope you are you getting tired of indulging my learning experience with documentation. But because there were several issues that bugged me, I found it necessary to fix them. I think the result is better for it, and I hope you agree ^_^

Thanks again a lot for improving the docs, and I agree that they are nicer! I guess they are ready to be merged, but prefer if you ping me when they are ready; I checked the proposed changes, and they LGTM!

KronosTheLate:
Glad the value added is worth the trouble ^_^

I don't have any other improvements in mind, so it is "full speed ahead" from me! If you are able to generate the docs locally, and just see that everything makes sense and is in the right order, that would be great :D

docs/src/examples.md
```diff
@@ -7,63 +7,55 @@ CurrentModule = TaylorSeries
...
 taylor_expand(exp, 0, order=2)
 ```
 
-If the final BigO-notation error is not interesting, you can make it not print with the [`displayBigO`](@ref) function:
-```@repl taylor_expand
-displayBigO(false)
-
-taylor_expand(exp, 0, order=3)
+And voìla! It really is that simple to calculate a simple taylor polynomial. The next example is slightly more complicated.
 ```
```
lbenet, line 17: The opening triple quotes (` ``` `) in line 17 are causing problems. Simply delete them; they are a leftover of some code which was removed.

docs/src/examples.md
```diff
@@ -7,63 +7,54 @@ CurrentModule = TaylorSeries
...
 
-### More awesome examples
+### More examples
 You can even use custum functions
-```@repl Taylor1_variable
-f(y) = 1/(y+1)
-
-x = Taylor1(3)
+```@repl 2
+f(a) = 1/(a+1)
```
lbenet, line 57: One more comment. In the two repl-blocks code below (lines 63 and 68), you need to have them as ` ```@repl 2` instead of ` ```@repl Taylor1_variable`. Otherwise they don't recognize the function `f` ot de (Taylor1) variable `x`.

KronosTheLate: Good catch! I really thought I changed all instances when I changed the name to `2`, but apparently not.

lbenet (approved):

lbenet:
Thanks a lot! LGTM! Merging!

PR #283, opened 2021-06-24
Add more concrete method for /

### LINKED ISSUE #282: Missing division method `HomogeneousPolynomial`
The problem reported in https://github.com/PerezHz/TaylorIntegration.jl/issues/132 is related to the fact that `/` involving `HomogenousPolynomial` and `Int` is not concrete enough to be properly dispatched, which triggers promotion and then throws the reported error.

cc @dfcaporale

lbenet:
Closes #282


lbenet:
Merging

Issue #281, opened 2021-06-19
Upcoming `lu` issue on nightly


dkarrasch:
On current master, there is a finer distinction in `copy_oftype`-like constructors, which will likely make this package fail on nightly, see

https://github.com/JuliaLang/julia/pull/40831#issuecomment-864385804

This affects your own `lu` implementation, which uses `copy_oftype(A, S)`, which used to call `convert(AbstractArray{S}(A))`, but now relies on 2-arg `similar`. So there are two options to fix this:

1. Write a 2-arg `similar` function for matrices with eltype `Taylor1`, then `copy_oftype` will start using this starting from Julia v1.8 "silently".
2. Replace `copy_oftype(A, S)` in `lu` by a direct call to `convert(AbstractArray{S}, A)`, which is handled internally in your package.

If needed, I'd be willing to help, probably just like @daanhb.

lbenet:
Hi @dkarrasch ! Thanks for reporting the future issue. If you can, we would be grateful if you can help us with this.

daanhb:
Hi, I'm happy to chime in. From the surrounding comments in the code, it seems the goal is to avoid pivoting, which LinearAlgebra would otherwise do by default but it may fail in this case. If that is so, might the following be sufficient?
```julia
lu(A::AbstractMatrix{Taylor1{T}}; check::Bool = true) where {T<:Number} = lu(A, Val(false); bool=bool)
```

This is not quite the same as the current implementation though, because the code tries again with pivoting if the first attempt fails. So perhaps
```julia
function lu(A::AbstractMatrix{Taylor1{T}}; check::Bool = true) where {T<:Number}
    F = lu(A, Val(false); bool=bool)
    if issuccess(F)
        return F
    else
        lu(A, Val(true); bool=bool)
    end
end
````

The advantage of the above is that you avoid using LinearAlgebra internals, but perhaps I'm missing something. If it turns out that LinearAlgebra doesn't properly copy the array, say, then we'll have to look into that.

dkarrasch:
I tested a little bit and found that the issue is the new `lu(::StridedMatrix)` method, which takes precedence over the `lu(::AbstractMatrix{Taylor1})` method defined here. In #41288, I have removed that method assuming (https://github.com/JuliaLang/julia/pull/40831#issuecomment-863261577) that `copy_toarray` and `copy_oftype` is no different on `StridedMatrix`es.

lbenet:
[...]

Issue #242, opened 2020-05-20
Evaluate for TaylorN fails with ModelingToolkit


dpsanders:
```jl
julia> using TaylorSeries

julia> x, y = set_variables("x y");

julia> t = x^2 + y;

julia> using ModelingToolkit

julia> @variables xx, yy;

julia> t([xx, yy])
ERROR: StackOverflowError:
Stacktrace:
 [1] evaluate(::TaylorN{Float64}, ::Tuple{Operation,Operation}) at /Users/dpsanders/.julia/packages/TaylorSeries/XWMlP/src/evaluate.jl:237 (repeats 79984 times)

```

ChrisRackauckas:
Fun

lbenet:
I do reproduce the problem. No idea where to look for the solution...

lbenet:
Yet, it does work with `HomogeneousPolynomial`s:
```julia
julia> x[1]([xx, yy])
0 + 1.0 * (xx ^ 1 * yy ^ 0)
```
So the problem may be related to the Horner implementation.

dpsanders:
I don't think Horner is used for multivariate polynomials?
Shouldn't it just be

```jl
sum(h([x, y]) for h in t)` 
```

to evaluate a `TaylorN`  `t`? 
Or am I missing something?

dpsanders:
Ah you mean that's too naive and there's some complicated extension of Horner for multivariate polynomials.

lbenet:
The way we do it is not really Horner, you are right, but some sort of generalization. If I recall correctly, evaluate separately the homogeneous polynomials of given order, sort them, and then add them. Sorting them is to avoid the usual cancellations.

In any case, the observation holds: it  seems evaluation works for `HomogenousPolynomial`s.

lbenet:
I think I found what the problem and it has nothing to do with the way  we compute the evaluation

`x([xx, yy])` is equivalent to `evaluate(x, [xx, yy])` which uses the method defined [here](https://github.com/JuliaDiff/TaylorSeries.jl/blob/22781aa2f60a107a68e7564a30af12f9e1e2c73b/src/evaluate.jl#L237), so it calls `evaluate(x, (xx, yy))`, which uses the same method again and again, until the stack overflow happens:

```julia
julia> using ModelingToolkit

julia> using TaylorSeries

julia> x, y = set_variables("x", numvars=2);

julia> @variables xx, yy;

[...]

PR #284, opened 2021-06-24
Fix lu deprecation warning on julia v1.7+

dkarrasch:
I'm not sure why you had the InteractiveUtils `subtype` method imported. I found no use of it in the package, so I removed that dependency. I also bumped the patch version. If you don't like any of those "extra" edits, no problem, I'll revert them if preferred.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,10 +1,9 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
-version = "0.11.0"
+version = "0.11.1"
 
 [deps]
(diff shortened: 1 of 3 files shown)


lbenet (approved):

lbenet:
Thanks a lot for cleaning this as well as solving #281!

Regarding `InteractiveUtils`, I guess it was a leftover of some experiments... I was trying to distinguish types which are `Number` but not `AbstractSeries`, which i named `NumberNotSeries`, from the universe of `Number`; my idea was to do this *dynamically*. But it did not worked out when other packages were loaded (say `ModelingToolKit.jl`); see #242. So now `NumberNotSeries` is simply the union of `Real` and `Complex`.

lbenet:
Anyway, I'm merging, and tag the new version. Thanks a lot!

PR #292, opened 2021-11-25
Fix abs2 and abs function for Taylor{Complex{T}}

patkwee:
Implemented `abs` and `abs2` for `Taylor1{Complex}` and `TaylorN{Complex}`. Also tried to implement it for mixtures, though, since I am not that familiar with them and never really used them, I am unsure if the implementation is correct.

### DIFF
--- a/src/other_functions.jl
+++ b/src/other_functions.jl
@@ -90,6 +90,10 @@ for T in (:Taylor1, :TaylorN)
         end
 
         abs2(a::$T) = a^2
+        
+        abs2(x::$T{T}) where {T<:Complex} = real(x)*real(x) + imag(x)*imag(x)
+
+        abs(x::$T{T}) where {T<:Complex} = sqrt(abs2(x))
     end
 end
 
@@ -128,6 +132,12 @@ function abs(a::Taylor1{TaylorN{T}}) where {T<:Real}
         (abs(x) is not differentiable at x=0)."""))
     end
 end
+
+abs2(x::Taylor1{TaylorN{T}}) where {T<:Complex} = real(x)*real(x) + imag(x)*imag(x)
+abs2(x::TaylorN{Taylor1{T}}) where {T<:Complex} = real(x)*real(x) + imag(x)*imag(x)
+
+abs(x::Taylor1{TaylorN{T}}) where {T<:Complex} = sqrt(abs2(x))
+abs(x::TaylorN{Taylor1{T}}) where {T<:Complex} = sqrt(abs2(x))
 
 @doc doc"""
     abs(a)
--- a/src/other_functions.jl
+++ b/src/other_functions.jl
@@ -89,10 +89,8 @@ for T in (:Taylor1, :TaylorN)
             end
         end
 
-        abs2(a::$T) = a^2
-        
-        abs2(x::$T{T}) where {T<:Complex} = real(x)*real(x) + imag(x)*imag(x)
-
+        abs2(a::$T) = real(a)^2 + imag(a)^2
+    
         abs(x::$T{T}) where {T<:Complex} = sqrt(abs2(x))
     end
 end
@@ -133,11 +131,9 @@ function abs(a::Taylor1{TaylorN{T}}) where {T<:Real}
     end
 end
 
-abs2(x::Taylor1{TaylorN{T}}) where {T<:Complex} = real(x)*real(x) + imag(x)*imag(x)
-abs2(x::TaylorN{Taylor1{T}}) where {T<:Complex} = real(x)*real(x) + imag(x)*imag(x)
-
 abs(x::Taylor1{TaylorN{T}}) where {T<:Complex} = sqrt(abs2(x))
 abs(x::TaylorN{Taylor1{T}}) where {T<:Complex} = sqrt(abs2(x))
+abs(x::Taylor1{Taylor1{T}}) where {T<:Complex} = sqrt(abs2(x))
 
 @doc doc"""
     abs(a)
--- a/src/other_functions.jl
+++ b/src/other_functions.jl
@@ -138,9 +138,12 @@ abs(x::Taylor1{Taylor1{T}}) where {T<:Complex} = sqrt(abs2(x))
 @doc doc"""
     abs(a)
 
-Returns `a` if `constant_term(a) > 0` and `-a` if `constant_term(a) < 0` for
+For a `Real` type returns `a` if `constant_term(a) > 0` and `-a` if `constant_term(a) < 0` for
 `a <:Union{Taylor1,TaylorN}`.
-Notice that `typeof(abs(a)) <: AbstractSeries`.
+For a `Complex` type, such as `Taylor1{ComplexF64}`, returns `sqrt(real(a)^2 + imag(a)^2)`. 
+
+Notice that `typeof(abs(a)) <: AbstractSeries` and 
+that for a `Complex` argument a `Real` type is returned (e.g. `typeof(abs(a::Taylor1{ComplexF64})) == Taylor1{Float64}`).
 
 """ abs
 
--- a/test/manyvariables.jl
+++ b/test/manyvariables.jl
@@ -206,6 +206,10 @@ using LinearAlgebra
     @test (rem(1+xT,1.0))[0] == 0
     @test abs(1-xT)  == 1-xT
     @test abs(-1-xT)  == 1+xT
+    @test abs2(im*xT) == abs2(xT)
+    @test abs(im*(1+xT)) == abs(1+xT)
+    @test isapprox(abs2(exp(im*xT)), one(xT))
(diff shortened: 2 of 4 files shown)


lbenet:
Thanks for this PR!

While I agree with your solution, there is still only one case not covered in your implementation, which is the *nested* `Taylor1`s. Using the definitions in `test/mixtures.jl` [beginning here](https://github.com/JuliaDiff/TaylorSeries.jl/blob/26befc5aeed016eddb6e907b22c8b1738cac475e/test/mixtures.jl#L314), we have:
```julia
julia> ti = Taylor1(3)  # internal Taylor `t`
 1.0 t + 𝒪(t⁴)

julia> to = Taylor1([zero(ti), one(ti)], 9) # "external" Taylor `t`
 ( 1.0 + 𝒪(t⁴)) t + 𝒪(t¹⁰)

julia> exp(im*to) == cos(to) + im* sin(to)  # this is ok
true

julia> abs2(exp(im*to)) # the result is incorrect !
  ( 1.0 + 0.0im ) + 𝒪(t⁴) + ( ( 0.0 + 2.0im ) + 𝒪(t⁴)) t + ( - ( 2.0 - 0.0im ) + 𝒪(t⁴)) t² + ( - ( -0.0 + 1.3333333333333333im ) + 𝒪(t⁴)) t³ + ( ( 0.6666666666666666 + 0.0im ) + 𝒪(t⁴)) t⁴ + ( ( 0.0 + 0.26666666666666666im ) + 𝒪(t⁴)) t⁵ + ( - ( 0.08888888888888888 - 0.0im ) + 𝒪(t⁴)) t⁶ + ( - ( -0.0 + 0.025396825396825397im ) + 𝒪(t⁴)) t⁷ + ( ( 0.006349206349206348 + 0.0im ) + 𝒪(t⁴)) t⁸ + ( ( 0.0 + 0.0014109347442680775im ) + 𝒪(t⁴)) t⁹ + 𝒪(t¹⁰)
```

However, the following yields approximately the correct result:
```julia
julia> real(ee)^2+imag(ee)^2
  1.0 + 𝒪(t⁴) + ( 4.336808689942018e-19 + 𝒪(t⁴)) t⁸ + 𝒪(t¹⁰)
```

My suggestion is to simply have [here](https://github.com/JuliaDiff/TaylorSeries.jl/blob/26befc5aeed016eddb6e907b22c8b1738cac475e/src/other_functions.jl#L92) the following code, which I believe is correct for all cases:
```julia
abs2(a::$T) = real(a)^2 + imag(a)^2
```
Admittedly, for non-complex Taylor expansions it computes `imag(a)` which is not needed...

patkwee:
Good suggestion, I changed `abs2` accordingly. Though, I think for the complex `abs` we still need to cover all combinations, since for a Real type, the implementation with `<0` and `>0` is much nicer.

lbenet:
> Good suggestion, I changed `abs2` accordingly. Though, I think for the complex `abs` we still need to cover all combinations, since for a Real type, the implementation with `<0` and `>0` is much nicer.

Good point, and thanks for bringing it up. I'm not quite sure what *all combinations* refers to. 

With this PR, we have:
```julia
julia> t = Taylor1(5)
 1.0 t + 𝒪(t⁶)

julia> abs(1+im*0+t)   # Taylor1{ComplexF64}, but actually purely real
 1.0 + 1.0 t + 𝒪(t⁶)

julia> abs(-1+im*0+t)  # Taylor1{ComplexF64}, but actually purely real
 1.0 - 1.0 t + 𝒪(t⁶)

julia> abs(1+im*0+t) == abs(1+t)
true

julia> abs(-1+im*0+t) == abs(-1+t)
true
```
which seems consistent. Note in pass that the `typeof(abs(::Taylor1{<:Complex}))` is `Taylor1{<:Real}`...

Also, if we use intervals, I think everything behaves consistently:
```julia
julia> using IntervalArithmetic

julia> abs((0.5 .. 1) + t)  # Constant term is purely positive
 [0.5, 1] + [1, 1] t + 𝒪(t⁶)

julia> abs(-(0.5 .. 1) + t) # constant term is purely negative
 [0.5, 1] + [-1, -1] t + 𝒪(t⁶)

julia> abs((-1 .. 1) + t) # constant term contains zero
ERROR: DomainError with  [-1, 1] + [1, 1] t + 𝒪(t⁶):
The 0th order Taylor1 coefficient must be non-zero
(abs(x) is not differentiable at x=0).
```
I haven't though about the case with mixtures...

Once said all this, is the following correct?
```julia
julia> abs(-1+im + t)
 1.4142135623730951 - 0.7071067811865475 t + 0.17677669529663692 t² + 0.08838834764831846 t³ + 0.03314563036811942 t⁴ + 0.005524271728019897 t⁵ + 𝒪(t⁶)
```

lbenet:
I've been doing some checks against gp/pari:
- Our current implementation of `abs` for Taylor polynomials with real coeffs is consistent with gp/pari. That is, if the coefficients are real, it checks the 0-order coefficient, and returns `a`, `-a` or a domain error, depending if it positive, negative or zero.
- Also, in the examples above where we have "fake" complex coefficients, i.e., complex coefficients but whose imaginary part is identically zero, the results are also consistent.
- For "non-fake" complex polynomials, gp/pari returns directly the original polynomial `a`; this is different to the proposal we have here.

In other words, for
```julia
julia> ee = exp(im*Taylor1(5))
 ( 1.0 + 0.0im ) + ( 0.0 + 1.0im ) t - ( 0.5 - 0.0im ) t² - ( -0.0 + 0.16666666666666666im ) t³ + ( 0.041666666666666664 + 0.0im ) t⁴ + ( 0.0 + 0.008333333333333333im ) t⁵ + 𝒪(t⁶)
```
gp/pari returns `ee` itself, while we return `sqrt(abs2(ee))`.

patkwee:
> Though, I think for the complex `abs` we still need to cover all combinations, since for a Real type, the implementation with `<0` and `>0` is much nicer.

Sorry, I should have been more clear. I think for `abs` with complex argument, we need the following type combinations, because I guess we don't want to change the `abs` definition for Real types.

```julia
abs(::Taylor1{<:Complex})
abs(::TaylorN{<:Complex})
abs(::Taylor1{TaylorN{<:Complex}})
abs(::TaylorN{Taylor1{<:Complex}})
abs(::Taylor1{Taylor1{<:Complex}})
```

I put in the `abs(x) = sqrt(abs2(x))` implementation. Not sure if there is a more clever implementation for complex functions. The Julia standard library uses `hypot(real(x), imag(x))`, though I don't know if there is a good `hypot` Taylor series implementation.  

> Once said all this, is the following correct?
> 
> ```julia
> julia> abs(-1+im + t)
>  1.4142135623730951 - 0.7071067811865475 t + 0.17677669529663692 t² + 0.08838834764831846 t³ + 0.03314563036811942 t⁴ + 0.005524271728019897 t⁵ + 𝒪(t⁶)
> ```

I think it is correct:

```julia
julia> t = Taylor1(5)
 1.0 t + 𝒪(t⁶)

julia> p = abs(-1+im + t)
 1.4142135623730951 - 0.7071067811865475 t + 0.17677669529663692 t² + 0.08838834764831846 t³ + 0.03314563036811942 t⁴ + 0.005524271728019897 t⁵ + 𝒪(t⁶)

julia> x = -0.1:1e-4:0.1
-0.1:0.0001:0.1

julia> all(@. abs(-1+im + x) ≈ p(x))
true
```

> * For "non-fake" complex polynomials, gp/pari returns directly the original polynomial `a`; this is different to the proposal we have here.

Interesting. I am not familiar with or used gp/pari, but such a result would surprise me. I would expect that `abs` gives the Taylor series expansion of the absolute value of its (here) complex argument, i.e. a real, positive number as result. For `Float64` / `ComplexF64` Julia standard library also returns `abs(exp(im*x)) == 1.0` (and not `exp(im*x)`). Do you know the rationale behind the gp/pari choice?

lbenet:
Thanks for clarifying what you meant. I'm not sure if all methods are truly needed, but let's leave that for the future.

With regards to my comment on gp/pari, [here](https://pari.math.u-bordeaux.fr/dochtml/html-stable/Transcendental_functions.html#abs) is the documentation for `abs`. While it mentions that for power series the constant coefficient is taken into account, as we do it too, there is no mention to the case complex power series. Once stated this, I do think the intuition we have been discussing here is consistent and correct, which is nicely reflected by the tests. 

One last request: would you mind updating the doc strings of `abs`, so the case with underlying complex coefficients is explained? I'm certainly in favor of merging this PR.

patkwee:
I updated the `abs` docstring. Feel free to propose specific changes, if you don't find it clear enough.

lbenet (approved):
Thanks truly a lot! LGTM, so I'll merge this PR.

patkwee:
I am glad that I could help, `TaylorSeries.jl` is a really nice package!

PR #309, opened 2022-11-23
Fix equation rendering in docstrings

navidcy:


lbenet:
Thanks for this PR! I didn't notice that there are some equations which were not rendered properly. 

Since you are onto this, could you also fix (delete the `{eqnarray*}` environments) for [`TaylorSeries.sqr!`](https://github.com/JuliaDiff/TaylorSeries.jl/blob/28ad329f06e8a50bdc372fc3c57f5939f9d59cfc/src/power.jl#L281) and [`TaylorSeris.sqrt!`](https://github.com/JuliaDiff/TaylorSeries.jl/blob/28ad329f06e8a50bdc372fc3c57f5939f9d59cfc/src/power.jl#L412)?

lbenet:
I deployed the docs locally, and it seems these fixes are not fixing the rendering of those equations. Checking the documentation on `Documenter.jl` [here](https://documenter.juliadocs.org/stable/man/latex/#Display-Equations), it seems we do need the `{aligned}` environments.

navidcy:
Done

lbenet:
I've tried locally and it still doesn't solve the problem. I'm not sure why is that happening; it may be related to the use of `@doc doc"""..."""`, or the fact that it is rendered in html; the docstrings seem ok though.

navidcy:
When I build the docs locally with
```
julia --color=yes --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate()'; julia --color=yes --project=docs docs/make.jl
```
they seem ok.

Screen Shot 2022-11-25 at 4 24 05 am

lbenet:
I can't check it, but let's go ahead and see if it works. So I'll merge.

Thanks for your contribution!

navidcy:
Seems ok!
https://juliadiff.org/TaylorSeries.jl/latest/api/#TaylorSeries.sqr!

lbenet:
Excellent! Thanks again!

PR #319, opened 2023-03-09
Make sure coeff_table is not changed by differentiate or integrate

lbenet:


lbenet:
I've cooked the following MWE:
```julia
using TaylorSeries

const order = 20
const x, y, z, w = set_variables(Int128, "x", numvars=4, order=2order)

function fun(degree::Int)
    s = x + y + z + w + 1
    s^degree
end

ctab = deepcopy(TS.coeff_table); # to check consistency of the coeff_table
f = fun(order);
function check_diff1(f)
    f1 = differentiate(f,1);
    f2 = differentiate(f,2);
    f3 = differentiate(f,3);
    f4 = differentiate(f,4);
    return f1 + f2 + f3 + f4;
end

df = check_diff1(f);
@assert ctab == TS.coeff_table # must be true; uses always 1 thread

dfexact = 4*20*fun(order-1);
@assert df == dfexact # true

g = fun(order);
function check_diff2(f)
    vder = zeros(typeof(f), get_numvars())
    Threads.@threads for i = 1:get_numvars()
        vder[i] = differentiate(f, i)
    end
    return sum(vder)
end

dg = check_diff2(g);
@assert ctab == TS.coeff_table # true, for any number of threads
@assert dg == df # true
```

I haven't checked any benchmarks, but I notice there is indeed some allocations (in comparison to current master) due to the `copy`.

lbenet:
This is essentially ready, but I'd like to be able to test using more threads...

PR #322, opened 2023-05-06
Update examples.md

mforets:
Fix what seems like a typo.


lbenet (approved):
Good catch @mforets, and thanks for correcting it!

Issue #209, opened 2019-05-24
Should `isless` be supported?


dlfivefifty:
I'm under the impression that a `TaylorN` should be thought of like a dual number, but there are some inconsistencies:
```julia
julia> dual(1,2) < 5
true

julia> x = set_variables("x", order=100)[1]; x < 5
ERROR: MethodError: no method matching isless(::TaylorN{Float64}, ::Int64)
Closest candidates are:
  isless(::Missing, ::Any) at missing.jl:70
  isless(::InfiniteArrays.Infinity, ::Real) at /Users/sheehanolver/.julia/packages/InfiniteArrays/24ELy/src/Infinity.jl:47
  isless(::InfiniteArrays.OrientedInfinity{Bool}, ::Number) at /Users/sheehanolver/.julia/packages/InfiniteArrays/24ELy/src/Infinity.jl:146
  ...
Stacktrace:
 [1] <(::TaylorN{Float64}, ::Int64) at ./operators.jl:260
 [2] top-level scope at none:0
```

https://github.com/JuliaApproximation/HypergeometricFunctions.jl/issues/11

lbenet:
Thanks for reporting. You have a point; would it be consistent to define `isless` caring only about the constant term? So in your example, it would return `true`

dlfivefifty:
Yes. Adding that and a few other overrides we can compute Taylor series of Hypergeometric functions: https://github.com/JuliaApproximation/HypergeometricFunctions.jl/issues/11

lbenet:
Excellent! Please go ahead and make the PR! If by some reason you can't, I'll try to do it, but I'm a bit busy now, so it will need to be delayed to next week.

What are the other few overrides that would be needed?

dlfivefifty:
They are listed in that other issue:
```julia
julia> x = set_variables("x", order=100)[1];

julia> Base.isless(y::Number, x::TaylorN) = isless(y, x.coeffs[1].coeffs[1])

julia> Base.log1p(x::TaylorN) = log(1+x)

julia> Base.eps(::Type{TaylorN{T}}) where T = eps(T)

julia> Base.isless(x::TaylorN, y::Number) = isless(x.coeffs[1].coeffs[1], y)

julia> _₂F₁(1.0,2.0,3.0,(x-1))
[...]

PR #323, opened 2023-05-06
Support `<` and `>`  through `isless`

lbenet:
Solves #209 

This PR implements the total order following [this paper](https://www.bmtdynamics.org/pub/papers/adtheory/adtheory.pdf) by M. Berz, "Automatic differentiation as nonarchimedean analysis", in Computer Arithmetic and Enclosure Methods, 439-450, Elsevier (1992).

cc @dlfivefifty, @KeithWM

### DIFF
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -360,7 +360,7 @@ x, y = set_variables("x y", order=10);
 
 show_monomials(2)
 ```
-Then, the following clearly holds:
+Then, the following then holds:
 ```@repl userguide
 0 < 1e8 * y^2 < x*y < x^2  < y < x/1e8 < 1.0
 ```
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -352,12 +352,23 @@ The ordering is consistent with the interpretation that there are infinitessimal
 elements in the algebra; for details see M. Berz, "Automatic Differentiation as
 Nonarchimedean Analysis", Computer Arithmetic and Enclosure Methods, (1992), Elsevier,
 439-450.
+Essentially, the lexicographic order works as follows: smaller order monomials
+are *larger* than higher order monomials; when the order is the same, *larger* monomials
+appear before in the hash-tables; the function [`show_monomials`](@ref).
+```@repl userguide
+x, y = set_variables("x y", order=10);
+
+show_monomials(2)
+```
+Then, the following clearly holds:
+```@repl userguide
+0 < 1e8 * y^2 < x*y < x^2  < y < x/1e8 < 1.0
+```
 
 The elementary functions have also been
 implemented, again by computing their coefficients recursively:
 
 ```@repl userguide
-x, y = set_variables("x y", order=10);
 exy = exp(x+y)
 ```
 
--- a/docs/src/userguide.md
+++ b/docs/src/userguide.md
@@ -105,6 +105,16 @@ of order 5; this is be consistent with the number of known coefficients of the
 returned series, since the result corresponds to factorize `t` in the numerator
 to cancel the same factor in the denominator.
 
+`Taylor1` is also equiped with a total order, provided by overloading [`isless`](@ref).
+The ordering is consistent with the interpretation that there are infinitessimal
+elements in the algebra; for details see M. Berz, "Automatic Differentiation as
+Nonarchimedean Analysis", Computer Arithmetic and Enclosure Methods, (1992), Elsevier,
+439-450. This is illustrated by:
+
+```@repl userguide
+1 > t > 2*t^2 > 100*t^3 >= 0
+```
+
 If no valid Taylor expansion can be computed an error is thrown, for instance,
 when a derivative is not defined at the expansion point, or it simply diverges.
 
@@ -336,6 +346,12 @@ Note that some of the arithmetic operations have been extended for
 [`HomogeneousPolynomial`](@ref); division, for instance, is not extended.
 The same convention used for `Taylor1` objects is used when combining
 `TaylorN` polynomials of different order.
+Both `HomogeneousPolynomial` and `TaylorN` are equiped with a total *lexicographical*
+order, provided by overloading [`isless`](@ref).
+The ordering is consistent with the interpretation that there are infinitessimal
+elements in the algebra; for details see M. Berz, "Automatic Differentiation as
+Nonarchimedean Analysis", Computer Arithmetic and Enclosure Methods, (1992), Elsevier,
+439-450.
 
 The elementary functions have also been
 implemented, again by computing their coefficients recursively:
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -112,7 +112,7 @@ end
 end
 @inline isless(a::TaylorN{Taylor1{T}}, b::TaylorN{Taylor1{T}}) where {T<:NumberNotSeries} = isless(a - b, zero(T))
 
-#= TODO: Nested Taylor1s; needs careful thinking. The following works:
+#= TODO: Nested Taylor1s; needs careful thinking; iss #326. The following works:
 @inline isless(a::Taylor1{Taylor1{T}}, b::Taylor1{Taylor1{T}}) where {T<:Number} = isless(a - b, zero(T))
 # Is the following correct?
 # ti = Taylor1(3)
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -10,7 +10,6 @@
 
 ## Equality ##
 for T in (:Taylor1, :TaylorN)
-
     @eval begin
         ==(a::$T{T}, b::$T{S}) where {T<:Number,S<:Number} = ==(promote(a,b)...)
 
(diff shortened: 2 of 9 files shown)


test/manyvariables.jl
```diff
@@ -133,6 +133,8 @@ end
     @test HomogeneousPolynomial(0,0)  == 0
     @test (@inferred conj(xH)) == (@inferred adjoint(xH))
     @test (@inferred real(xH)) == xH
+    @test xH > yH
```
KeithWM, line 136: It should be a lot more obvious from the test what it is that we are trying to test and why this is the expected result.

test/manyvariables.jl
```diff
@@ -223,6 +225,8 @@ end
     @test eltype(convert(TaylorN{Complex{Float64}},1)) == TaylorN{Complex{Float64}}
     @test TS.numtype(convert(TaylorN{Complex{Float64}},1)) == Complex{Float64}
     @test normalize_taylor(xT) == xT
+    @test 1 > xT > yT > xT^2 > 0
```
KeithWM, line 228: Maybe I'm misunderstanding something, but to me to it makes very little sense that `xT > yT` should hold  true. I would expect the Taylor series to be symmetrical with respect to which variable is put first.

test/onevariable.jl
```diff
@@ -45,6 +45,9 @@ Base.iszero(::SymbNumber) = false
     @test axes(t) == ()
     @test axes([t]) == (Base.OneTo(1),)
 
+    @test 1 > t > t^2
```
KeithWM, line 48: Could you also add a `> 0` or `>= 0` depending on what is expected?

test/onevariable.jl
```diff
@@ -45,6 +45,9 @@ Base.iszero(::SymbNumber) = false
...
+    @test 1 > -t^2 > -t
```
KeithWM, line 49: Would this better as `0 >=? -t^2 > -t > 0`?

src/arithmetic.jl
```diff
@@ -34,6 +33,145 @@ function ==(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
...
+    Enclosure Methods, (1992), Elsevier, 439-450.
+
+---
+
+    <(a::Taylor1{<:Real}, b::Taylor1{<:Real})
+    <(a::TaylorN{<:Real}, b::Taylor1{<:Real})
+
+Return `a - b < zero(b)`.
+"""
+function <(a::Taylor1{<:Real}, b::Real)
```
KeithWM, line 57: I would strongly advise checking for method ambiguities introduced by this change using Aqua.jl.

src/auxiliary.jl
```diff
@@ -268,13 +268,13 @@ end
 # Finds the first non zero entry; extended to Taylor1
 function Base.findfirst(a::Taylor1{T}) where {T<:Number}
     first = findfirst(x->!iszero(x), a.coeffs)
-    isa(first, Nothing) && return -1
+    isnothing(first) && return -1
```
KeithWM, line 271: Good improvement!

src/arithmetic.jl
```diff
@@ -34,6 +33,145 @@ function ==(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
...
+    @eval begin
+        <(a::$T{T}, b::$T{S}) where {T<:Number,S<:Number} = <(promote(a,b)...)
+        <(a::$T{T}, b::$T{T}) where {T<:Number} = a - b < zero(constant_term(a))
+
+        >(a::$T{T}, b::$T{S}) where {T<:Number,S<:Number} = >(promote(a,b)...)
+        >(a::$T{T}, b::$T{T}) where {T<:Number} = a - b > zero(constant_term(a))
+    end
+end
+
+<(a::HomogeneousPolynomial{T}, b::HomogeneousPolynomial{S}) where {T<:Number,S<:Number} = <(promote(a,b)...)
```
KeithWM, line 161: Why is this `<` "down here" with the `>` and not further above? Is the `>` equivalent of this defined?

src/arithmetic.jl
```diff
@@ -34,6 +33,145 @@ function ==(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
     return iszero(a.coeffs) && iszero(b.coeffs)
 end
 
+# Iss: #209
```
KeithWM, line 36: I don't think including an issue number is valuable. `## Inequality ##` would be better, in line with the above.

KeithWM:
> [Coverage Status](https://coveralls.io/builds/59695213)
> 
> Coverage: 95.371% (-1.1%) from 96.47% when pulling **[f466f19](https://github.com/JuliaDiff/TaylorSeries.jl/commit/f466f19230bdc7c0f5c9570bf9ac3f13abaca605) on lb/iss209_Berz** into **[3d443a1](https://github.com/JuliaDiff/TaylorSeries.jl/commit/3d443a1cdbed874341ccd2a58dc3d29b01a10f3f) on master**.

The drop in coverage should be fixed before accepting this PR.

src/arithmetic.jl
```diff
@@ -34,6 +33,145 @@ function ==(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
...
+        return a < zero(a[1])
+    else
+        return -b < zero(a[1])
+    end
+end
+
+
 for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
     @eval iszero(a::$T) = iszero(a.coeffs)
 end
```
KeithWM, line 177: Not your fault, but the `## zero and one ##` comment below should be above this block of metacode.

src/arithmetic.jl
```diff
@@ -34,6 +33,145 @@ function ==(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
...
+    else
+        return a[nz] < zero(b)
+    end
+end
+<(b::Real, a::TaylorN{<:Real}) = >(a, b)
+
+
+"""
+    >(a::Taylor1{<:Real}, b::Real)
+    >(a::Taylor1{<:Real}, b::Real)
```
KeithWM, line 96: should be `TaylorN`

src/arithmetic.jl
```diff
@@ -34,6 +33,145 @@ function ==(a::HomogeneousPolynomial, b::HomogeneousPolynomial)
...
+    if nz == -1
+        return zero(a0) > zero(b)
+    else
+        return a[nz] > zero(b)
+    end
+end
+>(b::Real, a::TaylorN{<:Real}) = <(a, b)
+
+#
+for T in (:Taylor1, :TaylorN,)
```
KeithWM, line 151: If there's metaprogramming anyway, might as well also metaprogram away the `>` and `<` repetition.

KeithWM (changes requested):
Hope you don't mind the review, not that used to giving reviews to people I don't know, but I've done my best to be thorough. Great work in finding a reference in literature to work from. I'd suggest making this clearer in the tests by specifying which demands we are satisfying in the form of test cases.

dlfivefifty:
Since it's a total order you should also overload `isless`

test/manyvariables.jl
lbenet: The test checks the *lexicographic* part in the ordering for `HomogeneousPolynomial`s, here comparing same orders; other tests (below) do something alike for different orders.

test/manyvariables.jl
lbenet: I agree with the expectation about the symmetry in labeling the variables; I expected that too... 

As I understand it, for many variables you have to impose something like that (which is the *lexicographic* part in the ordering), to have a *total* ordering (either `xT > yT` or `xT < yT`). Whatever lexicographic rule works fine. I've added some comments in the docstrings, hopefully making this clearer.

test/onevariable.jl
lbenet: Done!

test/onevariable.jl
lbenet: I've reverted the ordering in one of those tests (use `<` or `<=` instead of `>` or `>=`). this is the suggestion, right?

src/arithmetic.jl
lbenet: I've checked it, and it shows that there are `unbound_args` and `stale_deps` (`IntervalArithmetic`, which is rather a conditional dependency). So no method ambiguities! Thanks for the suggestion!

src/arithmetic.jl
lbenet: Somehow, the other one was unnecesary. My guess is that `>` relies on `<`, or the other way around... In any case, now, through metaprogramming, both methods are defined.

src/arithmetic.jl
lbenet: I added the comment `## Total ordering ##`.

src/arithmetic.jl
lbenet: Fixed!

src/arithmetic.jl
lbenet: Good catch! Fixed!

src/arithmetic.jl
lbenet: Done! Indeed, saves a lots of repetitions.

lbenet:
> Hope you don't mind the review, not that used to giving reviews to people I don't know, but I've done my best to be thorough. Great work in finding a reference in literature to work from. I'd suggest making this clearer in the tests by specifying which demands we are satisfying in the form of test cases.

Thanks a lot for the review @KeithWM , it was indeed useful !!

There are still few tests to add, in particular `findlast` (which I think I had a reason to add, but forgot why...), and also comparisons for the mixed cases (`Taylor1{TaylorN{T}}`, etc). I hope to have it totally ready in a couple of days.

lbenet:
> Since it's a total order you should also overload `isless`

Thanks for the suggestion @dlfivefifty, and I agree with the observation about the total order. Do you mean to have `isless` instead of `<`, or both (one relying on the other)?

dlfivefifty:
Overloading `isless` gives you all of them:
```julia
julia> struct Foo
       x
       end

julia> Base.isless(a::Foo, b::Foo) = isless(a.x, b.x)

julia> Foo(5) < Foo(6)
true

julia> Foo(6) ≥ Foo(4)
true

```

lbenet:
Finally, I got coverage back, and slightly increased :-)

I think this is ready to be merged... Any other comments?

src/arithmetic.jl
```diff
@@ -96,6 +96,33 @@ end
     end
 end
 
+# Mixtures
+@inline isless(a::Taylor1{TaylorN{T}}, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries} =
+    isless(a - b, zero(T))
+@inline function isless(a::HomogeneousPolynomial{Taylor1{T}}, b::HomogeneousPolynomial{Taylor1{T}}) where {T<:NumberNotSeries}
```
KeithWM, line 102: What is the reason for including the `@inline` macro?

KeithWM:
> Finally, I got coverage back, and slightly increased :-)
> 
> I think this is ready to be merged... Any other comments?

I see an 0.4% decrease in coverage according to Coveralls. I think that means we're still missing something part of this change.

src/arithmetic.jl
KeithWM: Also the `NumberNotSeries` seems very anti-Julian in that it prohibits the use of different number types (e.g. `Dual` or `Unitful`). I would rather see deriving from `<:Number` and explicity writing the unimplemented method for `{T<:Series}` to raise an error.

docs/src/userguide.md
```diff
@@ -336,6 +346,12 @@ Note that some of the arithmetic operations have been extended for
 [`HomogeneousPolynomial`](@ref); division, for instance, is not extended.
 The same convention used for `Taylor1` objects is used when combining
 `TaylorN` polynomials of different order.
+Both `HomogeneousPolynomial` and `TaylorN` are equiped with a total *lexicographical*
+order, provided by overloading [`isless`](@ref).
+The ordering is consistent with the interpretation that there are infinitessimal
+elements in the algebra; for details see M. Berz, "Automatic Differentiation as
+Nonarchimedean Analysis", Computer Arithmetic and Enclosure Methods, (1992), Elsevier,
+439-450.
```
KeithWM, line 354: I like the example provided for `Taylor1`. Can the same be done here?

test/mixtures.jl
```diff
@@ -22,6 +23,10 @@ using LinearAlgebra
     @test TS.numtype(tN) == TaylorN{Float64}
     @test normalize_taylor(tN) == tN
     @test tN.order == 3
+    @test HomogeneousPolynomial([1]) > xH > yH > 0.0
```
KeithWM, line 26: I think a (nested) `@testset` would be useful to indicate that these lines intend to show compliance with the reference by Berz would be useful.

test/mixtures.jl
```diff
@@ -340,12 +356,15 @@ end
     @test string(to^2) == " ( 1.0 + 𝒪(t⁴)) t² + 𝒪(t¹⁰)"
     @test ti + to == Taylor1([ti, one(ti)], 9)
     tito = ti * to
+    # @test ti > ti^2 > to > 0
```
KeithWM, line 359: I understand (very well) that you want to keep the question of these orderings out of scope of this PR. Can you create a separate issue for them and link to that issue in the comments? (Both in the tests and in the code.) Commented code is a bit of a smell, but I do agree that it would be a shame to delete this altogether.

test/onevariable.jl
```diff
@@ -138,14 +141,14 @@ Base.iszero(::SymbNumber) = false
...
     @test TS.numtype(TaylorSeries.fixorder(zt,Taylor1([1]))[1]) == Int
-    @test TaylorSeries.findfirst(t) == 1
-    @test TaylorSeries.findfirst(t^2) == 2
-    @test TaylorSeries.findfirst(ot) == 0
-    @test TaylorSeries.findfirst(zt) == -1
-    @test TaylorSeries.findlast(t) == 1
-    @test TaylorSeries.findlast(t^2) == 2
-    @test TaylorSeries.findlast(ot) == 0
-    @test TaylorSeries.findlast(zt) == -1
+    @test findfirst(t) == 1
```
KeithWM, line 144: Here too a (nested) `@testset` would be nice!

KeithWM (changes requested):
I think there's some room for improvement still. Also, it would be nice to see where the code coverage (for this change) is lacking.

src/arithmetic.jl
lbenet: This is indeed a fair criticism, yet beyond this PR. I opened #325 to remember it...

docs/src/userguide.md
lbenet: I just pushed a commit addressing this point. Let me know what you think

test/mixtures.jl
lbenet: Thanks for the suggestion; I opened #326, and include it in the comments

test/mixtures.jl
lbenet: I agree. But the whole tests are quite messy, so let me postpone this for another PR...

test/onevariable.jl
lbenet: Same comment as above... tests are overall messy...

lbenet:
> I think there's some room for improvement still. Also, it would be nice to see where the code coverage (for this change) is lacking.

Thanks for insisting 👍 

I think I've followed your recommendations, except regarding the (nested) `@testset`s comments. The tests are quite disorganized, so it is better to work them out consistently and thoroughly...

test/mixtures.jl
KeithWM: I know the rest is messy, but let's lot make a Herculean task of ever cleaning them byt not making a small start here!

KeithWM:
I'd still say at least grouping the new tests now would be helpful to any future effort to arrange them all, but I will approve this PR. (As if I have any power....)

docs/src/userguide.md
```diff
@@ -352,12 +352,23 @@ The ordering is consistent with the interpretation that there are infinitessimal
...
 439-450.
+Essentially, the lexicographic order works as follows: smaller order monomials
+are *larger* than higher order monomials; when the order is the same, *larger* monomials
+appear before in the hash-tables; the function [`show_monomials`](@ref).
+```@repl userguide
+x, y = set_variables("x y", order=10);
+
+show_monomials(2)
+```
+Then, the following clearly holds:
```
KeithWM, line 363: Maybe the "clearly" is pushing it a bit, but this is quite helpful.

KeithWM (approved):

KeithWM (approved):

dlfivefifty (approved):

docs/src/userguide.md
lbenet: Done!

lbenet:
Thanks for your reviews. I've corrected some further points, and bump a minor version. I'll merge when tests are ready.

PR #327, opened 2023-05-07
Have IntervalArithmetic as weakdeps (Julia v1.9)

lbenet:


lbenet:
Since Julia v1.9 is out there, I merge this...

PR #329, opened 2023-06-11
Add aqua tests

### LINKED ISSUE #328: Add Aqua tests
I recently precompiled a package that indirectly depends on TaylorSeries.jl, and saw these warnings:

```
julia> import ClimaCore
[ Info: Precompiling ClimaCore [d414da3d-4745-48bb-8d80-42e94e092884]
┌ Warning: AES-NI is not enabled, so AESNI and ARS are not available.
└ @ Random123 ~/.julia/packages/Random123/u5oEp/src/Random123.jl:55
WARNING: method definition for getindex at /Users/charliekawczynski/.julia/packages/TaylorSeries/YHybm/src/auxiliary.jl:125 declares type variable T but does not use it.
WARNING: method definition for getindex at /Users/charliekawczynski/.julia/packages/TaylorSeries/YHybm/src/auxiliary.jl:161 declares type variable T but does not use it.
```

This is perhaps an older version of TaylorSeries, but I think we should add Aqua.jl tests to prevent this from happening in the future.

charleskawczynski:
Closes #328.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -21,10 +21,11 @@ Requires = "0.5.2, 1"
 julia = "1"
 
 [extras]
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["IntervalArithmetic", "LinearAlgebra", "SparseArrays", "Test"]
+test = ["IntervalArithmetic", "LinearAlgebra", "SparseArrays", "Test", "Aqua"]
--- a/Project.toml
+++ b/Project.toml
@@ -9,16 +9,13 @@ Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
-[weakdeps]
-IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
-
-[extensions]
-TaylorSeriesIAExt = "IntervalArithmetic"
-
 [compat]
 IntervalArithmetic = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20"
 Requires = "0.5.2, 1"
 julia = "1"
+
+[extensions]
+TaylorSeriesIAExt = "IntervalArithmetic"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
@@ -29,3 +26,6 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
 test = ["IntervalArithmetic", "LinearAlgebra", "SparseArrays", "Test", "Aqua"]
+
+[weakdeps]
+IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
-version = "0.15.0"
+version = "0.15.1"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/test/aqua.jl
+++ b/test/aqua.jl
@@ -15,17 +15,15 @@ using Aqua
     pkg_match(pkgname, pkdir::Nothing) = false
     pkg_match(pkgname, pkdir::AbstractString) = occursin(pkgname, pkdir)
     filter!(x -> pkg_match("TaylorSeries", pkgdir(last(x).module)), ambs)
-    @test length(ambs) == 0
+    if VERSION < v"1.10.0-DEV"
+        @test length(ambs) == 0
+    end
 end
 
 @testset "Aqua tests (additional)" begin
     Aqua.test_undefined_exports(TaylorSeries)
     Aqua.test_deps_compat(TaylorSeries)
     Aqua.test_stale_deps(TaylorSeries; ignore=[:Requires])
-    if !isdefined(Base, :get_extension)
-        Aqua.test_project_extras(TaylorSeries) # failing on julia 1.9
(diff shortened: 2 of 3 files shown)


Project.toml
```diff
@@ -21,10 +21,11 @@ Requires = "0.5.2, 1"
...
 [extras]
+Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["IntervalArithmetic", "LinearAlgebra", "SparseArrays", "Test"]
+test = ["IntervalArithmetic", "LinearAlgebra", "SparseArrays", "Test", "aqua"]
```
lbenet, line 31: I think `"aqua"` here refers to the module, so it should be `"Aqua"`. See also [this line](https://github.com/JuliaDiff/TaylorSeries.jl/actions/runs/5236879611/jobs/9454701672?pr=329#step:5:24) from the tests.

lbenet (commented):
Thanks a lot for opening #328 and proposing a solution! We do indeed use Aqua *locally*, typically as `Aqua.test_all(TaylorSeries; unbound_args=false)`, but this PR certainly makes its use more consistent.

I left one comment but, since we are at this, there seem to be issues detected by Aqua related to `unbound_args` as noted in #328, but also in `project_extras` and `stale_deps`. Could you also address them?

Project.toml
charleskawczynski: Yes, thank you!

charleskawczynski:
> Thanks a lot for opening #328 and proposing a solution! We do indeed use Aqua _locally_, typically as `Aqua.test_all(TaylorSeries; unbound_args=false)`, but this PR certainly makes its use more consistent.
> 
> I left one comment but, since we are at this, there seem to be issues detected by Aqua related to `unbound_args` as noted in #328, but also in `project_extras` and `stale_deps`. Could you also address them?

Yes, I can take a look. I didn’t realize because I ran with Julia 1.8.5, and that version is passing all the tests. For those tests, should we `using Requires` unconditionally? Otherwise we could count the number of stale deps and restrict it for version past 1.9.

lbenet:
> Yes, I can take a look. 

Thanks for investing your time on this!

Regarding the failing tests, I think Aqua imposes some *stronger* formatting to the `Project.toml` file, which is the reason that it is *changed* according to the test; see [lines 130 through 159 here](https://github.com/JuliaDiff/TaylorSeries.jl/actions/runs/5237214837/jobs/9455247410?pr=329#step:6:133). (In #327, I edited `Project.toml` by hand following the manual; I didn't know there was a stronger format.)

>  I didn’t realize because I ran with Julia 1.8.5, and that version is passing all the tests. For those tests, should we using `Requires` unconditionally? Otherwise we could count the number of stale deps and restrict it for version past 1.9.

Regarding this, we should certainly have a test (since everything works fine on Julia 1.8.5), which depends on the Julia version. For Julia 1.9, either we skip it, mark it as broken, or have it commented; I actually do not quite see what stale dependencies we have... If it is related to `IntervalArithmetic`, which is conditionally loaded, I would suggest not to have is as an unconditionally dependence; while it is not too heavy, not too many people use it.

charleskawczynski:
I haven't looked at the `test_project_extras` or `test_project_toml_formatting` failures yet. I'm actually unfamiliar with those, maybe I can comment them for now and someone else can take those on?

> I actually do not quite see what stale dependencies we have... If it is related to `IntervalArithmetic`, which is conditionally loaded, I would suggest not to have is as an unconditionally dependence; while it is not too heavy, not too many people use it.

The stale deps in julia 1.9 are because 
```julia
if !isdefined(Base, :get_extension) # returns true for julia 1.8 and false in julia 1.9
    using Requires
end
```
So, we do not actually use `Requires` in julia 1.9. I think the simples solution (to continue keeping the package working with previous julia versions) is to update the test accordingly. I'll make some adjustments.

lbenet:
The following way of testing stale deps in Julia 1.9 pass: `Aqua.test_stale_deps(TaylorSeries; ignore=[:Requires])`

charleskawczynski:
> The following way of testing stale deps in Julia 1.9 pass: `Aqua.test_stale_deps(TaylorSeries; ignore=[:Requires])`

Updated!

lbenet:
There are two remaining problems in the tests:
- For Julia 1.6.7, the problem is the format of `Project.toml`. As I understand it, the `[extensions]` section should be after the `[compat]` section, and the `[weakdeps]` after `[targets]` (at the very end of the file).
- For the Julia nightly version, if I understand [this line](https://github.com/JuliaDiff/TaylorSeries.jl/actions/runs/5238100385/jobs/9456679883?pr=329#step:6:136) which is proposed in your tests, there is some ambiguity being detected. Perhaps we should skip that test for Julia 1.10.0-DEV versions, until we figure out what is the problem.

charleskawczynski:
Ok, I'll try to take a look tomorrow

lbenet:
Thanks a lot!

test/aqua.jl
```diff
@@ -0,0 +1,30 @@
...
+
+    # See: https://github.com/SciML/OrdinaryDiffEq.jl/issues/1750
+    # Test that we're not introducing method ambiguities across deps
+    ambs = Aqua.detect_ambiguities(TaylorSeries; recursive = true)
+    pkg_match(pkgname, pkdir::Nothing) = false
+    pkg_match(pkgname, pkdir::AbstractString) = occursin(pkgname, pkdir)
+    filter!(x -> pkg_match("TaylorSeries", pkgdir(last(x).module)), ambs)
+    if VERSION ≠ v"1.10.0-DEV"
+        @test length(ambs) == 0
+    end
```
lbenet, line 20: Julia nightly is still failing. I think the problem with this `if` block is that it still returns `true`:
```julia
julia> v"1.10.0-DEV.1471" ≠ v"1.10.0-DEV"
true
```
so that test is still evaluated in Julia nightly. For the time being, I think we should replace it by `if VERSION < v"1.10.0-DEV"`.

Hopefully it works...

lbenet:
> Ok, it seems silly that cross- julia version canonical forms are not the same (maybe this is an issue with Aqua.jl ?). I propose nixing the
> 
> ```julia
>         Aqua.test_project_extras(TaylorSeries) # failing on julia 1.9
>         Aqua.test_project_toml_formatting(TaylorSeries) # failing on julia 1.9
> ```
> 
> tests altogether. Frankly, aside from automating the processing of those toml files, I see no real benefit. Also, AFAIK, this won't impact TaylorSeries.jl users.

Regarding different Julia versions, see e.g. https://github.com/JuliaTesting/Aqua.jl/issues/137; at least type-piracy tests do behave differently.

I left one comment regarding the broken test for nightly; I think that change should yield green lights. Aside from that, I agree with you, that the other tests can be left aside.

lbenet:
Tests passed!

charleskawczynski:
Sure!

lbenet:
Once the green lights are there, I'll merge this!

Thanks a lot!

Issue #339, opened 2023-11-02
Invalidation


lrnv:
hey, 

using [this tutorial](https://timholy.github.io/SnoopCompile.jl/dev/tutorial/) on my particular workflow, I endeed up with a lot of invalidations in TaylorSeries.jl as follows: 

````julia
julia> trees[end]
inserting convert(::Type{Array{TaylorSeries.Taylor1{TaylorSeries.TaylorN{T}}, N}}, s::Array{TaylorSeries.TaylorN{TaylorSeries.Taylor1{T}}, N}) where {T<:Union{Real, Complex}, N} @ TaylorSeries C:\Users\lrnv\.julia\packages\TaylorSeries\OLHVb\src\conversion.jl:140 invalidated:
   backedges: 1: superseding convert(::Type{T}, a::AbstractArray) where T<:Array @ Base array.jl:613 with MethodInstance for convert(::Type{Vector{_A}} where _A, ::Vector) (1 children)
              2: superseding convert(::Type{T}, a::AbstractArray) where T<:Array @ Base array.jl:613 with MethodInstance for convert(::Type{<:Vector}, ::Vector) (263 children)

julia> 
````
Looks like the problematic method is the convert method there : https://github.com/JuliaDiff/TaylorSeries.jl/blob/ff1f6806dcd65ac6af9e417bcfd71ce483abfa3e/src/conversion.jl#L140-L148 which invalidates a bunch of convertion already defined in base. 

The second one is a promote rule, also from `TaylorSeries.jl`: 

````julia
julia> trees[end-1]
inserting promote_rule(::Type{S}, ::Type{T}) where {S<:Union{Real, Complex}, T<:AbstractSeries} @ TaylorSeries C:\Users\lrnv\.julia\packages\TaylorSeries\OLHVb\src\conversion.jl:201 invalidated:
   backedges:  1: superseding promote_rule(::Type, ::Type) @ Base promotion.jl:319 with MethodInstance for promote_rule(::Type{UInt64}, ::Type{<:Integer}) (1 children)
               2: superseding promote_rule(::Type, ::Type) @ Base promotion.jl:319 with MethodInstance for promote_rule(::Type{UInt8}, ::Type{<:Integer}) (1 children)       
               3: superseding promote_rule(::Type, ::Type) @ Base promotion.jl:319 with MethodInstance for promote_rule(::Type{UInt16}, ::Type{<:Integer}) (1 children)      
[...]

PR #341, opened 2023-11-24
Add compat for Aqua and stdlibs

Sbozzolo:
TaylorSeries.jl did not specify a compat for Aqua. Aqua 0.8 introduces some breaking changes, such as renaming `test_piracy` to `test_piracies`, leading to broken test if one has a recent version of Aqua in their environment.

I added a compat for Aqua. New versions of Aqua also check for compats for standard libraries (which will be a requirement to publish the package), so I added those as well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -10,8 +10,13 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
+Aqua = "0.8"
 IntervalArithmetic = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20"
+LinearAlgebra = "<0.0.1, 1"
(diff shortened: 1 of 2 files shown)


lbenet:
Hi @Sbozzolo, thanks a lot for opening this PR and for the contribution, and sorry for the time to respond to it. Your PR is very very timely. In relation to #339 I've been checking Aqua, so I will leave a few suggestions to add which simply exploit further Aqua. I hope this afternoon I'll send you these suggestions.

lbenet:
Tests pass, so I'm merging.

Thanks a lot @Sbozzolo !!

PerezHz:
> (and we should probably tag a new (patch?) version.

I'd be personally in favor of bumping the patch version 😄

lbenet:
Me too... hold me on just a couple of days; I'm addressing #339, and noticed that some ambiguities may occur. (If this takes too long, I'll simply tag the new patch version.)

PR #343, opened 2023-11-29
Fix bug in Taylor1 constructor for mixtures and add tests

PerezHz:
Fixes #342. Essentially, `fill` was used in a `Taylor1` constructor around here
https://github.com/JuliaDiff/TaylorSeries.jl/blob/5895948554c7161bf93aaa9f59543d8bbbbae236/src/constructors.jl#L51
which [according to the docs](https://docs.julialang.org/en/v1/base/arrays/#Base.fill) works fine when the underlying type is immutable, but causes issues for mutable types, such as `Array`.

Issue #342, opened 2023-11-29
Bug in Taylor1 constructor for mixtures


PerezHz:
When modifying in-place a single coefficient of a `Taylor1{TaylorN{Float64}}`, the value is copied almost everywhere:
```julia
julia> using TaylorSeries

julia> set_variables("x", numvars=2, order=6)
2-element Vector{TaylorN{Float64}}:
  1.0 x₁ + 𝒪(‖x‖⁷)
  1.0 x₂ + 𝒪(‖x‖⁷)

julia> z0N = -1.333+get_variables()[1]
 - 1.333 + 1.0 x₁ + 𝒪(‖x‖⁷)

julia> z = Taylor1(z0N,5)
  - 1.333 + 1.0 x₁ + 𝒪(‖x‖⁷) + 𝒪(t⁶)

julia> z[5][1][1] = 5.0
5.0

julia> z
  - 1.333 + 1.0 x₁ + 𝒪(‖x‖⁷) + ( 5.0 x₁ + 𝒪(‖x‖⁷)) t + ( 5.0 x₁ + 𝒪(‖x‖⁷)) t² + ( 5.0 x₁ + 𝒪(‖x‖⁷)) t³ + ( 5.0 x₁ + 𝒪(‖x‖⁷)) t⁴ + ( 5.0 x₁ + 𝒪(‖x‖⁷)) t⁵ + 𝒪(t⁶)
```

The expected output is simply:
```julia
julia> z
  - 1.333 + 1.0 x₁ + 𝒪(‖x‖⁷) + ( 5.0 x₁ + 𝒪(‖x‖⁷)) t⁵ + 𝒪(t⁶)
```

Update: while writing this issue I found the culprit and will submit a PR soon!

lbenet:
Grat catch, and thanks a lot for solving the problem in #343. Later this afternoon I'll review it.

PR #346, opened 2023-12-20
Small fix in Taylor1 constructor?

PerezHz:
This PR is a bit related to #342 and #343 and represents a solution to the following (unexpected?) behavior: if we have a `Taylor1{TaylorN{Float64}}` constructed from another variable of the same type, modifying in-place the former modifies also the latter:

```julia
julia> using TaylorSeries

julia> set_variables("x", numvars=2, order=2)
  1.0 x₁ + 𝒪(‖x‖³)
  1.0 x₂ + 𝒪(‖x‖³)

julia> a = sum(exp.(get_variables()).^2)
 2.0 + 2.0 x₁ + 2.0 x₂ + 2.0 x₁² + 2.0 x₂² + 𝒪(‖x‖³)

julia> b = Taylor1([a])
  2.0 + 2.0 x₁ + 2.0 x₂ + 2.0 x₁² + 2.0 x₂² + 𝒪(‖x‖³) + 𝒪(t¹)

julia> bcopy = deepcopy(b)
  2.0 + 2.0 x₁ + 2.0 x₂ + 2.0 x₁² + 2.0 x₂² + 𝒪(‖x‖³) + 𝒪(t¹)

julia> c = Taylor1(constant_term(b),0)
  2.0 + 2.0 x₁ + 2.0 x₂ + 2.0 x₁² + 2.0 x₂² + 𝒪(‖x‖³) + 𝒪(t¹)

julia> c[0][0][1] = 0.0
0.0

julia> b == bcopy # I would expect this to be true, since only c was modified
false
```

What are your thoughts over this @lbenet?

### DIFF
--- a/src/constructors.jl
+++ b/src/constructors.jl
@@ -49,7 +49,7 @@ Taylor1(coeffs::Array{T,1}, order::Int) where {T<:Number} = Taylor1{T}(coeffs, o
 Taylor1(coeffs::Array{T,1}) where {T<:Number} = Taylor1(coeffs, length(coeffs)-1)
 function Taylor1(x::T, order::Int) where {T<:Number}
     v = [zero(x) for _ in 1:order+1]
-    v[1] = x
+    v[1] = deepcopy(x)
     return Taylor1(v, order)
 end
 
--- a/test/mixtures.jl
+++ b/test/mixtures.jl
@@ -367,6 +367,13 @@ using Test
     for i in 2:19
         @test iszero(intz[i])
     end
+
+    a = sum(exp.(get_variables()).^2)
+    b = Taylor1([a])
+    bcopy = deepcopy(b)
+    c = Taylor1(constant_term(b),1)
+    c[0][0][1] = 0.0
+    b == bcopy
 end
 
 @testset "Tests with nested Taylor1s" begin
--- a/test/mixtures.jl
+++ b/test/mixtures.jl
@@ -371,7 +371,7 @@ using Test
     a = sum(exp.(get_variables()).^2)
     b = Taylor1([a])
     bcopy = deepcopy(b)
-    c = Taylor1(constant_term(b),1)
+    c = Taylor1(constant_term(b),0)
     c[0][0][1] = 0.0
     b == bcopy
 end


lbenet:
Good catch @PerezHz, and sorry for the delay to get into it... 

In general, I like the simplicity of the proposed solution, and agree to merge this. Before that, let me ask some questions: Do you know if there is a similar example/problem for `Taylor1{T}` for `T<:NumberNotSeries`, and what is the impact that `deepcopy` has in such cases? In other words, can we avoid the `deepcopy` (in some cases, if it impacts performace/allocations), and only use it for the mixed structs? Do we also need something similar for `TaylorN{Taylor1{T}}`?

lbenet:
Just to have it written somewhere, I think this issue is due to the `view`s: induced e.g. by [these lines](https://github.com/JuliaDiff/TaylorSeries.jl/blob/9437c3f2409558554ee137f544bd309800315d76/src/auxiliary.jl#L80C1-L85C1).

PerezHz:
Indeed, on current `main`, there's similar behavior for `Taylor1{T} where T<:NumberNotSeries` and `TaylorN{Taylor1{T}}`:

```julia
julia> using TaylorSeries
[ Info: Precompiling TaylorSeries [6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea]

julia> x = TaylorN(Taylor1(5),0)^C

julia> set_variables("x", numvars=2, order=2)
  1.0 x₁ + 𝒪(‖x‖³)
  1.0 x₂ + 𝒪(‖x‖³)

julia> x = TaylorN(Taylor1(5),0)
 ( 1.0 t + 𝒪(t⁶)) + 𝒪(‖x‖¹)

julia> y = x
 ( 1.0 t + 𝒪(t⁶)) + 𝒪(‖x‖¹)

julia> y[0][1][0] = 7.1
7.1

julia> x
 ( 7.1 + 1.0 t + 𝒪(t⁶)) + 𝒪(‖x‖¹)

julia> x = Taylor1(rand(5))
 0.6049707581312165 + 0.175608853487092 t + 0.16818194777538198 t² + 0.6251617825278187 t³ + 0.8179378046391547 t⁴ + 𝒪(t⁵)

julia> y = x
 0.6049707581312165 + 0.175608853487092 t + 0.16818194777538198 t² + 0.6251617825278187 t³ + 0.8179378046391547 t⁴ + 𝒪(t⁵)

julia> y[0] = 1.0
1.0

julia> x
 1.0 + 0.175608853487092 t + 0.16818194777538198 t² + 0.6251617825278187 t³ + 0.8179378046391547 t⁴ + 𝒪(t⁵)
```

And if I remove the `views` from the `getindex` methods, then the problem seems to persist...

PerezHz:
> and what is the impact that deepcopy has in such cases? In other words, can we avoid the deepcopy (in some cases, if it impacts performace/allocations), and only use it for the mixed structs?

I agree that in general it's best to avoid allocations as much as possible, but If I think about how we use `Taylor1` in `TaylorIntegration`, there we first allocate all the variables we need, and then start modifying their coefficients in-place throughout an integration. In that case, I think it actually makes sense to have the `deepcopy` at construction, since we are constructing a bunch of `Taylor1`s at the allocation stage which are in principle fully independent from each other (memory-wise); otherwise, we might end up modifying coefficients that were not meant to be updated. That said, if deemed a better solution, I can add a specialized constructor for mixed structs here which uses `deepcopy`, while leaving the constructor for `Taylor1{T} where T<:NumberNotSeries` as it is now (i.e., without `deepcopy`).

lbenet:
> I think it actually makes sense to have the deepcopy at construction, since we are constructing a bunch of Taylor1s at the allocation stage which are in principle fully independent from each other (memory-wise).

I fully agree with you. Let us go then that way. Can you also fix the `TaylorN{Taylor1{T}}` case, and bump a patch version?

lbenet:
Merging this; the patch will come eventually...

Thanks a lot @PerezHz

PerezHz:
@lbenet thanks for merging and sorry for delay with the patch, I'll try to come back around this in the next few days!

lbenet:
Nevermind... we can always bump the version later!

PR #347, opened 2024-01-31
Avoid some allocations in mutating methods

PerezHz:
In current `master`, some mutating methods allocate, especially for `TaylorN` and mixtures:

```julia
julia> using TaylorSeries, BenchmarkTools

julia> dq = set_variables("dq", order=2, numvars=2)
  1.0 dq₁ + 𝒪(‖x‖³)
  1.0 dq₂ + 𝒪(‖x‖³)

julia> x1N = Taylor1([dq[1],dq[2],dq[1]*dq[2]])
  1.0 dq₁ + 𝒪(‖x‖³) + ( 1.0 dq₂ + 𝒪(‖x‖³)) t + ( 1.0 dq₁ dq₂ + 𝒪(‖x‖³)) t² + 𝒪(t³)

julia> y1N = atan(x1N)
  1.0 dq₁ + 𝒪(‖x‖³) + ( 1.0 dq₂ + 𝒪(‖x‖³)) t + ( 1.0 dq₁ dq₂ + 𝒪(‖x‖³)) t² + 𝒪(t³)

julia> z1N = zero(x1N)
  0.0 + 𝒪(‖x‖³) + 𝒪(t³)

julia> @btime TaylorSeries.mul!($z1N,$x1N,$y1N,0)
  590.044 ns (14 allocations: 1.06 KiB)
```

This PR is a proof of concept which avoids allocations in key places, such as `zero` and `mul!` (similar things can be tried in `div!`, etc.), just by defining suitable inplace `zero!` methods:

```julia
julia> using TaylorSeries, BenchmarkTools

julia> dq = set_variables("dq", order=2, numvars=2)
  1.0 dq₁ + 𝒪(‖x‖³)
  1.0 dq₂ + 𝒪(‖x‖³)

julia> x1N = Taylor1([dq[1],dq[2],dq[1]*dq[2]])
  1.0 dq₁ + 𝒪(‖x‖³) + ( 1.0 dq₂ + 𝒪(‖x‖³)) t + ( 1.0 dq₁ dq₂ + 𝒪(‖x‖³)) t² + 𝒪(t³)

julia> y1N = atan(x1N)
  1.0 dq₁ + 𝒪(‖x‖³) + ( 1.0 dq₂ + 𝒪(‖x‖³)) t + ( 1.0 dq₁ dq₂ + 𝒪(‖x‖³)) t² + 𝒪(t³)

julia> z1N = zero(x1N)
  0.0 + 𝒪(‖x‖³) + 𝒪(t³)

julia> @btime TaylorSeries.mul!($z1N,$x1N,$y1N,0)
  83.679 ns (0 allocations: 0 bytes)
```

Using these `zero!` methods then we can do things like:

```julia

julia> x1 = Taylor1(rand(3))
 0.6727385330826134 + 0.023575091119572567 t + 0.408205561614307 t² + 𝒪(t³)

julia> xN = sin(dq[1]+dq[2])
 1.0 dq₁ + 1.0 dq₂ + 𝒪(‖x‖³)

julia> @btime TaylorSeries.zero!($x1)
  8.049 ns (0 allocations: 0 bytes)

julia> @btime TaylorSeries.zero!($xN)
  21.815 ns (0 allocations: 0 bytes)

julia> @btime TaylorSeries.zero!($x1N)
  63.053 ns (0 allocations: 0 bytes)
```
without a single allocation.

This is still very much WIP, but do you think this is an idea worth pursuing @lbenet?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
 version = "0.16.0"
 
 [deps]
-InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
-version = "0.16.0"
+version = "0.17.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -17,7 +17,7 @@ TaylorSeriesIAExt = "IntervalArithmetic"
 
 [compat]
 Aqua = "0.8"
-IntervalArithmetic = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20"
+IntervalArithmetic = "0.15 - 0.20"
 LinearAlgebra = "<0.0.1, 1"
 Markdown = "<0.0.1, 1"
 Requires = "0.5.2, 1"
--- a/Project.toml
+++ b/Project.toml
@@ -4,10 +4,17 @@ repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
 version = "0.16.0"
 
 [deps]
+InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+
+[weakdeps]
+IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
+
+[extensions]
+TaylorSeriesIAExt = "IntervalArithmetic"
 
 [compat]
 Aqua = "0.8"
@@ -19,9 +26,6 @@ SparseArrays = "<0.0.1, 1"
 Test = "<0.0.1, 1"
 julia = "1"
 
-[extensions]
-TaylorSeriesIAExt = "IntervalArithmetic"
-
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
@@ -31,6 +35,3 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
 test = ["IntervalArithmetic", "LinearAlgebra", "SparseArrays", "Test", "Aqua"]
-
-[weakdeps]
-IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
--- a/Project.toml
+++ b/Project.toml
@@ -4,7 +4,6 @@ repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
 version = "0.16.0"
 
 [deps]
-InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
@@ -18,7 +17,6 @@ TaylorSeriesIAExt = "IntervalArithmetic"
 
 [compat]
 Aqua = "0.8"
-InteractiveUtils = "1"
 IntervalArithmetic = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20"
 LinearAlgebra = "<0.0.1, 1"
 Markdown = "<0.0.1, 1"
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
 version = "0.16.0"
 
 [deps]
+InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
@@ -17,6 +18,7 @@ TaylorSeriesIAExt = "IntervalArithmetic"
 
 [compat]
 Aqua = "0.8"
+InteractiveUtils = "1"
 IntervalArithmetic = "0.15, 0.16, 0.17, 0.18, 0.19, 0.20"
 LinearAlgebra = "<0.0.1, 1"
 Markdown = "<0.0.1, 1"
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -796,7 +796,6 @@ end
         c[k] += c[i] * b[k-i]
     end
     @inbounds c[k] = -c[k]/b[0]
-    # @inbounds divsubst!(c[k], b[0])
     return nothing
 end
 
@@ -851,31 +850,6 @@ function mul!(c::TaylorN, a::TaylorN, b::TaylorN)
         mul!(c, a, b, k)
     end
 end
-
-# # inplace method for computing c = -c/b
-# # NOTE: Here `divsubst!` *accumulates* the result inplace
-# @inline function divsubst!(c::TaylorN, b::TaylorN, k::Int)
-#     if k==0
-#         @inbounds for i in eachindex(c[0])
-#             c[0][i] = -c[0][i] / constant_term(b)
-#         end
-#         return nothing
-#     end
-
-#     @inbounds for i = 0:k-1
-#         mul!(c[k], c[i], b[k-i])
-#     end
-#     @inbounds for i in eachindex(c[k])
-#         c[k][i] = (-2c[k][i]) / constant_term(b)
-#     end
-#     return nothing
-# end
-
-# function divsubst!(c::TaylorN, b::TaylorN)
-#     for k in eachindex(c)
-#         divsubst!(c, b, k)
-#     end
-# end
 
 # NOTE: Here `div!` *accumulates* the result of a / b in res[k] (k > 0)
 @inline function div!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -758,7 +758,7 @@ end
         @inbounds res[0] = cdivfact
         return nothing
     end
-    zero!(res, a, ordT)
+    zero!(res[ordT])
     imin = max(0, ordT+ordfact-b.order)
     aux = TaylorN(zero(a[ordT][0][1]), a[ordT].order)
     for k in imin:ordT-1
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -665,17 +665,32 @@ function /(a::Taylor1{TaylorN{T}}, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSe
     return res
 end
 
-function /(a::S, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries, S<:Union{T,TaylorN{T}}}
-    R = promote_type(S, TaylorN{T})
-    iszero(a) && !iszero(b) && return convert(Taylor1{R}, zero(b))
+function /(a::S, b::Taylor1{TaylorN{T}}) where {S<:NumberNotSeries, T<:NumberNotSeries}
+    R = promote_type(TaylorN{S}, TaylorN{T})
+    res = convert(Taylor1{R}, zero(b))
+    iszero(a) && !iszero(b) && return res
 
     # order and coefficient of first factorized term
     # In this case, since a[k]=0 for k>0, we can simplify to:
     # ordfact, cdivfact = 0, a/b[0]
 
-    res = Taylor1(a/b[0], b.order)
     for ordT in eachindex(res)
         div!(res, a, b, ordT)
+    end
+    return res
+end
+
+function /(a::TaylorN{T}, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries}
+    res = zero(b)
+    iszero(a) && !iszero(b) && return res
+
+    # order and coefficient of first factorized term
+    # In this case, since a[k]=0 for k>0, we can simplify to:
+    # ordfact, cdivfact = 0, a/b[0]
+
+    aa = Taylor1(a, b.order)
+    for ordT in eachindex(res)
+        div!(res, aa, b, ordT)
     end
     return res
 end
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -798,8 +798,8 @@ mul!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
 
 @inline function div!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         b::NumberNotSeries, k::Int) where {T<:NumberNotSeries}
-    for l in eachindex(b[k])
-        for m in eachindex(b[k][l])
+    for l in eachindex(a[k])
+        for m in eachindex(a[k][l])
             res[k][l][m] = a[k][l][m]/b
         end
     end
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -665,6 +665,19 @@ function /(a::Taylor1{TaylorN{T}}, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSe
     return res
 end
 
+function /(a::NumberNotSeries, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries}
+    iszero(a) && !iszero(b) && return zero(a)
+
+    # order and coefficient of first factorized term
+    # In this case, since a[k]=0 for k>0, we can simplify to:
+    # ordfact, cdivfact = 0, a/b[0]
+
+    res = Taylor1(a/b[0], b.order)
+    for ordT in eachindex(res)
+        div!(res, a, b, ordT)
+    end
+    return res
+end
 
 @inline function divfactorization(a1::Taylor1, b1::Taylor1)
     # order of first factorized term; a1 and b1 assumed to be of the same order
@@ -733,6 +746,7 @@ end
 
 @inline function div!(c::Taylor1, a::NumberNotSeries,
         b::Taylor1, k::Int)
+    iszero(a) && !iszero(b) && zero!(c, k)
     # order and coefficient of first factorized term
     # In this case, since a[k]=0 for k>0, we can simplify to:
     # ordfact, cdivfact = 0, a/b[0]
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -732,7 +732,7 @@ end
 end
 
 div!(v::Taylor1, b::NumberNotSeries, a::Taylor1, k::Int) =
-    div!(v::Taylor1, Taylor1(b, a.order), a, k)
+    div!(v::Taylor1, b*one(a), a, k)
 
 # NOTE: Here `div!` *accumulates* the result of a / b in c[k] (k > 0)
 @inline function div!(c::TaylorN, a::TaylorN, b::TaylorN, k::Int)
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -665,8 +665,9 @@ function /(a::Taylor1{TaylorN{T}}, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSe
     return res
 end
 
-function /(a::NumberNotSeries, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries}
-    iszero(a) && !iszero(b) && return zero(a)
+function /(a::S, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries, S<:Union{T,TaylorN{T}}}
+    R = promote_type(S, TaylorN{T})
+    iszero(a) && !iszero(b) && return convert(Taylor1{R}, zero(b))
 
     # order and coefficient of first factorized term
     # In this case, since a[k]=0 for k>0, we can simplify to:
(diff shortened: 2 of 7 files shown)


lbenet:
I think it is worth pursuing indeed, since the results are really good! 

Once said that, let me point to [this line](https://github.com/JuliaDiff/TaylorSeries.jl/blob/f5587a772e99d93428c5201f5422d5629fa4fcaf/src/functions.jl#L244), which defines `zero!(res, a, ord)`, and supposedly avoids allocations; I'm not sure why we are incurring into allocating there...

lbenet:
We should indeed keep the methods, e.g. for initialization of the aux quantities created by `@taylorize` in TaylorIntegration!

lbenet:
( I guess a better place for those functions would be in `src/functions.jl`...) The idea is truly nice! Thanks a lot!

src/functions.jl
```diff
@@ -655,6 +655,30 @@ end
...
+
+@inline function zero!(a::HomogeneousPolynomial{T}) where {T<:NumberNotSeries}
+    a.coeffs .= zero.(a.coeffs)
+    return nothing
+end
+
+@inline function zero!(a::TaylorN{T}) where {T<:NumberNotSeries}
+    for i in 0:a.order
+        zero!(a[i])
+    end
```
lbenet, line 671: You can write this as
```julia
for i in eachindex(a)
    zero!(a[i])
end
```

PerezHz:
I'm trying to also avoid allocations in `TaylorSeries.div!`, and around this line I found there's an auxiliary `TaylorN` being declared there, which allocates once for each order of expansion:
https://github.com/JuliaDiff/TaylorSeries.jl/blob/f5587a772e99d93428c5201f5422d5629fa4fcaf/src/arithmetic.jl#L763
I'm thinking that a possible workaround is to add it as an argument to `div!`, so that that auxiliary can be declared somewhere else, but that change would be perhaps a bit breaking? There are other mutating functions with similar auxiliaries defined in the function body, which could benefit from a similar treatment, what are your thoughts on this @lbenet?

PerezHz:
In other words, whereas this PR makes `mul!` allocation-free for mixtures, `div!` is not:
```julia
julia> @btime TaylorSeries.mul!($z1N,$x1N,$y1N,0)
  55.880 ns (0 allocations: 0 bytes)

julia> @btime TaylorSeries.div!($z1N,$x1N,$y1N,0)
  360.201 ns (15 allocations: 1.09 KiB)
```

lbenet:
> I'm trying to also avoid allocations in `TaylorSeries.div!`, and around this line I found there's an auxiliary `TaylorN` being declared there, which allocates once for each order of expansion:
> 
> https://github.com/JuliaDiff/TaylorSeries.jl/blob/f5587a772e99d93428c5201f5422d5629fa4fcaf/src/arithmetic.jl#L763
> 

I don't quite remember the actual reason why that `aux` is needed there, but I think it is related to the sum in the recurrence relation... I recall the dirty tricks needed when defining sums and using `@taylorize`...

> I'm thinking that a possible workaround is to add it as an argument to `div!`, so that that auxiliary can be declared somewhere else, but that change would be perhaps a bit breaking? There are other mutating functions with similar auxiliaries defined in the function body, which could benefit from a similar treatment, what are your thoughts on this @lbenet?

I don't think we should care about how breaking that change would be to the code; we simply release a new patch version 😄. Once said this, my guess is that such a change impacts in the infrastructure needed for `@taylorize` to work nicely. I suggest adding a couple of comments (on improving `div!`) and leave those changes to another PR. My guess is that such auxiliaries occur beyond `div!`, so solving `div!` will tell us the way to proceed.

lbenet:
@PerezHz I've checked that `aux` (in `div!`, and actually other functions!) is indeed an auxiliary `::TaylorN`. I've been also thinking about exploiting the kind of tricks we use in the mutating functions, but it is not too clear to me that we can actually exploit them in TaylorIntegration. My point is that `@taylorize` declares all those quantities as appropriate `Taylor1`s, but in this case `aux` is a `TaylorN`...

PerezHz:
> I don't quite remember the actual reason why that aux is needed there, but I think it is related to the sum in the recurrence relation... I recall the dirty tricks needed when defining sums and using @taylorize...

Even with the dirty tricks the specialized methods for mixtures allocate a bit less than the generic ones, so it's definitely a step forward I would say!

>  I don't think we should care about how breaking that change would be to the code; we simply release a new patch version 😄. 

Agreed!

> Once said this, my guess is that such a change impacts in the infrastructure needed for @taylorize to work nicely.

A bit yes, but not so much, as far as I've managed to test! (more details on this below...)

>  suggest adding a couple of comments (on improving div!) and leave those changes to another PR.

So far the mutating methods have allowed very good speedups while using `@taylorize` and yeah, some, or even most, of these mutating methods (`pow!` for example) allocate `Taylor1`s and/or `TaylorN`s explicitly, which kind of "defeats" the purpose of the mutating methods, which is avoiding allocations. My point is, there is still performance potential to gain. I will add these comments and we can try to tackle them in another PR.

> @PerezHz I've checked that aux (in div!, and actually other functions!) is indeed an auxiliary ::TaylorN. I've been also thinking about exploiting the kind of tricks we use in the mutating functions, but it is not too clear to me that we can actually exploit them in TaylorIntegration. My point is that @taylorize declares all those quantities as appropriate Taylor1s, but in this case aux is a TaylorN...

That's a good point, and finding a good solution can take more than one attempt, yet I think it's worth pursuing. In a nutshell I guess what I'm trying to say is, we should aim for making `TaylorIntegration.jetcoeffs!` allocation-free.

Now, I found an issue with `pow!` for which I found a workaround but still haven't managed to find the reason behind it. Using this PR the following happens:

```julia
julia> using TaylorSeries

julia> dq = set_variables("δq", order=1, numvars=2)
  1.0 δq₁ + 𝒪(‖x‖²)
  1.0 δq₂ + 𝒪(‖x‖²)

julia> x = Taylor1([1.0 - 1.8dq[1],-0.007 + 0.0194dq[1], -1e-5 -0.00027dq[2], 3e-7 - 1.8e-5dq[1]-dq[2],dq[1]])
  1.0 - 1.8 δq₁ + 𝒪(‖x‖²) + ( - 0.007 + 0.0194 δq₁ + 𝒪(‖x‖²)) t + ( - 1.0e-5 - 0.00027 δq₂ + 𝒪(‖x‖²)) t² + ( 3.0e-7 - 1.8e-5 δq₁ - 1.0 δq₂ + 𝒪(‖x‖²)) t³ + ( 1.0 δq₁ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)

julia> y = Taylor1([-2.0 + dq[1],-1dq[2],2dq[2], dq[1]+dq[2],dq[2]])
  - 2.0 + 1.0 δq₁ + 𝒪(‖x‖²) + ( - 1.0 δq₂ + 𝒪(‖x‖²)) t + ( 2.0 δq₂ + 𝒪(‖x‖²)) t² + ( 1.0 δq₁ + 1.0 δq₂ + 𝒪(‖x‖²)) t³ + ( 1.0 δq₂ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)

julia> z = deepcopy(x)
  1.0 - 1.8 δq₁ + 𝒪(‖x‖²) + ( - 0.007 + 0.0194 δq₁ + 𝒪(‖x‖²)) t + ( - 1.0e-5 - 0.00027 δq₂ + 𝒪(‖x‖²)) t² + ( 3.0e-7 - 1.8e-5 δq₁ - 1.0 δq₂ + 𝒪(‖x‖²)) t³ + ( 1.0 δq₁ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)

julia> z.coeffs[1] = x.coeffs[1]^1.5 # this line and the one below is similar to how variables are initialized in parsed jetcoeffs! methods in TaylorIntegration
 1.0 - 2.7 δq₁ + 𝒪(‖x‖²)

julia> z.coeffs[2:end] .= zero(z.coeffs[1]) # similar to what happens in parsed jetcoeffs! methods
4-element view(::Vector{TaylorN{Float64}}, 2:5) with eltype TaylorN{Float64}:
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)

julia> TaylorSeries.pow!(z, x, 1.5, 1) # compute 1st-order coefficients

julia> z # the 1st-order coefficients are copied into higher-order coefficients!!
  1.0 - 2.7 δq₁ + 𝒪(‖x‖²) + ( - 0.0105 + 0.03855000000000001 δq₁ + 𝒪(‖x‖²)) t + ( - 0.0105 + 0.03855000000000001 δq₁ + 𝒪(‖x‖²)) t² + ( - 0.0105 + 0.03855000000000001 δq₁ + 𝒪(‖x‖²)) t³ + ( - 0.0105 + 0.03855000000000001 δq₁ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)
```

 any hints about what might be happening here?

PerezHz:
A workaround is to simply broadcast `zero`:
```julia
z.coeffs[2:end] .= zero.(z.coeffs[1]) # <-- broadcasted zero
```

lbenet:
I can't reproduce the problem you report [here](https://github.com/JuliaDiff/TaylorSeries.jl/pull/347#issuecomment-1934890667)... I'm using Julia v1.10, and TaylorSeries from master...

PerezHz:
> I can't reproduce the problem you report https://github.com/JuliaDiff/TaylorSeries.jl/pull/347#issuecomment-1934890667... I'm using Julia v1.10, and TaylorSeries from master...

Sorry for the confusion, the issue happens with this PR's branch

PerezHz:
> I can't reproduce the problem you report https://github.com/JuliaDiff/TaylorSeries.jl/pull/347#issuecomment-1934890667... I'm using Julia v1.10, and TaylorSeries from master...

... now, if you meant that the issue pops up only using this PR then yes, although I actually would say the issue is hidden in TaylorIntegration. I'll explain in the next comment.

PerezHz:
Using current TS master:

```julia
julia> using TaylorSeries

julia> dq = set_variables("δq", order=1, numvars=2)
  1.0 δq₁ + 𝒪(‖x‖²)
  1.0 δq₂ + 𝒪(‖x‖²)

julia> x = Taylor1([1.0+dq[1],-dq[1], dq[2], dq[1]-dq[2],dq[1]])
  1.0 + 1.0 δq₁ + 𝒪(‖x‖²) + ( - 1.0 δq₁ + 𝒪(‖x‖²)) t + ( 1.0 δq₂ + 𝒪(‖x‖²)) t² + ( 1.0 δq₁ - 1.0 δq₂ + 𝒪(‖x‖²)) t³ + ( 1.0 δq₁ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)

julia> z = deepcopy(x)
  1.0 + 1.0 δq₁ + 𝒪(‖x‖²) + ( - 1.0 δq₁ + 𝒪(‖x‖²)) t + ( 1.0 δq₂ + 𝒪(‖x‖²)) t² + ( 1.0 δq₁ - 1.0 δq₂ + 𝒪(‖x‖²)) t³ + ( 1.0 δq₁ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)

julia> z.coeffs[1] = x.coeffs[1]^1.5
 1.0 + 1.5 δq₁ + 𝒪(‖x‖²)

julia> z.coeffs[2:end] .= zero(z.coeffs[1]) # no zero broadcasting here
4-element view(::Vector{TaylorN{Float64}}, 2:5) with eltype TaylorN{Float64}:
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)

julia> z[1][0] = HomogeneousPolynomial(-4.0,0) # only one z coefficient is updated
 - 4.0

julia> z # but the change is copied everywhere!
  1.0 + 1.5 δq₁ + 𝒪(‖x‖²) + ( - 4.0 + 𝒪(‖x‖²)) t + ( - 4.0 + 𝒪(‖x‖²)) t² + ( - 4.0 + 𝒪(‖x‖²)) t³ + ( - 4.0 + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)
```
but this can be easily solved if we broadcast zero while initially filling the coefficients of z:
```julia
julia> z.coeffs[2:end] .= zero.(z.coeffs[1]) # zero is broadcasted here
4-element view(::Vector{TaylorN{Float64}}, 2:5) with eltype TaylorN{Float64}:
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)

julia> z[1][0] = HomogeneousPolynomial(-4.0,0)
 - 4.0

julia> z
  1.0 + 1.5 δq₁ + 𝒪(‖x‖²) + ( - 4.0 + 𝒪(‖x‖²)) t + 𝒪(t⁵)
```
AFAIU, the issue is that if `zero` is not broadcasted then each of the elements of `z.coeffs[2:end]` point to the same place in memory:
```julia
julia> z.coeffs[2:end] .= zero(z.coeffs[1]) # no zero broadcast
4-element view(::Vector{TaylorN{Float64}}, 2:5) with eltype TaylorN{Float64}:
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)

julia> z.coeffs[2] == z.coeffs[3]
true

julia> z.coeffs[2] === z.coeffs[3] # check if memory address is the same
true

julia> z.coeffs[2:end] .= zero.(z.coeffs[1]) # now, broadcast zero
4-element view(::Vector{TaylorN{Float64}}, 2:5) with eltype TaylorN{Float64}:
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)
  0.0 + 𝒪(‖x‖²)

julia> z.coeffs[2] == z.coeffs[3]
true

julia> z.coeffs[2] === z.coeffs[3] # now each coefficient points to a different place in memory
false
```

PerezHz:
Now, this raises the question: why does this appear in this PR and not in current `master`? The line that makes the trick is this:
https://github.com/JuliaDiff/TaylorSeries.jl/blob/f5587a772e99d93428c5201f5422d5629fa4fcaf/src/power.jl#L279
the way it currently behaves it allocates a new place in memory of each of the elements of `res.coeffs`, but these lines are precisely the ones that this PR touches, updating coefficients inplace while avoiding allocations. Of course, as you mentioned earlier, there are other allocations that are more tricky to remove, but we can go step by step.

If downstream tests pass (TI, PE, NEOs) then I would propose to go ahead with this and make the update in TaylorIntegration, unless there are other considerations to be made?

lbenet:
I think that `z.coeffs[2:end] .= zero(z.coeffs[1])` is (in this PR) somehow behaving as a view, while before (e.g. master) it was behaving as a copy; hence the allocations. Did you check that the workaround, i.e., apply broadcasting to `zero` too, does allocate?

lbenet:
Just as a check of what I say, consider (in your example):
```julia
julia> z = deepcopy(x)
  1.0 + 1.0 δq₁ + 𝒪(‖x‖²) + ( - 1.0 δq₁ + 𝒪(‖x‖²)) t + ( 1.0 δq₂ + 𝒪(‖x‖²)) t² + ( 1.0 δq₁ - 1.0 δq₂ + 𝒪(‖x‖²)) t³ + ( 1.0 δq₁ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)

julia> z.coeffs[3:end] .= zero(z.coeffs[1]); # changing from the 2nd order onwards

julia> z[3][0] = HomogeneousPolynomial(6.0,0) # changing order 3!
 6.0

julia> z # order 2 is also changed, but not order 1
  1.0 + 1.0 δq₁ + 𝒪(‖x‖²) + ( - 1.0 δq₁ + 𝒪(‖x‖²)) t + ( 6.0 + 𝒪(‖x‖²)) t² + ( 6.0 + 𝒪(‖x‖²)) t³ + ( 6.0 + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)
```
So indeed that seems to change all coefficients from order 2. I think this is because `z[z.coeffs[3:end]` is being treated as a view.

PerezHz:
> I think that z.coeffs[2:end] .= zero(z.coeffs[1]) is (in this PR) somehow behaving as a view, while before (e.g. master) it was behaving as a copy; hence the allocations. Did you check that the workaround, i.e., apply broadcasting to zero too, does allocate?

That's a good point, it does allocate, and even more: I would expect this since in the zero-broadcasting case it is allocating different places in memory for each z coefficient:

```julia
julia> using BenchmarkTools

julia> function zeronobc(z)
           z.coeffs[2:end] .= zero(z.coeffs[1])
           return nothing
       end
zeronobc (generic function with 1 method)

julia> function zerobc(z)
           z.coeffs[2:end] .= zero.(z.coeffs[1])
           return nothing
       end
zerobc (generic function with 1 method)

julia> @btime zeronobc($z)
  219.727 ns (10 allocations: 752 bytes)

julia> @btime zerobc($z)
  880.123 ns (40 allocations: 2.94 KiB)
```

lbenet:
How would using `zerobc` would compare with what we have in master? I mean, [this kind](https://github.com/JuliaDiff/TaylorSeries.jl/pull/347#issue-2111087411) of benchmark you did.

lbenet:
Using the last commit, i.e. [0187b2d](https://github.com/JuliaDiff/TaylorSeries.jl/pull/347/commits/0187b2d5ef6f0e83f472296efc98dbf8351f7967), breaks the tests of TI...

lbenet:
(Also, there are deprecation warnings related to the common interface)

PerezHz:
> How would using zerobc would compare with what we have in master? I mean, https://github.com/JuliaDiff/TaylorSeries.jl/pull/347#issue-2111087411 of benchmark you did.

Well I get similar results as with this PR:

```julia
julia> @btime zeronobc($z)
  217.885 ns (10 allocations: 752 bytes)

julia> @btime zerobc($z)
  878.714 ns (40 allocations: 2.94 KiB)
```

PerezHz:
> Using the last commit, i.e. [0187b2d](https://github.com/JuliaDiff/TaylorSeries.jl/pull/347/commits/0187b2d5ef6f0e83f472296efc98dbf8351f7967), breaks the tests of TI...

Thank you, yes, I was looking into that, the issues are resolved by changing the corresponding bits in `parse_eqs.jl` (i.e., using broadcasted zero to initialize arrays)

PerezHz:
> (Also, there are deprecation warnings related to the common interface)

Yes, we should fix those

PerezHz:
Just opened PerezHz/TaylorIntegration.jl#181 to test changes introduced by this PR

src/functions.jl
```diff
@@ -655,6 +655,75 @@ end
 
 
 # Mutating functions for Taylor1{TaylorN{T}}
+@inline function zero!(a::Taylor1{T}, k::Int) where {T<:NumberNotSeries}
```
lbenet, line 658: Sorry for this late suggestion: what about redefining `zero!(c, a, k)` simply to point to `zero!(c, k)`? That would avoid lots of changes (which can be kept here), but in particular those related to `@taylorize` (in [dictmutfunct.jl](https://github.com/JuliaDiff/TaylorSeries.jl/blob/master/src/dictmutfunct.jl).

src/arithmetic.jl
```diff
@@ -665,6 +665,19 @@ function /(a::Taylor1{TaylorN{T}}, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSe
...
+    # order and coefficient of first factorized term
+    # In this case, since a[k]=0 for k>0, we can simplify to:
+    # ordfact, cdivfact = 0, a/b[0]
+
+    res = Taylor1(a/b[0], b.order)
+    for ordT in eachindex(res)
+        div!(res, a, b, ordT)
+    end
+    return res
+end
```
lbenet, line 680: I propose the following here to ensure type stability (this is untested)
```julia
function /(a::NumberNotSeries, b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries}
    R = promote_type(typeof(a), TaylorN{T})
    iszero(a) && !iszero(b) && return convert(Taylor1{R}, zero(b))

    # order and coefficient of first factorized term
    # In this case, since a[k]=0 for k>0, we can simplify to:
    # ordfact, cdivfact = 0, a/b[0]

    res = Taylor1(a/b[0], b.order)
    for ordT in eachindex(res)
        div!(res, a, b, ordT)
    end
    return res
end
```

lbenet: Note that `a::NumberNotSeries` is perhaps too strong, because it leaves outside `a::TaylorN{T}`, which is a method worth having too.

src/arithmetic.jl
```diff
@@ -731,8 +744,25 @@ end
     return nothing
 end
 
-div!(v::Taylor1, b::NumberNotSeries, a::Taylor1, k::Int) =
-    div!(v::Taylor1, Taylor1(b, a.order), a, k)
+@inline function div!(c::Taylor1, a::NumberNotSeries,
+        b::Taylor1, k::Int)
```
lbenet, line 748: Perhaps we should impose `c::Taylor1{T}` as well as `b::Taylor1{T}`...

src/arithmetic.jl
```diff
@@ -731,8 +744,25 @@ end
...
+    iszero(a) && !iszero(b) && zero!(c, k)
+    # order and coefficient of first factorized term
+    # In this case, since a[k]=0 for k>0, we can simplify to:
+    # ordfact, cdivfact = 0, a/b[0]
+    if k == 0
+        @inbounds c[0] = a/b[0]
+        return nothing
+    end
+
+    imin = max(0, k-b.order)
```
lbenet, line 758: `imin` should be 0 always: because 0<= k <= b.order...

PerezHz: Good catch! It's now fixed 👍

src/arithmetic.jl
PerezHz: I agree, although I think we can make an extra tweak to avoid some allocations for `T<:TaylorN{<:NumberNotSeries}`, so I added an extra method for the latter case (will fix it in the next few commits)

src/arithmetic.jl
PerezHz: Thanks for the suggestion! I added the fix for type-stability and widened the type for `a` to allow for `TaylorN{T}`

src/functions.jl
PerezHz: Thanks for the suggestion! I changed the definition of `zero!(c, a, k)`

lbenet: One more suggestion on this: can you move the code (of the `zero!` methods introduced in this PR), closer to other `zero!` methods?

lbenet:
Tests in TI.jl are passing now. I'll try to complete my review in the next few days. Thanks a lot!

Project.toml
```diff
@@ -4,13 +4,21 @@ repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
 version = "0.16.0"
 
 [deps]
+InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
```
lbenet, line 7: Question: Why do we need InteractiveUtils? (I couldn't find where it was used...)

PerezHz: Thanks, it was a leftover from some local tests, it's now removed

PerezHz:
> Tests in TI.jl are passing now. I'll try to complete my review in the next few days. Thanks a lot!

Sure! There's still a couple of things I'd like to try here to save some further allocs, but indeed this is essentially ready for review.

PerezHz:
I think there's indeed many things that can still be done in order to avoid allocations in mutating functions, but the changes introduced here I think are in a sense self-contained, so I'd rather leave further changes to another PR. Therefore, as it is standing now, this PR is ready for review.

PerezHz:
(maybe I'm repeating myself but just wanted to be clear that modulo reviews and some fixes, I don't intend a priori to add more things to this PR)

PerezHz:
I've been running some benchmarks and using this PR for JT integrations in NEOs.jl we get 30% less memory used and about 20% better (shorter) runtime in long (>100 timesteps) integrations. Short integrations, where "time to first time-step" is dominant, will not benefit that much from this PR.

lbenet:
I'll have a look later; the results you just quoted are very encouraging, so we should think about merging and tagging a new version!

PerezHz:
> I'll have a look later; the results you just quoted are very encouraging, so we should think about merging and tagging a new version!

Alright! Just bumped the minor version, let me know if there is anything else you would like to improve around here.

lbenet (approved):
Thanks a lot @PerezHz, this is a nice improvement! I have left one single comment (with code), which is a suggestion that may improve performance. In any case, I'm approving this PR and merging it, so we can exploit it in TaylorIntegration!

Thanks again a lot!

src/arithmetic.jl
```diff
@@ -731,23 +773,94 @@ end
...
+    # In this case, since a[k]=0 for k>0, we can simplify to:
+    # ordfact, cdivfact = 0, a/b[0]
+    if k == 0
+        @inbounds div!(c[0], a, b[0])
+        return nothing
+    end
+
+    @inbounds mul!(c[k], c[0], b[k])
+    @inbounds for i = 1:k-1
+        c[k] += c[i] * b[k-i]
```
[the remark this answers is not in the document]
PerezHz: Thank you for the suggestion, I'll give it a try and see if we can get rid of some more allocations 😄 !

PR #349, opened 2024-02-25
Further reduction of allocs in mutating methods for mixtures

PerezHz:
During the development of #347, @lbenet suggested a way to further reduce allocations in mutating division methods for mixtures (see [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/347#discussion_r1501682516)). After running some local tests, it indeed looks like an option worthwhile pursuing.

On current master:
```julia
julia> using BenchmarkTools

julia> using TaylorSeries
Precompiling TaylorSeries
  1 dependency successfully precompiled in 3 seconds. 3 already precompiled.

julia> dq = set_variables("δq", order=1, numvars=2)
  1.0 δq₁ + 𝒪(‖x‖²)
  1.0 δq₂ + 𝒪(‖x‖²)

julia> x = Taylor1([2+2dq[1],-1.1dq[1], 0.7dq[2], 0.5dq[1]-0.45dq[2],0.9dq[1]])
  2.0 + 2.0 δq₁ + 𝒪(‖x‖²) + ( - 1.1 δq₁ + 𝒪(‖x‖²)) t + ( 0.7 δq₂ + 𝒪(‖x‖²)) t² + ( 0.5 δq₁ - 0.45 δq₂ + 𝒪(‖x‖²)) t³ + ( 0.9 δq₁ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)

julia> z = deepcopy(x)
  2.0 + 2.0 δq₁ + 𝒪(‖x‖²) + ( - 1.1 δq₁ + 𝒪(‖x‖²)) t + ( 0.7 δq₂ + 𝒪(‖x‖²)) t² + ( 0.5 δq₁ - 0.45 δq₂ + 𝒪(‖x‖²)) t³ + ( 0.9 δq₁ + 𝒪(‖x‖²)) t⁴ + 𝒪(t⁵)

julia> TaylorSeries.zero!(z)

julia> iszero(z)
true

julia> z[0] = 1/x[0]
 0.5 - 0.5 δq₁ + 𝒪(‖x‖²)

julia> @btime TaylorSeries.div!($z, 1, $x, 1)
  735.992 ns (18 allocations: 1.27 KiB)

julia> @btime TaylorSeries.div!($z, 1, $x, 2)
  1.342 μs (36 allocations: 2.53 KiB)

julia> @btime TaylorSeries.div!($z, 1, $x, 3)
  2.000 μs (54 allocations: 3.80 KiB)

julia> @btime TaylorSeries.div!($z, 1, $x, 4)
  2.657 μs (72 allocations: 5.06 KiB)
```
This PR:
```julia
julia> @btime TaylorSeries.div!($z, 1, $x, 1)
  685.610 ns (18 allocations: 1.27 KiB)

julia> @btime TaylorSeries.div!($z, 1, $x, 2)
  708.029 ns (18 allocations: 1.27 KiB)

julia> @btime TaylorSeries.div!($z, 1, $x, 3)
  724.354 ns (18 allocations: 1.27 KiB)

julia> @btime TaylorSeries.div!($z, 1, $x, 4)
  740.625 ns (18 allocations: 1.27 KiB)
```
~~The remaining 18 allocations are coming from this line:
https://github.com/JuliaDiff/TaylorSeries.jl/blob/bc1735c32546415c1c254a342537c28c361f988b/src/arithmetic.jl#L815~~
EDIT: the afore-mentioned 18 allocations are not there anymore.

UPDATE: after successfully reducing allocation in mutating division and product methods for mixtures, these changes enabled further allocation reduction in sqrt, pow, and other functions. These improvements have also been added to this PR.

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -251,7 +251,13 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
 
             ## add! and subst! ##
             function ($fc)(v::$T{T}, a::$T{T}, k::Int) where {T<:Number}
-                @inbounds v[k] = ($f)(a[k])
+                if $T == Taylor1
+                    @inbounds v[k] = ($f)(a[k])
+                else
+                    @inbounds for l in eachindex(v[k])
+                        v[k][l] = ($f)(a[k][l])
+                    end
+                end
                 return nothing
             end
 
@@ -376,6 +382,14 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
             @inbounds for i in eachindex(v[k])
                 for j in eachindex(v[k][i])
                     v[k][i][j] = ($f)(a[k][i][j], b[k][i][j])
+                end
+            end
+            return nothing
+        end
+        function ($fc)(v::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
+            @inbounds for l in eachindex(v[k])
+                for m in eachindex(v[k][l])
+                    v[k][l][m] = ($f)(a[k][l][m])
                 end
             end
             return nothing
@@ -505,7 +519,7 @@ for T in (:Taylor1, :TaylorN)
         if $T == Taylor1
             @inbounds v[k] = a[k] * b
         else
-            for i in eachindex(v[k])
+            @inbounds for i in eachindex(v[k])
                 v[k][i] = a[k][i] * b
             end
         end
@@ -515,7 +529,7 @@ for T in (:Taylor1, :TaylorN)
         if $T == Taylor1
             @inbounds v[k] = a * b[k]
         else
-            for i in eachindex(v[k])
+            @inbounds for i in eachindex(v[k])
                 v[k][i] = a * b[k][i]
             end
         end
@@ -872,7 +886,7 @@ end
 end
 
 # TODO: avoid allocations when T isa Taylor1
-@inline function div!(v::HomogeneousPolynomial{T}, a::HomogeneousPolynomial{T}, b::T) where {T <: Number}
+@inline function div!(v::HomogeneousPolynomial{T}, a::HomogeneousPolynomial{T}, b::NumberNotSeriesN) where {T <: Number}
     @inbounds for k in eachindex(v)
         v[k] = a[k] / b
     end
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -810,7 +810,10 @@ end
 
     @inbounds mul!(c[k], c[0], b[k])
     @inbounds for i = 1:k-1
-        c[k] += c[i] * b[k-i]
+        # c[k] += c[i] * b[k-i]
+        for ord in eachindex(c[k])
+            mul!(c[k], c[i], b[k-i], ord)
+        end
     end
     @inbounds c[k] = -c[k]/b[0]
     return nothing
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -813,7 +813,8 @@ end
         # c[k] += c[i] * b[k-i]
         mul!(c[k], c[i], b[k-i])
     end
-    @inbounds c[k] = -c[k]/b[0]
+    # @inbounds c[k] = -c[k]/b[0]
+    @inbounds divsubst!(c[k], b[0])
     return nothing
 end
 
@@ -829,6 +830,53 @@ end
     end
     @inbounds for i in eachindex(c[k])
         c[k][i] = (a[k][i] - c[k][i]) / constant_term(b)
+    end
+    return nothing
+end
+
+# in-place division and assignment: c = c / a
+# NOTE: Here `div!` *accumulates* the result of c / a in c[k] (k > 0)
+#
+# Recursion algorithm:
+#
+# k = 0: c[0] <-  c[0]/a[0]
+# k = 1: c[1] <- -c[1]
+#        c[1] <-  c[1] + c[0]*a[1]
+#        c[1] <- -c[1]/a[0]
+# k = 2: c[2] <- -c[2]
+#        c[2] <-  c[2] + c[0]*a[2] + c[1]*a[1]
+#        c[2] <- -c[2]/a[0]
+# etc.
+@inline function div!(c::TaylorN, a::TaylorN, k::Int)
+    if k==0
+        @inbounds c[0][1] = constant_term(c) / constant_term(a)
+        return nothing
+    end
+
+    @inbounds for i in eachindex(c[k])
+        c[k][i] = -c[k][i]
+    end
+    @inbounds for i = 0:k-1
+        mul!(c[k], c[i], a[k-i])
+    end
+    @inbounds for i in eachindex(c[k])
+        c[k][i] = (-c[k][i]) / constant_term(a)
+    end
+    return nothing
+end
+
+# In-place, allocation-free, computation of k-th order coefficient of c = - c / a
+@inline function divsubst!(c::TaylorN, a::TaylorN, k::Int)
+    if k==0
+        @inbounds c[0][1] = - constant_term(c) / constant_term(a)
+        return nothing
+    end
+
+    @inbounds for i = 0:k-1
+        mul!(c[k], c[i], a[k-i])
+    end
+    @inbounds for i in eachindex(c[k])
+        c[k][i] = (-c[k][i]) / constant_term(a)
     end
     return nothing
 end
@@ -850,6 +898,12 @@ end
 end
 
 # in-place division (assumes equal order among TaylorNs)
+function div!(c::TaylorN, a::TaylorN)
+    for k in eachindex(c)
+        div!(c, a, k)
+    end
+end
+
 function div!(c::TaylorN, a::TaylorN, b::TaylorN)
     for k in eachindex(c)
         div!(c, a, b, k)
@@ -859,6 +913,12 @@ end
 function div!(c::TaylorN, a::NumberNotSeries, b::TaylorN)
     for k in eachindex(c)
         div!(c, a, b, k)
+    end
+end
+
+function divsubst!(c::TaylorN, a::TaylorN)
+    for k in eachindex(c)
+        divsubst!(c, a, k)
     end
 end
 
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -979,8 +979,8 @@ end
     return nothing
 end
 
-# In-place division and assignment: c[k] = c[k] / a[k]
-# NOTE: Here `div!` *accumulates* the result of c[k] / a[k] in c[k] (k > 0)
+# In-place division and assignment: c[k] = (c/a)[k]
+# NOTE: Here `div!` *accumulates* the result of (c/a)[k] in c[k] (k > 0)
 #
 # Recursion algorithm:
 #
@@ -1005,8 +1005,8 @@ end
     return nothing
 end
 
-# In-place division and assignment: c[k] <- scalar * c[k] / a[k]
-# NOTE: Here `div!` *accumulates* the result of scalar * c[k] / a[k] in c[k] (k > 0)
+# In-place division and assignment: c[k] <- scalar * (c/a)[k]
+# NOTE: Here `div!` *accumulates* the result of scalar * (c/a)[k] in c[k] (k > 0)
 #
 # Recursion algorithm:
 #
@@ -1064,7 +1064,7 @@ function div_scalar!(c::TaylorN, scalar::NumberNotSeries, a::TaylorN)
     return nothing
 end
 
-# c[k] <- a[k]/b[k]
+# c[k] <- (a/b)[k]
 function div!(c::TaylorN, a::TaylorN, b::TaylorN)
     @inbounds for k in eachindex(c)
         div!(c, a, b, k)
@@ -1072,6 +1072,7 @@ function div!(c::TaylorN, a::TaylorN, b::TaylorN)
     return nothing
 end
 
+# c[k] <- (a/b)[k], where a is a scalar
 function div!(c::TaylorN, a::NumberNotSeries, b::TaylorN)
     @inbounds for k in eachindex(c)
         div!(c, a, b, k)
@@ -1079,6 +1080,7 @@ function div!(c::TaylorN, a::NumberNotSeries, b::TaylorN)
     return nothing
 end
 
+# c[k] <- a[k]/b, where b is a scalar
 function div!(c::TaylorN, a::TaylorN, b::NumberNotSeries)
     @inbounds for k in eachindex(c)
         div!(c[k], a[k], b)
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -529,10 +529,12 @@ for T in (:Taylor1, :TaylorN)
         else
             @inbounds mul_scalar!(c[k], scalar, a[0], b[k])
         end
-        @inbounds for i = 1:k
-            if $T == Taylor1
+        if $T == Taylor1
+            @inbounds for i = 1:k
                 c[k] += a[i] * b[k-i]
-            else
+            end
+        else
+            @inbounds for i = 1:k
                 mul_scalar!(c[k], scalar, a[i], b[k-i])
             end
         end
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -504,7 +504,7 @@ end
 mul!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
 b::NumberNotSeries, k::Int) where {T<:NumberNotSeries} = mul!(res, b, a, k)
 
-# in-place division (assumes equal order among TaylorNs)
+# in-place product (assumes equal order among TaylorNs)
 function mul!(c::TaylorN, a::TaylorN, b::TaylorN)
     for k in eachindex(c)
         mul!(c, a, b, k)
@@ -811,9 +811,7 @@ end
     @inbounds mul!(c[k], c[0], b[k])
     @inbounds for i = 1:k-1
         # c[k] += c[i] * b[k-i]
-        for ord in eachindex(c[k])
-            mul!(c[k], c[i], b[k-i], ord)
-        end
+        mul!(c[k], c[i], b[k-i])
     end
     @inbounds c[k] = -c[k]/b[0]
     return nothing
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -260,9 +260,22 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
                 return nothing
             end
 
-            function ($fc)(v::$T, a::$T, b::$T, k::Int)
-                @inbounds v[k] = ($f)(a[k], b[k])
-                return nothing
+            # function ($fc)(v::$T, a::$T, b::$T, k::Int)
+            #     @inbounds v[k] = ($f)(a[k], b[k])
+            #     return nothing
+            # end
+            if $T == Taylor1
+                function ($fc)(v::$T, a::$T, b::$T, k::Int)
+                    @inbounds v[k] = ($f)(a[k], b[k])
+                    return nothing
+                end
+            else
+                function ($fc)(v::$T, a::$T, b::$T, k::Int)
+                    @inbounds for i in eachindex(v[k])
+                        v[k][i] = ($f)(a[k][i], b[k][i])
+                    end
+                    return nothing
+                end
             end
 
             function ($fc)(v::$T, a::$T, b::Number, k::Int)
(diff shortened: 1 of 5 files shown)


src/arithmetic.jl
```diff
@@ -530,14 +592,14 @@ c_k = \sum_{j=0}^k a_j b_{k-j}.
...
 
-Accumulates in `c[k]` the result of `a*b` with minimum allocation; all arguments
-are `HomogeneousPolynomial`.
+Accumulates in `c` the result of `a*b*d` with minimum allocation. Arguments
+c, a and b are `HomogeneousPolynomial`; d is a NumberNotSeries whose default value is 1.
 
 """
 @inline function mul!(c::HomogeneousPolynomial, a::HomogeneousPolynomial,
-        b::HomogeneousPolynomial)
+        b::HomogeneousPolynomial, d::T=1) where {T<:NumberNotSeries}
```
lbenet, line 602: I do see the interest to have this method. Yet, I wonder if it does induce some sort of type unstability... My point is, since the default value is `1::Int`, if we set it to something else (say 1.0), do we use the same method or a different one? Perhaps it is better to have two methods, `mul!(c, a, b)` and `mul!(c, a, b, d::T)`, where the latter includes multiplying by the "scalar" `d::T<:NumberNotSeries`.

src/arithmetic.jl
```diff
@@ -823,6 +890,8 @@ end
         return nothing
     end
 
+    zero!(c, k)
+
```
lbenet, line 894: I think that `zero(c, k)` makes that `div!` does not accumulate anymore the result. So I suggest to correct the note. 

Another suggestion would be to have two methods `div_accumul!` and `div` (and `mul_accumul!` and `mul!`), and use them properly. I think this would simplify `div!` below a lot...

src/arithmetic.jl
```diff
@@ -832,6 +901,63 @@ end
...
+#
+# Recursion algorithm:
+#
+# k = 0: c[0] <-  c[0]/a[0]
+# k = 1: c[1] <- -c[1]
+#        c[1] <-  c[1] + c[0]*a[1]
+#        c[1] <- -c[1]/a[0]
+# k = 2: c[2] <- -c[2]
+#        c[2] <-  c[2] + c[0]*a[2] + c[1]*a[1]
+#        c[2] <- -c[2]/a[0]
```
lbenet, line 915: This resembles `fma`... maybe we should (in another PR) implement it

lbenet:
@PerezHz I have left three comments on what you have so far in this PR; not sure if they are good ideas or not. In any case, this looks very promising!

PerezHz:
> @PerezHz I have left three comments on what you have so far in this PR; not sure if they are good ideas or not. In any case, this looks very promising!

Many thanks @lbenet! This is still WIP but even so, many thanks for the comments you left; I'll try to address them in the next few days! I very much agree the changes here can be improved 😄

src/arithmetic.jl
PerezHz: I agree there's some potential for type-instability here; this way it just seemed really convenient to avoid duplicating code. I'll see if I can find another solution.

src/arithmetic.jl
PerezHz: > I think that zero(c, k) makes that div! does not accumulate anymore the result. So I suggest to correct the note.

Good catch! Okay, I'll fix it 👍

src/arithmetic.jl
PerezHz: I agree it's worth implementing!

PerezHz:
@lbenet I managed to get rid of almost all allocations in `pow!`, except for this line:
https://github.com/PerezHz/TaylorSeries.jl/blob/f95b797a00766396f46f378416129875dbe7528a/src/power.jl#L299
I'm thinking a possible solution could be to simply define a mutating method `power_by_squaring!` for `TaylorN`, waht do you think? Or maybe there's a better way?

lbenet:
Thanks again, truly, this is looking awesome!!

> @lbenet I managed to get rid of almost all allocations in `pow!`, except for this line: https://github.com/PerezHz/TaylorSeries.jl/blob/f95b797a00766396f46f378416129875dbe7528a/src/power.jl#L299 I'm thinking a possible solution could be to simply define a mutating method `power_by_squaring!` for `TaylorN`, what do you think? Or maybe there's a better way?

While I think it is possible to reduce the number of allocations with respect to the current implementation, the implementation (seems to me) not trivial, because of the recursive nature of the function and the fact that products (and squares) are computed order-by-order. I thus suggest you give it a try, with the possibility to do it in another PR.

PerezHz:
> While I think it is possible to reduce the number of allocations with respect to the current implementation, the implementation (seems to me) not trivial, because of the recursive nature of the function and the fact that products (and squares) are computed order-by-order. I thus suggest you give it a try, with the possibility to do it in another PR.

This is something I'm still looking at, but perhaps we can leave it as a to-do for the time being. In the end, one allocation per a full expansion is still way better than it was before. I'm still tackling a few remaining cases, but this should be ready for review in the next couple of days...

lbenet:
Fully agree...

PerezHz:
Actually, I think the changes here so far are in a sense self-contained and by themselves achieve good performance improvements. If you agree I'll leave further changes for another PR, so this is now ready for review 😄.

src/arithmetic.jl
PerezHz: Regarding this point, the current approach is to add `scalar::NumberNotSeries` as a keyword argument, which avoids dispatching on this argument. Adding a `mul!(a, b, c, d::T)` can call the wrong method of `mul!` when `d` is an integer. Another approach can also be to simply define `mul_scalar!(a, b, c, d::NumberNotSeries)` and call this method where appropriate. What are your thoughts on this?

lbenet: I favor your the proposal with `mul_scalar!`, simply because it will have less methods than `mul!`...

src/functions.jl
```diff
@@ -303,10 +303,35 @@ end
...
-        @inline function identity!(c::$T{T}, a::$T{T}, k::Int) where {T<:Number}
-            @inbounds c[k] = identity(a[k])
+        @inline function identity!(c::$T{T}, a::$T{T}, k::Int) where {T<:NumberNotSeries}
+            if $T == Taylor1
+                @inbounds c[k] = identity(a[k])
+            else
+                @inbounds for l in eachindex(c[k])
+                    identity!(c[k], a[k], l)
+                end
+            end
```
lbenet, line 328: This `if` is within a `for` loop, which I guess is bad for performance. I suggest to have it the other way around, though I admit that it implies duplicating the `for` loops in both `if` branches...

src/functions.jl
```diff
@@ -345,19 +370,18 @@ for T in (:Taylor1, :TaylorN)
...
-                    mul!(c[k], (k-i) * a[k-i], c[i])
+                    mul!(c[k], a[k-i], c[i], scalar=k-i)
                 end
             end
-            @inbounds c[k] = c[k] / k
+            if $T == Taylor1
+                @inbounds div!(c, c, k, k)
+            else
+                @inbounds div!(c[k], c[k], k)
+            end
```
lbenet, line 384: Same comment as above; in this case, I think the second `if`s can be included in the first.

src/functions.jl
```diff
@@ -371,16 +395,20 @@ for T in (:Taylor1, :TaylorN)
...
```
lbenet, line 411: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -393,16 +421,11 @@ for T in (:Taylor1, :TaylorN)
...
-            end
-            @inbounds for i = 2:k-1
+            @inbounds for i = 1:k-1
                 if $T == Taylor1
                     c[k] += (k-i) * a[i] * c[k-i]
                 else
-                    mul!(c[k], (k-i)*a[i], c[k-i])
+                    mul!(c[k], a[i], c[k-i], scalar=k-i)
                 end
             end
```
lbenet, line 430: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -423,16 +446,11 @@ for T in (:Taylor1, :TaylorN)
...
```
lbenet, line 455: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -442,70 +460,79 @@ for T in (:Taylor1, :TaylorN)
...
-            @inbounds c[k] = c[k] / k
+            if $T == Taylor1
+                s[k] = s[k] / k
+                c[k] = c[k] / k
+            else
+                @inbounds div!(s[k], s[k], k)
+                @inbounds div!(c[k], c[k], k)
+            end
             return nothing
         end
```
lbenet, line 491: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -442,70 +460,79 @@ for T in (:Taylor1, :TaylorN)
...
+                    mul!(c[k], a[k-i], c2[i], scalar=k-i)
                 end
             end
-            @inbounds c[k] = a[k] + c[k]/k
+            # c[k] <- c[k]/k
+            if $T == Taylor1
+                div!(c, c, k, k) # Taylor1
+            else
+                div!(c[k], c[k], k) # TaylorN
+            end
```
lbenet, line 533: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -514,25 +541,48 @@ for T in (:Taylor1, :TaylorN)
...
-            end
-            @inbounds for i in 2:k-1
+            @inbounds for i in 1:k-1
                 if $T == Taylor1
                     c[k] += (k-i) * r[i] * c[k-i]
                 else
-                    mul!(c[k], (k-i) * r[i], c[k-i])
+                    mul!(c[k], r[i], c[k-i], scalar=k-i)
+                end
+            end
```
lbenet, line 560: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -514,25 +541,48 @@ for T in (:Taylor1, :TaylorN)
...
+                subst!(s, s, i)
+                if i == 0
+                    if $T == Taylor1
+                        s[0] = one(s[0]) - s[0]
+                    else
+                        s[0][1] = one(s[0][1]) - s[0][1]
+                    end
+                    add!(s, one_s, s, 0)
+                end
+            end
```
lbenet, line 576: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -569,16 +614,11 @@ for T in (:Taylor1, :TaylorN)
                 return nothing
             end
```
lbenet, line 615: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -569,16 +614,11 @@ for T in (:Taylor1, :TaylorN)
...
-            end
-            @inbounds for i in 2:k-1
+            @inbounds for i in 1:k-1
                 if $T == Taylor1
                     c[k] += (k-i) * r[i] * c[k-i]
                 else
-                    mul!(c[k], (k-i) * r[i], c[k-i])
+                    mul!(c[k], r[i], c[k-i], scalar=k-i)
                 end
             end
```
lbenet, line 623: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -595,14 +635,7 @@ for T in (:Taylor1, :TaylorN)
...
             zero!(c, k)
-            if $T == Taylor1
-                @inbounds s[k] = x * c[k-1]
-                @inbounds c[k] = x * s[k-1]
-            else
-                @inbounds mul!(s[k], x, c[k-1])
-                @inbounds mul!(c[k], x, s[k-1])
-            end
-            @inbounds for i = 2:k
+            @inbounds for i = 1:k
```
lbenet, line 638: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -625,16 +658,11 @@ for T in (:Taylor1, :TaylorN)
...
-            end
-            @inbounds for i = 1:k-1
+            @inbounds for i = 0:k-1
                 if $T == Taylor1
                     c[k] += (k-i) * a[k-i] * c2[i]
                 else
-                    mul!(c[k], (k-i) * a[k-i], c2[i])
+                    mul!(c[k], a[k-i], c2[i], scalar=k-i)
                 end
             end
```
lbenet, line 667: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -651,16 +679,11 @@ for T in (:Taylor1, :TaylorN)
...
-            end
-            @inbounds for i in 2:k-1
+            @inbounds for i in 1:k-1
                 if $T == Taylor1
                     c[k] += (k-i) * r[i] * c[k-i]
                 else
-                    mul!(c[k], (k-i) * r[i], c[k-i])
+                    mul!(c[k], r[i], c[k-i], scalar=k-i)
                 end
             end
```
lbenet, line 688: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -676,16 +699,11 @@ for T in (:Taylor1, :TaylorN)
...
```
lbenet, line 708: Same comment about order of `if` and `for`

src/functions.jl
```diff
@@ -701,16 +719,11 @@ for T in (:Taylor1, :TaylorN)
...
```
lbenet, line 728: Same comment about order of `if` and `for`

src/arithmetic.jl
```diff
@@ -453,28 +507,53 @@ end
 # Internal multiplication functions
 for T in (:Taylor1, :TaylorN)
     # NOTE: For $T = TaylorN, `mul!` *accumulates* the result of a * b in c[k]
-    @eval @inline function mul!(c::$T{T}, a::$T{T}, b::$T{T}, k::Int) where {T<:Number}
+    @eval @inline function mul!(c::$T{T}, a::$T{T}, b::$T{T}, k::Int; scalar::NumberNotSeries=1) where {T<:Number}
```
lbenet, line 510: Eventhough I think `mul_scalar!` is better in this context (too many `mul!` methods, I find it ok to have the keyword argument.

lbenet (commented):
I've left a bunch of comments, mainly in src/functions.jl; they are about inverting (outer) `for` loops, and `if` blocks. Other than that, LGTM.

src/arithmetic.jl
PerezHz: Thanks for this comment, I agree that it's better not to "over-overload" `mul!` so actually in the end I did add `mul_scalar!` and also the equivalent `div_scalar!`.

src/functions.jl
PerezHz: Thanks for spotting this, I'm now changing the occurrences of this pattern, although this case in particular has the `for` within the `else` so this one is actually correct, right?

src/functions.jl
PerezHz: Done.

src/functions.jl
PerezHz: Done.

src/functions.jl
PerezHz: Done.

src/functions.jl
PerezHz: Done.

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: Done 👍

src/functions.jl
PerezHz: (actually the code changed here a bit so this `for` is not there anymore)

src/functions.jl
PerezHz: Done 👍

PerezHz:
> I've left a bunch of comments, mainly in src/functions.jl; they are about inverting (outer) for loops, and if blocks. Other than that, LGTM.

Thank you for spotting this! I've pushed some changes addressing your comments, I'm guessing this time we can bump a patch version, since only the internal API is changing? Tests on TaylorIntegration and NEOs pass with this branch without any changes 👍

src/arithmetic.jl
```diff
@@ -469,12 +523,53 @@ for T in (:Taylor1, :TaylorN)
...
+        else
+            @inbounds mul_scalar!(c[k], scalar, a[0], b[k])
+        end
+        @inbounds for i = 1:k
+            if $T == Taylor1
+                c[k] += a[i] * b[k-i]
+            else
+                mul_scalar!(c[k], scalar, a[i], b[k-i])
+            end
+        end
```
lbenet, line 538: Same comment wrt order of `for` and `if`

lbenet: This can be rewritten as
```julia
if $T == Taylor1
    @inbounds c[k] = a[0] * b[k]
    @inbounds for i = 1:k
        c[k] += a[i] * b[k-i]
     end
else
    @inbounds mul_scalar!(c[k], scalar, a[0], b[k])
    @inbounds for i = 1:k
        mul_scalar!(c[k], scalar, a[i], b[k-i])
    end
end
```

src/functions.jl
lbenet: Ups... sorry

src/functions.jl
```diff
@@ -442,70 +459,83 @@ for T in (:Taylor1, :TaylorN)
...
+                @inbounds for i = 0:k-1
+                    mul_scalar!(c[k], k-i, a[k-i], c2[i])
+                end
+            end
+            # c[k] <- c[k]/k
+            if $T == Taylor1
+                div!(c, c, k, k) # Taylor1
+            else
+                div!(c[k], c[k], k) # TaylorN
             end
```
lbenet, line 536: This `if` can be merged to the previous one

src/functions.jl
```diff
@@ -596,24 +653,24 @@ for T in (:Taylor1, :TaylorN)
...
+                    mul_scalar!(s[k], i, a[i], c[k-i])
+                    mul_scalar!(c[k], i, a[i], s[k-i])
+                end
+            end
+            if $T == Taylor1
+                @inbounds div!(s, s, k, k)
+                @inbounds div!(c, c, k, k)
+            else
+                @inbounds div!(s[k], s[k], k)
+                @inbounds div!(c[k], c[k], k)
```
lbenet, line 672: This if block can also be merged with the previous one

src/functions.jl
```diff
@@ -626,18 +683,21 @@ for T in (:Taylor1, :TaylorN)
...
-            @inbounds for i = 1:k-1
-                if $T == Taylor1
-                    c[k] += (k-i) * a[k-i] * c2[i]
-                else
-                    mul!(c[k], (k-i) * a[k-i], c2[i])
+            if $T == Taylor1
+                @inbounds c[k] = a[k] - c[k]/k
+            else
+                @inbounds for l in eachindex(c[k])
+                    c[k][l] = a[k][l] - c[k][l]/k
```
lbenet, line 698: Same comment: `if`s can be merged

lbenet:
> Thank you for spotting this! I've pushed some changes addressing your comments, I'm guessing this time we can bump a patch version, since only the internal API is changing? Tests on TaylorIntegration and NEOs pass with this branch without any changes 👍

I've left few more comments along the same lines as before. In any case, I do agree with you that we should bump a patch version before merging this PR. It's also very nice that tests pass smoothly in TaylorIntegration.jl and NEOs.jl; have you checked the performance?

PerezHz:
> have you checked the performance?

So far I have found ~40% improvements in memory usage and performance, sometimes even better!

lbenet:
> So far I have found ~40% improvements in memory usage and performance, sometimes even better!

Very impressive! Please, let me know when you think this is ready, and I'll go ahead and merge it, and bump the patch version.

PerezHz:
> Very impressive! Please, let me know when you think this is ready, and I'll go ahead and merge it, and bump the patch version.

Just rebased to current master, and unless CI complains, this is ready to be merged!

lbenet:
Thanks a lot again @PerezHz. I'll wait for the tests to complete, and then merge it!

lbenet:
Merging!

PR #350, opened 2024-03-06
use iszero in inverse

adamnemecek:


lbenet:
Thanks a lot! LGTM!

What about tagging a new patch version, and then merge it?

lbenet:
I'll go ahead and merge. The patch version release will be made later (probably when #349 is merged).

Thanks again!

PR #353, opened 2024-03-29
Add StaticArrays extension

PerezHz:
Currently, when doing some operations with SMatrix and SVector, and TaylorSeries defined types are involved, the type of the returned value is not inferred correctly (especially for mixtures), which then have to be converted back to the desired types, but we can solve these issues by defining a couple of `Base.promote_op` dispatches for `Taylor1`, `TaylorN` and mixes thereof.

For example, in the following code
```julia
using TaylorSeries, StaticArrays
q = set_variables("q", order=2, numvars=2)
m = @SMatrix fill(Taylor1(rand(2).*q), 3, 3)
mt = m'
v = @SVector [-1.1, 3.4, 7.62345e-1]
mtv = mt * v
```
the matrix `m` is a `SMatrix{3, 3, Taylor1{TaylorN{Float64}}, 9}`, while `mt` is a `SMatrix{3, 3, Any, 9}` and `mtv` is a `SVector{3, Any}`. This PR solves these issues.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -11,7 +11,8 @@ testfiles = (
     "intervals.jl",
     "broadcasting.jl",
     "identities_Euler.jl",
-    "fateman40.jl"
+    "fateman40.jl",
+    "staticarrays.jl"
     )
 
 for file in testfiles
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -3,7 +3,6 @@
 # Tests for TaylorSeries
 
 testfiles = (
-    "aqua.jl",
     "onevariable.jl",
     "manyvariables.jl",
     "mixtures.jl",
--- a/Project.toml
+++ b/Project.toml
@@ -11,9 +11,11 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [weakdeps]
 IntervalArithmetic = "d1acc4aa-44c8-5952-acd4-ba5d80a2a253"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [extensions]
 TaylorSeriesIAExt = "IntervalArithmetic"
+TaylorSeriesSAExt = "StaticArrays"
 
 [compat]
 Aqua = "0.8"
@@ -22,6 +24,7 @@ LinearAlgebra = "<0.0.1, 1"
 Markdown = "<0.0.1, 1"
 Requires = "0.5.2, 1"
 SparseArrays = "<0.0.1, 1"
+StaticArrays = "1"
 Test = "<0.0.1, 1"
 julia = "1"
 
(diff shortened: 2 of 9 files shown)


test/runtests.jl
```diff
@@ -11,7 +11,8 @@ testfiles = (
     "intervals.jl",
     "broadcasting.jl",
     "identities_Euler.jl",
-    "fateman40.jl"
+    "fateman40.jl",
+    "staticarrays.jl"
```
lbenet, line 15: Currently, Aqua tests are run twice; profitting this PR, could you remove the "aqua.jl" entry (in line 6 of this file)?

lbenet (commented):
Thanks @PerezHz for addressing this. 
I left one comment that suggests avoiding run twice the "aqua.jl" tests (the latter is the important one). Also, can you bump a patch version?
Aside from that, LGTM, and I think it is ready to be merged.

test/runtests.jl
PerezHz: It's done now 👍

PerezHz:
Bumped the patch version and removed the duplicated tests, so looks like this is now ready to be merged. Thanks @lbenet for taking a look!

PerezHz:
Pushed a commit which overloads `Base.float` for types defined in TaylorSeries, in a manner consistent with similar methods for other numeric types in Julia. It is not in principle in the scope of this PR, but took the opportunity to add them here (with corresponding tests), since it allows to avoid overloading some `SatelliteToolboxTransformations` methods in `NEOs`. Otherwise, from my side this is ready to be merged.

PerezHz:
See also related discussion in JuliaSpace/SatelliteToolboxTransformations.jl#6

lbenet:
> Pushed a commit which overloads `Base.float` for types defined in TaylorSeries, in a manner consistent with similar methods for other numeric types in Julia. It is not in principle in the scope of this PR, but took the opportunity to add them here (with corresponding tests), since it allows to avoid overloading some `SatelliteToolboxTransformations` methods in `NEOs`. Otherwise, from my side this is ready to be merged.

I agree that this PR did not have this overloading purpose 😄. Tests are passing, and everything seems to behave as expected, which is ok. Yet, before merging I think we should double-check that other packages behave as expected (TaylorIntegration and NEOs, specially); such an overload may have unexpected consequences elsewhere...

lbenet:
I checked TaylorIntegratio, TaylorModels, and NEOs, and all packages behave ok wrt overloading `float` here. So. I'l go ahead and merge this PR, and release the new version.

Thanks a lot @PerezHz !

PerezHz:
> Yet, before merging I think we should double-check that other packages behave as expected (TaylorIntegration and NEOs, specially); such an overload may have unexpected consequences elsewhere...

That is actually a very good point, many thanks for double-checking tests on dependant packages!

PR #354, opened 2024-03-31
Add JLD2 extension

PerezHz:
This PR adds functionality currently hosted in NEOs, but which should be moved over here for at least two reasons: avoiding type-piracy 🏴‍☠️ (see PerezHz/NEOs.jl#62), as well as enabling users in general to store TaylorN polynomials in `.jld2` files.

Acknowledgments: original `TaylorNSerialization` implementation is due to @LuEdRaMo.


PerezHz:
@lbenet just pushed a commit removing a leftover comment, but this ready to be merged 👍

lbenet (approved):
Thanks a lot! Merging

PR #355, opened 2024-04-10
Add TaylorN constructor

PerezHz:
This PR adds a `TaylorN` constructor which takes simply a `<:NumberNotSeries` variable and the current order from `get_order()`, which allows to do e.g.:
```julia
julia> a = TaylorN{Float64}(-1//3)
 -0.3333333333333333 + 𝒪(‖x‖⁷)

julia> typeof(a)
 TaylorN{Float64}
```

### DIFF
--- a/test/manyvariables.jl
+++ b/test/manyvariables.jl
@@ -6,6 +6,9 @@ using Test
 # using LinearAlgebra
 
 @testset "Test hash tables" begin
+    a = TaylorN{Float64}(9//10)
+    a isa TaylorN{Float64}
+    @test constant_term(a) == Float64(9//10)
     # Issue #85 is solved!
     set_variables("x", numvars=66, order=1)
     @test TS._params_TaylorN_.order == get_order() == 1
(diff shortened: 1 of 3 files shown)


test/manyvariables.jl
```diff
@@ -6,6 +6,9 @@ using Test
 # using LinearAlgebra
 
 @testset "Test hash tables" begin
+    a = TaylorN{Float64}(9//10)
+    a isa TaylorN{Float64}
+    @test constant_term(a) == Float64(9//10)
```
lbenet, line 11: Could you add a test that checks that an error (`MethodError`) is thrown if you do not make explicit the "numtype" `T` of the `TaylorN{T}` when using this constructor?

lbenet (commented):
Thanks a lot for this addition (and sorry for the time to send my review). In general, LGTM, though I have included a comment requesting to add another tests. Aside from that, once that is implemented, I'm happy to merge it.

test/manyvariables.jl
PerezHz: > Thanks a lot for this addition (and sorry for the time to send my review). In general, LGTM, though I have included a comment requesting to add another tests. Aside from that, once that is implemented, I'm happy to merge it.

Thanks for taking a look! I added your suggestion, so this is now ready for merging.

lbenet (approved):
Thanks a lot @PerezHz! PR approved, and I'll merge it.

PR #357, opened 2024-05-02
Fixes and additions in `evaluate`, including some in-line methods

lbenet:


lbenet:
cc @LuEdRaMo

lbenet:
Merged, and bumping a new patch version

PR #361, opened 2024-07-09
Further allocation reduction in power computations

PerezHz:
This PR continues in the spirit of #347 and #349, introducing an in-place form of power-by-squaring. It manages to avoid almost all allocations, except for one that I haven't managed to take care of so far, [here](https://github.com/PerezHz/TaylorSeries.jl/blob/6cce90ade660a4a583236a26cc64d93453160e4f/src/power.jl#L361). One way to get rid of it would be to simply add an argument to in-place methods of power, i.e, `pow!`, such that instead of calling such methods as `pow!(res, a, p, k)`, they would now be called as `pow!(res, a, p, aux, k)`. Or maybe there is another way to get rid of this auxiliary variable? Essentially the problem is that `power_by_squaring` uses some intermediate variables which don't allocate for non-composite numeric types (structs <: Number), but do allocate otherwise. What is your opinion on this @lbenet?


lbenet:
Thanks @PerezHz for addressing this!

Regarding the question on adding `pow!(res, a, p, aux, k)`, I also thought about this possibility, I agree this is a way to go. The sole inconvenient I see is that we will have to adapt `@taylorize` macro in TaylorIntegration, which is always a subtle change, yet worth it! Among other, we will have to adapt/extend the internal dictionaries for this 5-argument function. Once said this, I propose to continue with this PR, and check the performance wrt master. If the improvements are worth, we then adapt what is needed here and in TaylorIntegration.

src/arithmetic.jl
```diff
@@ -162,7 +162,10 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
 end
 
 for T in (:Taylor1, :TaylorN)
-    @eval zero(a::$T) = $T(zero.(a.coeffs))
+    @eval function zero(a::$T)
+        return $T(zero.(a.coeffs))
+        return za
```
lbenet, line 167: I guess this is some leftover of some tests you previously made, right?

src/arithmetic.jl
```diff
@@ -572,6 +575,19 @@ for T in (:Taylor1, :TaylorN)
     end
 end
 
+# in-place product: `a` <- `a*b`
+# this method computes the product `a*b` and saves it back into `a`
+# assumes `a` and `b` are of same order
+function mul!(a::TaylorN{T}, b::TaylorN{T}) where {T<:Number}
```
lbenet, line 581: An equivalent method `mul!(a::Taylor1{T}, b::Taylor1{T}) where {T<:Number})` could be useful for the extension of `pow!` for `Taylor1{TaylorN{T}}`?

src/arithmetic.jl
```diff
@@ -572,6 +575,19 @@ for T in (:Taylor1, :TaylorN)
...
+    for k in reverse(eachindex(a))
+        mul!(a, a, b[0][1], k)
+        for l in 1:k
+            mul!(a[k], a[k-l], b[l])
+        end
+    end
+    return nothing
+end
+
```
lbenet, line 590: Not sure if this is a good proposal, but here it goes: the code for the next method of `mul!` could be simplified using the new one proposed in this PR, e.g., by copying `a` into `res`, and then using `mul!(res, b)`.

src/power.jl
```diff
@@ -58,37 +58,95 @@ end
 
 ^(a::Taylor1{TaylorN{T}}, r::Rational) where {T<:NumberNotSeries} = a^(r.num/r.den)
 
+# in-place form of power_by_squaring
+# this method assumes `y`, `x` and `aux1` are of same order
+function power_by_squaring!(y::TaylorN{T}, x::TaylorN{T}, aux1::TaylorN{T}, p::Integer) where {T<:NumberNotSeries}
```
lbenet, line 63: Could an equivalent method be added for `Taylor1` inputs? Or are they not necessary?

src/power.jl
```diff
@@ -58,37 +58,95 @@ end
...
-        p >>= t
-        while (t -= 1) > 0
-            x = square(x)
-        end
-        y = x
-        while p > 0
+    @eval power_by_squaring(x::$T, p::Integer) = power_by_squaring(x, Val(p))
+    @eval power_by_squaring(x::$T, ::Val{0}) = one(x)
+    @eval power_by_squaring(x::$T, ::Val{1}) = copy(x)
+    @eval power_by_squaring(x::$T, ::Val{2}) = square(x)
```
lbenet, line 102: Perhaps adding `power_by_squaring(x::$T, ::Val{3})` could be useful too, since it includes the "odd" powers part of the function.

src/power.jl
```diff
@@ -58,37 +58,95 @@ end
...
+            end
         end
         return y
     end
 end
 
+power_by_squaring(x::TaylorN{Taylor1{T}}, ::Val{0}) where {T<:NumberNotSeries} = one(x)
+power_by_squaring(x::TaylorN{Taylor1{T}}, ::Val{1}) where {T<:NumberNotSeries} = copy(x)
+power_by_squaring(x::TaylorN{Taylor1{T}}, ::Val{2}) where {T<:NumberNotSeries} = square(x)
+function power_by_squaring(x::TaylorN{Taylor1{T}}, ::Val{P}) where {P, T<:NumberNotSeries}
```
lbenet, line 132: What about `Taylor1{TaylorN{T}}` method? Or is it not necessary?

lbenet:
I'v left few comments, which are simple suggestions

src/arithmetic.jl
PerezHz: Good catch, thanks! It's now fixed

PerezHz:
> Regarding the question on adding pow!(res, a, p, aux, k), I also thought about this possibility, I agree this is a way to go. The sole inconvenient I see is that we will have to adapt @taylorize macro in TaylorIntegration, which is always a subtle change, yet worth it! Among other, we will have to adapt/extend the internal dictionaries for this 5-argument function. Once said this, I propose to continue with this PR, and check the performance wrt master. If the improvements are worth, we then adapt what is needed here and in TaylorIntegration.

Alright, sounds good to me! I'll carry on forward then and report back after comparing performance wrt master.

src/power.jl
PerezHz: You mean adding something along the lines of `power_by_squaring(x::$T, ::Val{3}) = x*square(x)`?

src/arithmetic.jl
PerezHz: It would indeed by useful, I've added the corresponding method, thanks!

src/power.jl
PerezHz: I think it would be useful, yes, I've added it, thanks!

PerezHz:
> I'v left few comments, which are simple suggestions

Many thanks for your suggestions! I've added some of the methods you suggested (the modification to `pow!` call signature are still pending, I'm on it); things are working fine except for intervals 😅. Not sure what's happening there, since all the other tests are passing otherwise. Any hints about what might be happening here?

src/power.jl
lbenet: You are right, so please ignore my comment. Once said this, i think there will be no changes in `@taylorize` that involve `power_by_squaring!`. I thus mark this comment as resolved.

src/power.jl
lbenet: Good point... I guess this is the kind of question that we should define in terms of performance. My naive guess is that `x*square(x)` will allocate more than the direct call to `power_by_squaring`, but we should definetily check this.

lbenet:
I'm not sure where the problem is, yet, but it lies in `power_by_squaring` itself. Using this PR:
```julia
julia> t = Taylor1(10)
 1.0 t + 𝒪(t¹¹)

julia> @which Base.power_by_squaring(t-1,4)
power_by_squaring(x::Taylor1, p::Integer)
     @ TaylorSeries ~/.julia/dev/TaylorSeries/src/power.jl:102

julia> Base.power_by_squaring(t-1,4) # WRONG!
 1.0 - 4.0 t + 4.0 t² - 4.0 t³ + 1.0 t⁴ + 𝒪(t¹¹)
```

The last result is wrong(the second-order term which should be 6). You get similar problems using higher values of the power.

lbenet:
Tests are broken since [2e4e60a](https://github.com/JuliaDiff/TaylorSeries.jl/pull/361/commits/2e4e60af40611b37491440f65e97f43d2cd3cbe5), so we should begin there.

src/power.jl
```diff
@@ -447,6 +511,37 @@ for T = (:Taylor1, :TaylorN)
 
             return nothing
         end
+
+        # in-place squaring: given `c`, compute expansion of `c^2` and save back into `c`
+        @inline function sqr!(c::$T{T}, k::Int) where {T<:NumberNotSeries}
```
lbenet, line 516: I think the problem in the tests is related to this method. The following function
```julia
function mysquare(x::Taylor1)
     aux = copy(x)
     for k in reverse(eachindex(aux))
         TS.sqr!(aux, k)
     end
     return aux
 end
```
should compute the square of a Taylor1 object, and has the same problem for `(Taylor1(10)-1)^4`.

src/power.jl
```diff
@@ -447,6 +511,37 @@ for T = (:Taylor1, :TaylorN)
...
+            # Recursion formula
+            kodd = k%2
+            kend = (k - 2 + kodd) >> 1
+            if $T == Taylor1
+                (kend ≥ 0) && ( @inbounds c[k] = c[0] * c[k] )
+                @inbounds for i = 1:kend
+                    c[k] += c[i] * c[k-i]
+                end
+                @inbounds c[k] = 2 * c[k]
+                (kodd == 0) && ( @inbounds c[k] = c[k >> 1]^2 )
```
lbenet, line 531: I think the following solves the problem...

```suggestion
                (kodd == 0) && ( @inbounds c[k] += c[k >> 1]^2 )
```

PerezHz: Great catch! It does the trick, many thanks!

PerezHz:
Many thanks for the fix @lbenet! I've updated here the call signature for `pow!` and am in the process of doing the corresponding changes in TaylorIntegration; I'll report back as soon as this is done.

src/power.jl
```diff
@@ -58,37 +58,98 @@ end
...
+            y = x
+            while p > 0
+                t = trailing_zeros(p) + 1
+                p >>= t
+                while (t -= 1) ≥ 0
+                    x = square(x)
+                end
+                y *= x
+            end
         end
```
lbenet, line 127: I think this `if` block is now unnecessary, and can be replaced simply by:
```suggestion
        y = zero(x)
        aux = zero(x)
        power_by_squaring!(y, x, aux, p)
```

PerezHz: Thank you for noticing this @lbenet! I agree there's better ways to deal with this `if` block, I'm just wondering if it makes sense to have a `power_by_squaring!` for `HomogeneousPolynomial`, since inside such methods, there's a lot of in-place squaring going around, which would require a different treatment, since squaring a `HomogeneousPolynomial` changes its order, so some extra auxiliaries might be needed, I think? Anyway, I'm pushing a commit where the `power_by_squaring!` method for `HomogeneousPolynomial` is declared separately, so that this `if` does not occur inside the method.

PerezHz:
> I think now we should somehow check the performance implications.

As far as I've checked with TaylorIntegration (without ADS) and NEOs tests, performance does not improve that much from this change. Yet, I think it's worthwhile to keep this change, since then we can be sure that in-place operations performed with parsed jetcoeffs! methods are allocation-free. One thing I noticed is that in JT integrations, some allocations occur for example in the preamble of parsed jetcoeffs! methods. Since the preamble is called once per time-step, this can induce performance penalties. I'm looking into this, but that will be the target of another PR in TaylorIntegration. Otherwise, this PR is essentially ready for review.

src/power.jl
PerezHz: I just checked this, and it turns out doing `x*square(x)` allocates less than `power_by_squaring(x)` (!), do you think it's worth then to add the `::Val{3}` method?

src/power.jl
lbenet: I agree it is worth keeping *both* methods for `HomogeneousPolynomial`s. I think they can be useful in methods that exploit `evaluate`, which is worth checking, though perhaps in another PR.

src/power.jl
lbenet: Have you also teste with intervals? I would choose the case that behaves better wrt allocations.

lbenet:
> As far as I've checked with TaylorIntegration (without ADS) and NEOs tests, performance does not improve that much from this change. Yet, I think it's worthwhile to keep this change, since then we can be sure that in-place operations performed with parsed jetcoeffs! methods are allocation-free. 

I fully agree. I was checking where is `power_by_squaring` used, and is used for integer powers (when the parametric type is `Int` or `Interval`, with `TaylorN`s), obviously, and in `evaluate` of `HomogeneousPolynomial`s. I think the case of evaluating `HomogeneousPolynomial`s is where this PR could be quite important. And `pow!` should be interesting in combination with `@taylorize`.

src/power.jl
PerezHz: With intervals `power_by_squaring(x, 3)` on `Taylor1{Interval{Float64}}` allocates around half as `x*square(x)`, while for `TaylorN{Interval{Float64}}` allocations are essentially the same

src/power.jl
PerezHz: Okay, I added a comment with a TODO for another PR

PerezHz:
I added `power_by_squaring(x, ::Val{3})` methods, and some extra tests which check consistency between products and powers of Taylor1 and TaylorN variables. Also bumped the patch version.

PerezHz:
As things are standing now, in parsed jetcoeffs methods for JT integrations, such as:

```julia
function TaylorIntegration.jetcoeffs(::Val{myodefun!}, ...)
    order = t.order
    # preamble (0-th order)
    for k in 1:order
        # recursion loop
    end
    return nothing
end
```
we have now managed to make recursion loop essentially allocation free (except perhaps for some edge cases), but one of the lessons of this PR I think is that most allocations incurred in parsed `jetcoeffs!` methods for JT integrations are happening now in the preamble. I'll try to address this in TaylorIntegration.

PerezHz:
> I fully agree. I was checking where is power_by_squaring used, and is used for integer powers (when the parametric type is Int or Interval, with TaylorNs), obviously, and in evaluate of HomogeneousPolynomials. I think the case of evaluating HomogeneousPolynomials is where this PR could be quite important. And pow! should be interesting in combination with @taylorize.

On this point, together with @LuEdRaMo at JuliaCon we were figuring out a new `evaluate!` method for arrays of `TaylorN` evaluated at another `TaylorN`. I have a prototype for this and will open a PR in the next few days.

src/power.jl
```diff
@@ -58,38 +58,105 @@ end
...
+        power_by_squaring(x::$T, ::Val{1}) = copy(x)
+        power_by_squaring(x::$T, ::Val{2}) = square(x)
+        if $T != HomogeneousPolynomial
+            function power_by_squaring(x::$T, ::Val{P}) where P
+                p = P # copy static parameter `P` into local variable `p`
+                y = zero(x)
+                aux = zero(x)
+                power_by_squaring!(y, x, aux, p)
+                return y
+            end
```
lbenet, line 116: I think this `if` is overwritten for `HomogeneousPolynomial` by the method defined in line 118...

src/power.jl
```diff
@@ -58,38 +58,105 @@ end
...
+        end
+        y *= x
+    end
+    return y
+end
+
+
+power_by_squaring(x::TaylorN{Taylor1{T}}, ::Val{0}) where {T<:NumberNotSeries} = one(x)
+power_by_squaring(x::TaylorN{Taylor1{T}}, ::Val{1}) where {T<:NumberNotSeries} = copy(x)
+power_by_squaring(x::TaylorN{Taylor1{T}}, ::Val{2}) where {T<:NumberNotSeries} = square(x)
```
lbenet, line 142: I think these methods are defined above in lines 104-106...

src/power.jl
lbenet: Incidentally, I think this code in nicer than the one in line 118...

src/power.jl
```diff
@@ -184,18 +254,24 @@ exploits `k_0`, the order of the first non-zero coefficient of `a`.
...
-    elseif r == 1
-        return identity!(c, a, k)
-    elseif r == 2
-        return sqr!(c, a, k)
-    elseif r == 0.5
-        return sqrt!(c, a, k)
-    end
+# pow! main dispatcher
+@inline pow!(c::Taylor1{T}, a::Taylor1{T}, b::Taylor1{T}, r::S, k::Int) where
+        {T<:Number, S <: Real} = pow!(c, a, b, Val(r), k)
```
lbenet, line 262: Since `r` is not necessarily an `Int`, I fear that this method could overpopulate methods to dispatch, making slower the dispatch stage. 

If we add a concrete method for `Int` and for other numeric types (whose value is not an integer and will not use `Val`), I think this would be nice.

lbenet (commented):
Thanks @PerezHz again for addressing this! I've left few comments which I think are useful. Also, while I was reviewing this PR, I noticed you pushed some new commits. I therefore send this now, and will review back again later.

lbenet:
> I added `power_by_squaring(x, ::Val{3})` methods, and some extra tests which check consistency between products and powers of `Taylor1` and `TaylorN` variables. Also bumped the patch version.

Tests are passing, which is always nice!

> > I fully agree. I was checking where is power_by_squaring used, and is used for integer powers (when the parametric type is Int or Interval, with TaylorNs), obviously, and in evaluate of HomogeneousPolynomials. I think the case of evaluating HomogeneousPolynomials is where this PR could be quite important. And pow! should be interesting in combination with `@taylorize`.

> On this point, together with @LuEdRaMo at JuliaCon we were figuring out a new evaluate! method for arrays of TaylorN evaluated at another TaylorN. I have a prototype for this and will open a PR in the next few days.

Regarding the patch version, I think in this case (including the future PR you mention on `evaluate!`) we should release a minor version, since `pow!` changes a lot, and it is not a simple performance improvement or bug release. In any case, I think we should only bump that version once everything works nicely in TaylorIntegration, in particular wrt `@taylorize`.

PerezHz:
> Regarding the patch version, I think in this case (including the future PR you mention on evaluate!) we should release a minor version, since pow! changes a lot, and it is not a simple performance improvement or bug release. In any case, I think we should only bump that version once everything works nicely in TaylorIntegration, in particular wrt @taylorize.

That is actually a good point. I agree that if we bump a patch release for this PR this can break code relying on `pow!`, so should we skip tagging a release for this PR? Or maybe tagging a new patch version with deprecation warnings and then bumping the minor version without deprecation warnings?

PerezHz:
Just pushed a commit switching back from the dispatches-by-value to if/else in `pow!`

lbenet:
> That is actually a good point. I agree that if we bump a patch release for this PR this can break code relying on `pow!`, so should we skip tagging a release for this PR? Or maybe tagging a new patch version with deprecation warnings and then bumping the minor version without deprecation warnings?

I think this PR merits a new minor version release; once said this, we may merge it to master without tagging a new release, in order to include the ideas you have on `evaluate!`, or tagging it and then have a new patch version for the other PR (as long as it doesn't change too much the internals). In my opinion, we should wait to have both PRs merged and then tag the minor version, but I'm open to your suggestions.

PerezHz:
> I think this PR merits a new minor version release; once said this, we may merge it to master without tagging a new release, in order to include the ideas you have on evaluate!, or tagging it and then have a new patch version for the other PR (as long as it doesn't change too much the internals). In my opinion, we should wait to have both PRs merged and then tag the minor version, but I'm open to your suggestions.

Okay, let's merge this on master and then merge the corresponding PR on TaylorIntegration without tagging, and tag after we're sure everything is looking fine. The `evaluate!` PR I'll try to open in the next few days, in the end I think it won't change the internals too much.

PerezHz:
I reverted the change in the `setindex!` method for mixtures with `Taylor1` and added corresponding tests. Even though removing the `deepcopy` did not affect tests neither here nor in TaylorIntegration, it's perhaps better to look at it in more detail afterwards. The now-reverted change in `setindex!` also does not affect this PR directly.

lbenet:
> Okay, let's merge this on master and then merge the corresponding PR on TaylorIntegration without tagging, and tag after we're sure everything is looking fine. The `evaluate!` PR I'll try to open in the next few days, in the end I think it won't change the internals too much.

Fine with me. Naive question: have all suggestions of the review been addressed? Shall I make another (final) review?

PerezHz:
> Fine with me. Naive question: have all suggestions of the review been addressed? Shall I make another (final) review?

As far as I've looked I think I managed to address everything, but perhaps another quick round of review can help just in case I missed something. One thing to note is that I reverted the patch version bumping, so this PR does not change version (it will still be v0.17.8). Otherwise, from my perspective this is ready for merging.

lbenet:
I'll have a look today; regarding the version, v0.17.8 will be the tag of the proper branch, and this PR will be available from master. In other words, dev version and the last tagged version will be different.

src/power.jl
lbenet, line 1: Sorry, I still have two suggestions on this file:
- `power_by_squaring` should rely on `power_by_squaring!`; this is already the case for `Taylor1` and `TaylorN`, but not for `HomogeneousPolynomial` and `TaylorN{Taylor1}`
- The methods for `HomogeneousPolynomial` and `TaylorN{Taylor1}` seem identical, so I would exploit metaprogramming for them

lbenet: Let's for get this suggestions, and I'll go ahead and merge

lbenet (approved):
Thanks a lot. I'll go ahead and merge

lbenet:
Mrged!

PerezHz:
@lbenet sorry I think I pushed a commit on my fork just before you merged which I think better reflects where the mutating `power_by_squaring!` method is used, I can try to open a new PR with the remaining commit

PerezHz:
... and thanks for mergning by the way!

lbenet:
Sorry... I was simply too fast... (I thought over my suggestions and they may not be really crutial at this stage.) So a new PR is ok

PerezHz:
> Sorry... I was simply too fast... (I thought over my suggestions and they may not be really crutial at this stage.) So a new PR is ok

On the contrary thank you for looking into all this! Just opened a new PR (see #363)

PR #362, opened 2024-07-09
Avoid runtime-evaluated redundant `if`s

PerezHz:
Currently, there are some patterns in the code which go along the lines of
```julia
for T in (:Taylor1, :TaylorN)
    # NOTE: For $T = TaylorN, `mul!` *accumulates* the result of a * b in c[k]
    @eval @inline function mul!(c::$T{T}, a::$T{T}, b::$T{T}, k::Int) where {T<:Number}
        if $T == Taylor1
            @inbounds c[k] = a[0] * b[k]
            @inbounds for i = 1:k
                c[k] += a[i] * b[k-i]
            end
        else
            @inbounds mul!(c[k], a[0], b[k])
            @inbounds for i = 1:k
                mul!(c[k], a[i], b[k-i])
            end
        end
        return nothing
    end
# ...
end
```
(these lines can be be found around here:https://github.com/JuliaDiff/TaylorSeries.jl/blob/05858833de9434cdc910eaa09eec557a114e1d2f/src/arithmetic.jl#L511 )

I gather such patterns avoid code repetitions, while generally the "cost" of running these `if`s, which are evaluated at runtime, is quite low. Just for the sake of confirming this, I experimented a bit with generated functions, which allow to preserve the specialization for each type while avoiding code repetition, with the advantage of resolving them at compilation time, instead of being checked during run time once per method call. This can become important when handling mixtures of `Taylor1{TaylorN{T}}` in TaylorIntegration, where the number of coefficients can be quite large, and these if's can be checked many times during run time, when it's only necessary to check them at dispatch.

Gotten to this point, though, maybe an equally good solution would be to simply separate `Taylor1` and `TaylorN` methods for patterns such as above? The latter can be more friendly to develop, and would avoid the use of generated functions, which looks maybe a bit overkill here?

So essentially what I'm proposing is to either remove the `if $T == Taylor1 ...` in the way they are now, or keep them by using generated functions, such that performance can improve a bit. Benchmarks I've run with TaylorIntegration on JT integrations show ~5% performance improvement. Not much indeed, but can be meaningful for longer runs.

### DIFF
--- a/ext/TaylorSeriesIAExt.jl
+++ b/ext/TaylorSeriesIAExt.jl
@@ -163,7 +163,7 @@ function TS.sqr!(c::Taylor1{Interval{T}}, a::Taylor1{Interval{T}},
     TS.zero!(c, k)
     # Recursion formula
     kodd = k%2
-    kend = div(k - 2 + kodd, 2)
+    kend = (k - 2 + kodd) >> 1
     @inbounds for i = 0:kend
         c[k] += a[i] * a[k-i]
     end
--- a/ext/TaylorSeriesIAExt.jl
+++ b/ext/TaylorSeriesIAExt.jl
@@ -74,128 +74,161 @@ for T in (:Taylor1, :TaylorN)
 
         ^(a::$T{Interval{T}}, r::Rational) where {T<:NumTypes} = a^float(r)
 
-        function ^(a::$T{Interval{T}}, r::S) where {T<:NumTypes, S<:Real}
-            isinteger(r) && r ≥ 0 && return power_by_squaring(a, Integer(r))
-            a0 = _intersect_domain_nonstd(constant_term(a), interval(zero(T), T(Inf)))
-            @assert !isempty_interval(a0)
-            aux = one(a0^r)
-            aa = one(aux) * a
-            r == 0.5 && return sqrt(aa)
-            if $T == TaylorN
-                if TS._isthinzero(a0)
-                    throw(DomainError(aa,
-                    """The 0-th order TaylorN coefficient must be non-zero
-                    in order to expand `^` around 0."""))
-                end
-            end
-            return _pow(aa, r)
-        end
-
         # _pow
         function _pow(a::$T{Interval{S}}, n::Integer) where {S<:NumTypes}
             n < 0 && return _pow(a, float(n))
             return power_by_squaring(a, n)
         end
-
-        function _pow(a::$T{Interval{T}}, r::S) where {T<:NumTypes, S<:Real}
-            isinteger(r) && r ≥ 0 && return power_by_squaring(a, Integer(r))
-            a0 = _intersect_domain_nonstd(constant_term(a), interval(zero(T), T(Inf)))
-            @assert !isempty_interval(a0)
-            aux = one(a0^r)
-            a[0] = aux * a0
-            r == 0.5 && return sqrt(a)
-            a_order = get_order(a)
-            if $T == Taylor1
-                l0 = findfirst(a)
-                # Index of first non-zero coefficient of the result; must be integer
-                !isinteger(r*l0) && throw(DomainError(a,
-                    """The 0-th order Taylor1 coefficient must be non-zero
-                    to raise the Taylor1 polynomial to a non-integer exponent."""))
-                lnull = trunc(Int, r*l0 )
-                (lnull > a_order) && return $T( zero(aux), a_order)
-                c_order = l0 == 0 ? a_order : min(a_order, trunc(Int, r*a_order))
-            else
-                if TS._isthinzero(a0)
-                    throw(DomainError(a,
-                        """The 0-th order TaylorN coefficient must be non-zero
-                        in order to expand `^` around 0."""))
-                end
-                c_order = a_order
-            end
-            #
-            c = $T(zero(aux), c_order)
-            aux0 = zero(c)
-            for k in eachindex(c)
-                TS.pow!(c, a, aux0, r, k)
-            end
-            return c
-        end
     end
+end
+
+function ^(a::Taylor1{Interval{T}}, r::S) where {T<:NumTypes, S<:Real}
+    isinteger(r) && r >= 0 && return power_by_squaring(a, Integer(r))
+    a0 = _intersect_domain_nonstd(constant_term(a), interval(zero(T), T(Inf)))
+    @assert !isempty_interval(a0)
+    aux = one(a0^r)
+    aa = one(aux) * a
+    r == 0.5 && return sqrt(aa)
+    return _pow(aa, r)
+end
+
+function ^(a::TaylorN{Interval{T}}, r::S) where {T<:NumTypes, S<:Real}
+    isinteger(r) && r >= 0 && return power_by_squaring(a, Integer(r))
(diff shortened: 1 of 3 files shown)


PerezHz:
Just updated this PR to current `master`; here's an example of how `@code_lowered` shows the `if`s occur in compiled code, so those `if`s are evaluated at runtime. But with this PR those `if`s disappear:

```julia
using TaylorSeries
using InteractiveUtils

const TS = TaylorSeries

t = Taylor1(Float64, 4)
x, y = set_variables("x y", order=4)

ct = zero(t)
cx = zero(x)

@code_lowered t^2.5
@code_lowered x^2.5
@code_lowered TS.sqr!(ct, t, zero(t[0]), 2)
@code_lowered TS.sqr!(cx, x, zero(x[0][1]), 2)
```

PerezHz:
Probably worth mentioning that since I first opened this PR most of the occurrences of the pattern
```julia
@eval function f(x::$T)
    if $T == Taylor1
        ...
    else
        ...
    end
end
```
have been removed, which I would say is a good thing and that's what this PR was trying to solve. The important thing is not the use of generated functions per-se, but rather the removal of such `if`s in compiled code which are then evaluated at runtime. So in the current proposal generated functions are not used to solve this problem. I see most of the occurrences of this pattern in the IA extension of TaylorSeries.

PerezHz:
Essentially, if we have

```julia
for V in (:Int, :Float64)
    @eval begin
        function f(x::$V)
            if $V == Int
                println("I'm Int")
            else
                println("I'm Float64")
            end
        end
    end
end
methods(f) # two methods defined
f(1)
f(1.0)
```

Then,
```julia-repl
julia> @code_lowered f(1)
CodeInfo(
1 ─ %1  = Main.:(==)
│   %2  = Main.Int
│   %3  = Main.Int
│   %4  =   dynamic (%1)(%2, %3)
└──       goto #3 if not %4
2 ─ %6  = Main.println
│   %7  =   dynamic (%6)("I'm Int")
└──       return %7
3 ─ %9  = Main.println
│   %10 =   dynamic (%9)("I'm Float64")
└──       return %10
)

julia> @code_lowered f(1.0)
CodeInfo(
1 ─ %1  = Main.:(==)
│   %2  = Main.Float64
│   %3  = Main.Int
│   %4  =   dynamic (%1)(%2, %3)
└──       goto #3 if not %4
2 ─ %6  = Main.println
│   %7  =   dynamic (%6)("I'm Int")
└──       return %7
3 ─ %9  = Main.println
│   %10 =   dynamic (%9)("I'm Float64")
└──       return %10
)
```

The lowered code shows the `if`s are kept, and therefore those are evaluated at runtime. So this PR is proposing to replace the remaining type-conditional @eval methods with separate concrete methods in `^` and `sqr!` in `src/power.jl`, as has already been done in other parts of the code (as mentioned before, except for the IA extension).

PerezHz:
@lbenet since this PR was first opened, code in TaylorSeries has changed enough so that the issue is now almost gone except for some methods in `src/power.jl`. So I've changed those as well (without using generated functions), which I think can help avoid evaluating some `if`s at runtime, which can be important for nested types. So this is now ready for review.

lbenet:
Thanks for bringing this back. Yes, in some of the PRs that have happened since you open this PR/Draft, I avoided some of the if's, mostly in arithmetic.jl and functions.jl, if I recall correctly. The example [above](https://github.com/JuliaDiff/TaylorSeries.jl/pull/362#issuecomment-4416399940) is quite clear. So, I agree to push this further. Once said this, I'm surprised that the error message that appears in the old version for `@code_lowered t^2.5`, does not appear in the new one... Is that ok?

PerezHz:
> Once said this, I'm surprised that the error message that appears in the old version for @code_lowered t^2.5, does not appear in the new one... Is that ok?

Thanks for pointing this out; as it turns out, the missing error message in `@code_lowered t^2.5` was an unreachable code branch in the lowered code for `Taylor1` (it was actually only reachable for `TaylorN`). But the error message still happens at runtime through the `_pow` method, so using this PR:

```julia-repl
julia> using TaylorSeries

julia> t = Taylor1(Float64, 4)
 1.0 t + 𝒪(t⁵)

julia> t^2.5
ERROR: DomainError with  1.0 t + 𝒪(t⁵):
The 0-th order Taylor1 coefficient must be non-zero
to raise the Taylor1 polynomial to a non-integer exponent.
Stacktrace:
 [1] pow!(c::Taylor1{Float64}, a::Taylor1{Float64}, aux::Taylor1{Float64}, r::Float64, k::Int64)
   @ TaylorSeries ~/projects/TaylorSeries.jl/src/power.jl:229
 [2] _pow(a::Taylor1{Float64}, r::Float64)
   @ TaylorSeries ~/projects/TaylorSeries.jl/src/power.jl:94
 [3] ^(a::Taylor1{Float64}, r::Float64)
   @ TaylorSeries ~/projects/TaylorSeries.jl/src/power.jl:49
 [4] top-level scope
   @ REPL[7]:1
```

PerezHz:
@lbenet as it stands now this PR is in principle ready for a first round of review; but I'll try adding the corresponding changes to the IA extension

PerezHz:
@lbenet I ran an attempt to fix the same issues in the IA extension, I think this takes care of all occurrences of the pattern in the bodies of all TS methods (or at least that was the intention) so this is now ready for review

PerezHz:
In the end the strategy I went for avoiding `@generated` functions and instead went for concrete methods. In the future, if we find generated functions are more maintainable/less repetitive, we might want to try them. But for now I think we're well covered.

ext/TaylorSeriesIAExt.jl
```diff
@@ -74,128 +74,161 @@ for T in (:Taylor1, :TaylorN)
...
+        ::Interval{T}, k::Int) where {T<:NumTypes}
+    if k == 0
+        TS.sqr_orderzero!(c, a)
+        return nothing
+    end
+    # Sanity
+    TS.zero!(c, k)
+    # Recursion formula
+    kodd = k%2
+    kend = div(k - 2 + kodd, 2)
```
lbenet:
```suggestion
    kend = (k - 2 + kodd) >> 1
```

ext/TaylorSeriesIAExt.jl
```diff
@@ -74,128 +74,161 @@ for T in (:Taylor1, :TaylorN)
...
```
lbenet:
```suggestion
    kend = (k - 2 + kodd) >> 1
```

src/power.jl
lbenet:
```suggestion
aaux = r*(kprime-i) - i
@inbounds c[k] += aaux * c[i+lnull] * a[l0+kprime-i]
```

lbenet (commented):
Thanks a lot for your contribution. I left three minor suggestions, one is admitedly unreleated to this PR (but worth fixing, a type unstability). Can you also bump a patch version?

src/power.jl
PerezHz: nice catch!

PerezHz:
@lbenet thank you for all your comments, went through all of them so this is ready for another look

lbenet:
LGTM. Thanks again. Merging!

PR #363, opened 2024-07-17
Refactor power_by_squaring code (meant to be put into #361)

PerezHz:
This is a small update from #361 which follows from a suggestion by @lbenet. It updates a comment and refactors code such that it better expresses where is the mutating `power_by_squaring!` method used.


lbenet (approved):
LGTM. Thanks! Merging!

PR #368, opened 2024-08-03
Override Base.sum!

LuEdRaMo:
This PR leverages `TS.add!` to reduce allocations when summing an array of Taylors. For instance:
```julia
# Generate a vector of random Taylor1s
order = 25
n = 10_000
v = [Taylor1(randn(order)) for _ in 1:n]
# New sum! method
y = zero(v[1])
sum!(y, v)
# Compare with existing methods
y == +(v...) # true
y ≈ sum(v) # true
# Benchmark
@time +(v...)
# 0.002061 seconds (30.00 k allocations: 3.357 MiB)
@time sum(v)
# 0.001163 seconds (10.00 k allocations: 2.441 MiB)
@time begin
    y = zero(v[1])
    sum!(y, v)
end
# 0.000688 seconds (3 allocations: 320 bytes)
```

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -418,6 +418,15 @@ for T in (:Taylor1, :TaylorN)
     end
 end
 
+function sum!(v::TaylorN{S}, a::AbstractArray{HomogeneousPolynomial{S}}) where {S <: Number}
+    for i in eachindex(a)
+        for k in eachindex(v)
+            add!(v, v, a[i], k)
+        end
+    end
+    return nothing
+end
+
 ## Multiplication ##
 for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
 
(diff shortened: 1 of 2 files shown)


lbenet:
Thanks a lot for this addition! I just restarted the problematic job (in the tests); I guess it will run ok. Please ping me when you want me to review this PR.

src/arithmetic.jl
```diff
@@ -405,6 +405,18 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
     end
 end
 
+for T in (:Taylor1, :TaylorN)
+    @eval begin
+        function sum!(v::$T{S}, a::AbstractArray{$T{S}}) where {S <: Number}
```
lbenet, line 410: Outside from this for-loop, I suggest to add a method for `sum!` with the signature`sum!(v::TaylorN{S}, a::AbstractArray{HomogeneousPolynomial{S}})`, which is used in some `evaluate` and/or related functions.

LuEdRaMo: Thank you for the suggestion! I've added the corresponding method, although I don't know which method of `evaluate` will benefit the most from it.

PR #369, opened 2024-08-29
Fixes in evaluate and pow

lbenet:


lbenet:
This PR reduces marginally some allocations when using `pow` or `^`, in some cases reaching 10% of improvement. It also solves an issue appearing in TaylorModels with `pow`. Tests pass, so I merge this.

As a side comment, I note a dramatic performance hit from v1.10 to v1.11...

PerezHz:
> This PR reduces marginally some allocations when using pow or ^, in some cases reaching 10% of improvement. It also solves an issue appearing in TaylorModels with pow. Tests pass, so I merge this.

This PR looks pretty neat! Just o try to understand it better, did you find allocations happening in `pow!`? Or were the allocations coming from `pow` (without bang `!`)?

lbenet:
I think it was some sort of type unstability somewhere in the mess of `pow`, and some small fixes in `pow!`. But I see no dramatic improvements in time.

lbenet:
@PerezHz @LuEdRaMo I didn't check the consequences on NEOs.jl or PlanetaryEphemeris.jl; can you check it? (Though it's a bit late, I could revert the merge commit...)

PR #384, opened 2025-08-15
New `_horner!` method to evaluate an array of Taylor1{TaylorN{T}} on a Number

LuEdRaMo:
This PR adds a new method of `_horner!` to reduce allocations in the in-place evaluation of an array of `Taylor1{TaylorN{T}}`  on a `Number`, which is a common operation in the jet transport integrations of `TaylorIntegration`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
 repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
-version = "0.20.1"
+version = "0.20.2"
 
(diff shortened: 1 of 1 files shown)


lbenet:
Thanks a lot @LuEdRaMo for this contribution. LGTM! 

Could you tag a new patch version; I don't think it is necessary to tag a minor one? Then I merge it...

LuEdRaMo:
> Thanks a lot @LuEdRaMo for this contribution. LGTM!
> 
> Could you tag a new patch version; I don't think it is necessary to tag a minor one? Then I merge it...

The changes in this PR are non-breaking, so I think a new patch version is enough.

lbenet:
Merging and tagging! Thanks a lot!

PR #387, opened 2025-09-15
New method of `div!` for the division of a `NumberNotSeries` by a `Taylor1{T}`

LuEdRaMo:
This PR introduces a new method of `div!` designed to reduce allocations when dividing a `NumberNotSeries` by a `Taylor1{T}`. This optimization significantly reduces the number of allocations in `Taylor1{Taylor1{Float64}}` integrations with `TaylorIntegration.jl` which will be important in future PRs in `NEOs.jl`.

### DIFF
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -473,6 +473,13 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
     end
 end
 
+function subst!(v::Taylor1{Taylor1{T}}, a::Taylor1{Taylor1{T}}, k::Int) where {T <: TS.NumberNotSeries}
+    @inbounds for i in eachindex(v[k])
+        v[k][i] = -a[k][i]
+    end
+    return nothing
+end
+
 for T in (:Taylor1, :TaylorN)
     @eval begin
         function sum!(v::$T{S}, a::AbstractArray{$T{S}}) where {S <: Number}
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -640,6 +640,15 @@ function muladd!(c::Taylor1{T}, a::Taylor1{T}, b::Taylor1{T}, k::Int) where
     end
     return nothing
 end
+
+function muladd!(c::Taylor1{T}, a::Taylor1{T}, b::Taylor1{T}) where
+        {T<:NumberNotSeries}
+    for k in eachindex(c)
+        muladd!(c, a, b, k)
+    end
+    return nothing
+end
+
 # function muladd!(v::Taylor1{T}, a::Taylor1{T}, b::NumberNotSeries, k::Int) where
 #         {T<:NumberNotSeries}
 #     @inbounds v[k] += a[k] * b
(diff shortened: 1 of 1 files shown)


LuEdRaMo:
When running the following code
```Julia
begin
    using TaylorSeries
    import Random
    Random.seed!(1234)
    μ = rand()
    x = Taylor1([Taylor1(rand(7), 6) for _ in 1:26], 25)
    y = μ / x
    println("::NumberNotSeries / ::Taylor1{T}")
    @time y = μ / x
    z = deepcopy(x)
    for k in eachindex(z)
        TS.div!(z, μ, x, k)
    end
    println("TS.div!(::Taylor1{T}, ::NumberNotSeries, ::Taylor1{T})")
    @time for k in eachindex(z)
        TS.div!(z, μ, x, k)
    end
    println("|z - y |∞ = ", norm(z - y, Inf))
end
```
I get the following output in master
```
::NumberNotSeries / ::Taylor1{T}
  0.000055 seconds (102 allocations: 11.141 KiB)
TS.div!(::Taylor1{T}, ::NumberNotSeries, ::Taylor1{T})
  0.000449 seconds (2.82 k allocations: 272.703 KiB)
|z - y |∞ = 7.275957614183426e-12
```
vs in this PR
```
::NumberNotSeries / ::Taylor1{T}
  0.000058 seconds (102 allocations: 11.141 KiB)
TS.div!(::Taylor1{T}, ::NumberNotSeries, ::Taylor1{T})
  0.000031 seconds (35 allocations: 1.219 KiB)
|z - y |∞ = 2.7284841053187847e-12
```
@lbenet Any thoughts on why `|z - y |∞` changes?

lbenet:
First, thanks a lot and very nice that you addressed this missing method, and that it yields really good performance!

My guess is that the differences arise somewhere in the new method `div_scalar!`; if I use line 1254 instead of 1255, I get no difference in the result, but loose most performance improvements.

src/arithmetic.jl
```diff
@@ -1381,6 +1405,21 @@ end
...
+        a::Taylor1{T}, k::Int) where {T <: NumberNotSeries}
+    if k==0
+        @inbounds c[0] = scalar*constant_term(c) / constant_term(a)
+        return nothing
+    end
+
+    @inbounds c[k] = scalar * c[k]
+    @inbounds for i = 0:k-1
+        c[k] -= c[i] * a[k-i]
+    end
```
lbenet, line 1418: The following seems to solve the differences you noticed, but you need to test this thoroughly:

```suggestion
    aux = c[k] = scalar * c[k]
    @inbounds zero!(c, k)
    @inbounds for i = 0:k-1
        c[k] -= c[i] * a[k-i]
    end
    c[k] += aux
   ```

LuEdRaMo: Thanks! Your suggestion indeed solves the differences while retaining the performance gains.

LuEdRaMo:
The following test is failing in Julia LTS (Windows)
https://github.com/LuEdRaMo/TaylorSeries.jl/blob/40bb44f7374655070cf32ebca7cafd4dd266d8ce/test/onevariable.jl#L731

@lbenet Is that a theoretic/strict upper bound for the error on matrix multiplication?

lbenet:
It would be very nice to add a test like the one you wrote [above](https://github.com/JuliaDiff/TaylorSeries.jl/pull/387#issuecomment-3292398212). I guess that at some point you will change the method used by `y = μ / x` (which currently defaults to `/(promote(μ / x)...)`) to avoid promotion.

LuEdRaMo:
Unless you have further comments @lbenet, this PR is ready!

lbenet (approved):
LGTM, thanks a lot.

PR #388, opened 2025-09-19
Reduce memory allocation when squaring a nested Taylor1

LuEdRaMo:
In master, running the following block of code:
```Julia
begin
    using TaylorSeries
    import Random
    Random.seed!(1234)
    inorder = 6
    outorder = 25
    x = Taylor1([Taylor1(rand(inorder+1), inorder) for _ in 1:outorder+1], outorder)
    y, z = zero(x), zero(x)
    for r in [0, 1, 2, 3, 4]
        println("x^$r")
        for k in eachindex(z)
            TS.pow!(z, x, y, r, k)
        end
        @time for k in eachindex(z)
            TS.pow!(z, x, y, r, k)
        end
        println("|z - x^$r|∞ = ", norm(z - x^r, Inf))
    end
 end
```
produces the following output:
```
x^0
  0.000005 seconds (35 allocations: 1.219 KiB)
|z - x^0|∞ = 0.0
x^1
  0.000005 seconds (35 allocations: 1.219 KiB)
|z - x^1|∞ = 0.0
x^2
  0.000034 seconds (60 allocations: 3.953 KiB)
|z - x^2|∞ = 0.0
x^3
  0.000059 seconds (35 allocations: 1.219 KiB)
|z - x^3|∞ = 0.0
x^4
  0.000041 seconds (35 allocations: 1.219 KiB)
|z - x^4|∞ = 0.0
```
showing that squaring a nested Taylor1 allocates twice the memory. This PR proposes a quick fix.

### DIFF
--- a/src/power.jl
+++ b/src/power.jl
@@ -322,7 +322,7 @@ end
 end
 
 @inline function pow!(c::Taylor1{Taylor1{T}}, a::Taylor1{Taylor1{T}}, aux::Taylor1{Taylor1{T}},
-        r::S, k::Int) where {T<:NumberNotSeries, S<:Real}
+        r::S, k::Int) where {T<:NumberNotSeriesN, S<:Real}
     (r == 0) && return one!(c, a, k)
     (r == 1) && return identity!(c, a, k)
     (r == 2) && return sqr!(c, a, aux[k], k)
--- a/src/power.jl
+++ b/src/power.jl
@@ -321,11 +321,11 @@ end
     return nothing
 end
 
-@inline function pow!(c::Taylor1{T}, a::Taylor1{T}, aux::Taylor1{T},
-        r::S, k::Int) where {T<:NumberNotSeriesN, S<:Real}
+@inline function pow!(c::Taylor1{Taylor1{T}}, a::Taylor1{Taylor1{T}}, aux::Taylor1{Taylor1{T}},
+        r::S, k::Int) where {T<:NumberNotSeries, S<:Real}
     (r == 0) && return one!(c, a, k)
     (r == 1) && return identity!(c, a, k)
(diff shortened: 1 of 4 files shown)


src/power.jl
```diff
@@ -321,11 +321,11 @@ end
     return nothing
 end
 
-@inline function pow!(c::Taylor1{T}, a::Taylor1{T}, aux::Taylor1{T},
-        r::S, k::Int) where {T<:NumberNotSeriesN, S<:Real}
+@inline function pow!(c::Taylor1{Taylor1{T}}, a::Taylor1{Taylor1{T}}, aux::Taylor1{Taylor1{T}},
+        r::S, k::Int) where {T<:NumberNotSeries, S<:Real}
```
lbenet, line 325: What about use the limit `T<:NumberNotSeriesN`? That would allow to use this function in two and more nested Taylor1 structs...

(I think that the same comment applies to the `div!` method introduced in #387.)

lbenet:
Incidentally, when you bump the patch version related to this PR, can you also include compatibility with IA v0.23 (so it works with 0.23 and 1 as well)? See [this](https://github.com/JuliaIntervals/TaylorModels.jl/pull/178#issuecomment-3315088736) for some context.

LuEdRaMo:
> Incidentally, when you bump the patch version related to this PR, can you also include compatibility with IA v0.23 (so it works with 0.23 and 1 as well)? See [this](https://github.com/JuliaIntervals/TaylorModels.jl/pull/178#issuecomment-3315088736) for some context.

Done. 

If you have no further comments, I think this PR is ready...

lbenet:
Thanks a lot for this improvement! Is this ready (after the tests pass) to be merged?

PR #403, opened 2026-04-06
Inplace multiplicacion for Matrix{Taylor1} and Matrix{TaylorN}

lbenet:


lbenet:
I used the name `matmultay!` to still allow `LinearAlgebra.mul!` to work, but it admitedly is awful. Other suggestions @LuEdRaMo ?

LuEdRaMo:
> I used the name `matmultay!` to still allow `LinearAlgebra.mul!` to work, but it admitedly is awful. Other suggestions @LuEdRaMo ?

Other taylorized methods in TS do not use the suffix `tay`, e.g., `add!`, `subst!`, `mul!`, `div!`, etc. So, I would suggest to simply call the function `matmul!`. I believe there is no `LinearAlgebra.matmul!` and, even if it were the case, the should not be a problem as long as TS does not export that name.

lbenet:
There s no matmul! indeed... ok, let's use it.

PR #409, opened 2026-05-09
New functions `jacobianmatrix` and `hessianmatrix`

LuEdRaMo:
This PR defines two new functions `jacobianmatrix` and `hessianmatrix` which compute the matrix of first (second) partial derivatives of a `Vector{TaylorN}` (`TaylorN`) without evaluating.

@lbenet Any comments?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
-version = "0.21.7"
+version = "0.21.8"
 repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
 
 [deps]
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -41,6 +41,7 @@ integrate
 gradient
 jacobian
 jacobian!
+jacobianmatrix
 hessian
 hessian!
 hessianmatrix
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -43,6 +43,7 @@ jacobian
 jacobian!
 hessian
 hessian!
+hessianmatrix
 constant_term
 linear_polynomial
 nonlinear_polynomial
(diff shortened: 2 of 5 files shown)


lbenet:
Thanks for this addition. I think it would be nice to document the difference with `hessian` (or `hessian!`, and include it in the documentation. I guess the difference is the evaluation, which doesn't occur here, right?

LuEdRaMo:
> Thanks for this addition. I think it would be nice to document the difference with `hessian` (or `hessian!`, and include it in the documentation. I guess the difference is the evaluation, which doesn't occur here, right?

Indeed, the new functions do not evaluate the jet transport deltas, thus preserving the `TaylorN` form of the derivatives.

Do you mean the documentation in `docs/` or the function's docstrings?

P.S. In the last commit I added an analogous new function for the jacobian.

lbenet:
Thanks again, this looks really useful. 

> Do you mean the documentation in docs/ or the function's docstrings?

Yes, I meant to add their existence to the userguide (`docs/src/userguide.md`). 

I wonder if, by rewritting the current functions `jacobian` and `hessian` (and their in-place versions) using `jacobianmatrix` and `hessianmatrix`, they could have better performance. Have you checked this?

LuEdRaMo:
> Yes, I meant to add their existence to the userguide (`docs/src/userguide.md`).

Done.
 
> I wonder if, by rewritting the current functions `jacobian` and `hessian` (and their in-place versions) using `jacobianmatrix` and `hessianmatrix`, they could have better performance. Have you checked this?

I did thought of this, but it seems to me that `jacobian` and `hessian` can implement some optimizations that `jacobianmatrix` and `hessianmatrix` cannot, since the former do not need to compute the full expansion of the partial derivatives to evaluate them (specially when the evaluation point is the origin).

LuEdRaMo:
If you have no further comments I think this is ready.

Should I bump the patch version?

lbenet:
> Should I bump the patch version?

LGTM. I think we should indeed bump the patch version, then I merge, and release the new version.

LuEdRaMo:
> LGTM. I think we should indeed bump the patch version, then I merge, and release the new version.

Done!

lbenet:
Thanks a lot!! Merging

PR #410, opened 2026-05-17
Introduce explicit `JetSpace` for multivariate Taylor algebras

PerezHz:
This PR introduces explicit multivariate Taylor algebra spaces via `JetSpace`, replacing the previous implicit reliance of `TaylorN` on a single global multivariate algebra while preserving the existing workflow using `set_variables`, etc. A `JetSpace` owns the truncation order, variable metadata, hash/index lookup tables, coefficient tables, position tables, and multiplication-table cache used by `HomogeneousPolynomial` and `TaylorN` arithmetic.

`HomogeneousPolynomial` and `TaylorN` now carry a `space::JetSpace` field, so multivariate polynomial objects are associated with the algebra that created them. Constructors were extended so users can construct variables and constants either from the default space defined through `set_variables` or from an explicit space.  

The existing `set_variables(...)` workflow remains supported and therefore this PR is backwards-compatible. Internally, `set_variables` now creates/updates the default `JetSpace`, synchronizes legacy global tables, and returns variables from that default space. This keeps existing user code working while enabling the new explicit-space API. This has been tested by running `TaylorIntegration` and `NEOs` tests, which both pass with essentially no change (except for a few numerical thresholds in tests).

This PR also adds same-space checks for `TaylorN` and `HomogeneousPolynomial` operations. Mixed-space arithmetic now throws an informative error, avoiding accidental operations between incompatible algebras. It also adds exports for `JetSpace` and `variables` .

The multiplication table machinery has been moved into the space abstraction, including cached homogeneous product schedules protected by a `ReentrantLock`. This has enabled some improvements in performance in preliminary tests, especially in products of nested types and jet transport integrations, but dedicated benchmarks should be run. 

@lbenet: ~dedicated `show` methods for `JetSpace` are still WIP (suggestions are welcome)~ (EDIT: added `show` methods as suggested by @LuEdRaMo), but as it stands now this PR is ready for a first round of review.

@LuEdRaMo: locally NEOs.jl tests pass although I did have to change some thresholds for the JPL compatibility tests. With this change, NEOs.jl tests are also passing [on CI](https://github.com/PerezHz/NEOs.jl/actions/runs/26002250804): can you please check how does performance in NEOs.jl change by using this branch of TS?

### Some examples

```julia-repl
julia> using TaylorSeries

julia> space1 = JetSpace(order=4, variables=[:x, :y]);

julia> space2 = JetSpace(order=3, variables=[:a, :b]);

julia> x, y = variables(space1)
  1.0 x + 𝒪(‖x‖⁵)
  1.0 y + 𝒪(‖x‖⁵)

julia> a, b = get_variables(space2)
  1.0 a + 𝒪(‖x‖⁴)
  1.0 b + 𝒪(‖x‖⁴)

julia> f = exp(x + y) + x^2*y
 1.0 + 1.0 x + 1.0 y + 0.5 x² + 1.0 x y + 0.5 y² + 0.16666666666666666 x³ + 1.5 x² y + 0.5 x y² + 0.16666666666666666 y³ + 0.041666666666666664 x⁴ + 0.16666666666666666 x³ y + 0.25 x² y² + 0.16666666666666666 x y³ + 0.041666666666666664 y⁴ + 𝒪(‖x‖⁵)

julia> g = sin(a*b) + a^2
 1.0 a² + 1.0 a b + 𝒪(‖x‖⁴)

julia> x + a   # throws ArgumentError
ERROR: ArgumentError: TaylorN algebra mismatch: operands belong to different spaces. Use an explicit projection or conversion before combining them.
Stacktrace:
 [1] _space_mismatch_error(space_a::TaylorNSpace, space_b::TaylorNSpace)
   @ TaylorSeries ~/projects/TaylorSeries.jl/src/auxiliary.jl:24
 [2] _check_same_space
   @ ~/projects/TaylorSeries.jl/src/auxiliary.jl:30 [inlined]
 [3] _check_same_space
   @ ~/projects/TaylorSeries.jl/src/auxiliary.jl:33 [inlined]
 [4] +(a::TaylorN{Float64}, b::TaylorN{Float64})
   @ TaylorSeries ~/projects/TaylorSeries.jl/src/arithmetic.jl:20
 [5] top-level scope
   @ REPL[8]:1
```


PerezHz:
Docs build failed, but tests passed...

UPDATE: fixed docs build it's now passing

src/parameters.jl
```diff
@@ -84,17 +84,90 @@ ParamsTaylorN(order, num_vars, variable_names) = ParamsTaylorN(order, num_vars,
...
+    mul_table        :: Vector{Vector{HomogeneousProductTable}}
+    mul_table_lock   :: ReentrantLock
+end
+
+const JetSpace = TaylorNSpace
+
+const default_space = Ref{TaylorNSpace}()
+
+Base.deepcopy_internal(space::TaylorNSpace, stackdict::IdDict) = space
+
```
LuEdRaMo, line 179: I would suggest to print a `TaylorNSpace` as follows:
```Julia
function Base.show(io::IO, s::TaylorNSpace)
    tab = "    "
    println(io,
        "TaylorNSpace\n",
        tab, rpad("Expansion order:",     24), get_order(s),            '\n',
        tab, rpad("Number of variables:", 24), get_numvars(s),          '\n',
        tab, rpad("Variable names:",      24), get_variable_names(s),   '\n',
        tab, rpad("Variable symbols:",    24), get_variable_symbols(s), '\n',
    )
end
```
which produces 
```
TaylorNSpace
    Expansion order:        4
    Number of variables:    2
    Variable names:         ["x", "y"]
    Variable symbols:       [:x, :y]
```

LuEdRaMo:
Nice work @PerezHz! This looks very promising. Before running more extensive tests in NEOs, I have a couple of questions:
- AFAIU, now every `HomogenousPolynomial` contains a `TaylorNSpace`. Does this increase the size in memory of such objects? If so, that could potentially be detrimental to packages like TaylorIntegration and NEOs which use high-order `Taylor1{TaylorN}}` nested polynomials.
- How do computation time and memory allocation change with respect to master (e.g., for the TS tests). Is that change significant at this level?

lbenet:
Thanks a lot for the nice idea @PerezHz! I think the concern of @LuEdRaMo is a fair one. I'll try to come back and have a closer look to this in the next few days.

PerezHz:
> AFAIU, now every HomogenousPolynomial contains a TaylorNSpace. Does this increase the size in memory of such objects? If so, that could potentially be detrimental to packages like TaylorIntegration and NEOs which use high-order Taylor1{TaylorN}} nested polynomials.

To answer in one word, yes, the extra `::space` field adds to memory allocation, but I would note julia does this by reference, i.e., the `::space` field is just a pointer to the actual `space` object (i.e., not an inline copy of the whole space), so for a `TaylorN` this is one pointer per `HomogenenousPolynomial` and one pointer for the whole `TaylorN` to the same `space` object which defines a particular algebra. 

This by itself does not change performance a lot, but this PR also adds some pre-computed product tables which help a lot, especially when computing nested-type products like the ones used in jet transport, which is the main rationale for this PR. In some preliminary benchmarks I ran for jet transport in TaylorIntegration I'm seeing 20% to 40% speedups when using this PR compared to master. In a sense, this is a two-fold PR: it adds the `TaylorNSpace`, and improves internal multiplication methods using pre-computed product tables: these are orthogonal changes in the sense they could be added in separate, self-contained PRs.

> How do computation time and memory allocation change with respect to master (e.g., for the TS tests). Is that change significant at this level?

I see a similar performance in time and allocations in this PR compared to master when looking at the package-level test suite; and in particular I do not see a significant package-level allocation regression in `TaylorSeries` tests. Full `Pkg.test()` comparison:

master: time 86 s, allocation 962.91 MB
branch: time 90s, allocation 968.18 MB

also this PR adds some tests, so it's testing a bit more stuff, but also it's true for very high-orders e.g. initialization in Fateman tests use order 40, the initialization overhead can be noticeable. But the tradeoff is in the heavy operation loops: those are running faster in this PR vs `master`. So e.g. Fateman test is a bit slower overall because of initialization overhead, but workflows like JT in NEOs can be quicker precisely because nested-type objects are initialized only once and reused a lot. I would say the differences in time/allocations as quoted above for the whole test suite of TaylorSeries do not reflect potential performance wins in JT workflows, where a heavy load of arithmetic operations between nested types occur. Why is also why I'm adding a benchmark for nested-type products in the `perf/` folder, so that we can keep track of this.

PerezHz:
> Thanks a lot for the nice idea @PerezHz! I think the concern of @LuEdRaMo is a fair one. I'll try to come back and have a closer look to this in the next few days.

Thanks! Sure, happy to discuss so that this can really become a clear step forward!

PerezHz:
@lbenet I think you closed this 😅 perhaps by mistake, I hope?

lbenet:
Uuuups.... yes, I closed it by mistake....

PerezHz:
Thanks for reopening!

src/hash_tables.jl
```diff
@@ -48,6 +48,120 @@ function generate_tables(num_vars, order)
...
+end
+
+@inline function _product_table(space::TaylorNSpace, order_a::Int, order_b::Int)
+    @inbounds table = space.mul_table[order_a][order_b]
+    !isempty(table.input_positions) && return table
+    return _init_product_table!(space, order_a, order_b)
+end
+
+function _init_product_table!(space::TaylorNSpace, order_a::Int, order_b::Int)
+    degree_a = order_a - 1
```
lbenet: Why `degree_a = order_a - 1`, and the same for `degree_b`? Don't you get in problems if one of the orders is 1?

src/hash_tables.jl
```diff
@@ -48,6 +48,120 @@ function generate_tables(num_vars, order)
...
+        throw(DimensionMismatch("homogeneous product order exceeds TaylorNSpace order"))
+    lock(space.mul_table_lock)
+    try
+        @inbounds table = space.mul_table[order_a][order_b]
+        if isempty(table.input_positions)
+            table = _homogeneous_product_table(space.index_table, space.pos_table,
+                degree_a, degree_b)
+            @inbounds space.mul_table[order_a][order_b] = table
+        end
+        return table
```
lbenet, line 148: Should the lock be unlocked before leaving this function?

src/parameters.jl
```diff
@@ -84,17 +84,90 @@ ParamsTaylorN(order, num_vars, variable_names) = ParamsTaylorN(order, num_vars,
...
+Precomputed input-pair and output-major schedules for multiplying two
+homogeneous polynomials in a single `TaylorNSpace`.
+"""
+mutable struct HomogeneousProductTable
+    input_positions :: Vector{Int}
+    output_offsets  :: Vector{Int}
+    output_pairs    :: Vector{UInt32}
+    num_right       :: Int
+end
```
lbenet, line 131: (Comment on lines 93-98) 
Why do you need `HomogeneousProductTable` mutable? Why `output_pairs` is a `Vector{UInt32}`?

src/parameters.jl
```diff
@@ -84,17 +84,90 @@ ParamsTaylorN(order, num_vars, variable_names) = ParamsTaylorN(order, num_vars,
...
+mutable struct TaylorNSpace
+    order            :: Int
+    num_vars         :: Int
+    variable_names   :: Vector{String}
+    variable_symbols :: Vector{Symbol}
+    coeff_table      :: Vector{Vector{Vector{Int}}}
+    index_table      :: Vector{Vector{Int}}
+    size_table       :: Vector{Int}
+    pos_table        :: Vector{Dict{Int,Int}}
+    mul_table        :: Vector{Vector{HomogeneousProductTable}}
```
lbenet, line 167: It is a nice idea to have the multiplication table somewhere stored. I'm not yet quite far in reading the code, but in [the example](https://github.com/JuliaDiff/TaylorSeries.jl/pull/410#issue-4464675692) I see that `space1.mul_table` has lots of *trivial* `HomogeneousProductTable` entries, and I think all entries have empty vectors for the `output_offsets` and `output_pairs`. Maybe you should add some comments or docstrings explaining what are their role (or when they become non-trivial).

src/parameters.jl
```diff
@@ -110,6 +183,14 @@ get_variables(::Type{T}, order::Int=get_order()) where {T} =
     [TaylorN(T, i, order=order) for i in 1:get_numvars()]
 get_variables(order::Int=get_order()) =
     [TaylorN(Float64, i, order=order) for i in 1:get_numvars()]
+variables(::Type{T}, space::TaylorNSpace; order::Int=get_order(space)) where {T} =
+    [TaylorN(space, T, i, order=order) for i in 1:get_numvars(space)]
+variables(space::TaylorNSpace; order::Int=get_order(space)) =
+    variables(Float64, space, order=order)
+get_variables(::Type{T}, space::TaylorNSpace; order::Int=get_order(space)) where {T} =
+    variables(T, space, order=order)
+get_variables(space::TaylorNSpace; order::Int=get_order(space)) =
```
lbenet: (Comment on lines 186-192)

Why do we need *both* `variables` and `get_variables`? If your aim is to replace `get_variables` with `variables`, which indeed may be a good reason, perhaps it is more julian to use `getvariables` (and similarly `getnumvars` and `setvariables`, without underscore).

Related comment: I don't know if using the comprehension `[TaylorN(space, T, i, order=order) for i in 1:get_numvars(space)]` is better over using broadcasting `TaylorN.(space, T, 1:get_numvars(space), order=order)`. The comment follows because in previous line 174 (which used broadcasting) was replaced by the comprehension.

LuEdRaMo:
Below are the results for NEOs' orbit determination benchmark using TaylorSeries master vs this PR:
```Julia
# master

BenchmarkTools.Trial: 10 samples with 1 evaluation per sample.
 Range (min … max):  1.322 s …   1.364 s  ┊ GC (min … max): 2.62% … 2.89%
 Time  (median):     1.341 s              ┊ GC (median):    2.79%
 Time  (mean ± σ):   1.343 s ± 15.427 ms  ┊ GC (mean ± σ):  2.80% ± 0.21%

  ▁▁                 ▁▁  ▁   ▁ ▁                      ▁   █  
  ██▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁██▁▁█▁▁▁█▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁█ ▁
  1.32 s         Histogram: frequency by time        1.36 s <

 Memory estimate: 497.75 MiB, allocs estimate: 7695692.

# This PR

BenchmarkTools.Trial: 10 samples with 1 evaluation per sample.
 Range (min … max):  923.942 ms …   1.145 s  ┊ GC (min … max): 3.03% … 21.07%
 Time  (median):     935.480 ms              ┊ GC (median):    4.04%
 Time  (mean ± σ):   955.141 ms ± 67.009 ms  ┊ GC (mean ± σ):  6.02% ±  5.47%

  █ ▃ ▃                                                         
  █▁█▇█▁▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇ ▁
  924 ms          Histogram: frequency by time          1.14 s <

 Memory estimate: 464.42 MiB, allocs estimate: 6902908.
```

In summary, approximately -29% in mean computation time, -7% in memory and -10% in allocations.

I am yet to verify that results remain consistent, but these numbers are very encouraging!

src/hash_tables.jl
```diff
@@ -110,6 +224,50 @@ end
...
+        dst.num_vars = space.num_vars
+        dst.variable_names = space.variable_names
+        dst.variable_symbols = space.variable_symbols
+        dst.coeff_table = space.coeff_table
+        dst.index_table = space.index_table
+        dst.size_table = space.size_table
+        dst.pos_table = space.pos_table
+        dst.mul_table = space.mul_table
+        dst.mul_table_lock = space.mul_table_lock
+        dst
```
lbenet: ?

src/hash_tables.jl
```diff
@@ -110,6 +224,50 @@ end
...
+    else
+        default_space[] = space
+        space
```
lbenet: ?

src/constructors.jl
```diff
@@ -180,13 +197,21 @@ Note that `TaylorN` variables are callable. For more information, see
 """
 struct TaylorN{T<:Number} <: AbstractSeries{T}
         coeffs  :: FixedSizeVectorDefault{HomogeneousPolynomial{T}}
+        space   :: TaylorNSpace
```
lbenet: Since for a `TaylorN{T}` we have `coeffs::FixedSizeVectorDefault{HomogeneousPolynomial{T}}`, and we check that all the components do belong to the same `TaylorNSpace`, can we spare having a field `space` for `TaylorN` (and simply obtain it from `coeffs[1]`)? (Perhaps a very small saving in allocations.)

lbenet: Note that some of the constructors, using a `TaylorNSpace` are still word having!

src/auxiliary.jl
```diff
@@ -8,34 +8,90 @@
...
+@inline space(a::TaylorN) = a.space
+
+function _space_mismatch_error(space_a::TaylorNSpace, space_b::TaylorNSpace)
+    throw(ArgumentError(
+        "TaylorN algebra mismatch: operands belong to different spaces. " *
+        "Use an explicit projection or conversion before combining them."))
+end
+
+@inline function _check_same_space(space_a::TaylorNSpace, space_b::TaylorNSpace)
+    space_a === space_b || _space_mismatch_error(space_a, space_b)
```
lbenet, line 40: Note the following:
```julia
julia> space1 = JetSpace(order=4, variables=[:x, :y]); # from your example

julia> space3 = JetSpace(order=4, variables=[:x, :y]); # same line, other name for the space

julia> space1 === space3
false
```
Is this expected?

src/auxiliary.jl
```diff
@@ -8,34 +8,90 @@
...
+    return nothing
+end
+@inline _check_same_space(a::Union{HomogeneousPolynomial,TaylorN},
+    b::Union{HomogeneousPolynomial,TaylorN}) =
+        _check_same_space(space(a), space(b))
+@inline function _check_same_space(a::Union{HomogeneousPolynomial,TaylorN},
+        b::Union{HomogeneousPolynomial,TaylorN},
+        c::Union{HomogeneousPolynomial,TaylorN})
+    _check_same_space(a, b)
+    _check_same_space(a, c)
```
lbenet, line 52:
```suggestion
    _check_same_space(a, b, c)
```

lbenet (commented):
First, thanks a lot @PerezHz for this PR, I think it is a very nice and powerful addition!

I've left few comments, which at this point may or may not be helpful. I'm currently concentrated in understanding how you are forming the `mul_table` property of the `HomogeneousProductTable`, which I find a brilliant addition and may truly speed up the package. I would think that they should be created at the  `TaylorNSpace` construction step, which I think it is not quite yet done at the moment (some "trivial" table is constructed). I'll be digging deeply onto this in the next few days.

src/hash_tables.jl
PerezHz: Thanks for this comment. I agree that the names can be confusing; the names actually mean 1-based indices. So index 1 means homogeneous degree 0, index 2 means degree 1, etc. So `degree_a = index_a - 1` is safe when the index is 1. I have updated the names and added comments and docstrings to this and other methods so that this is more explicit.

src/hash_tables.jl
PerezHz: Yes, I put the `try`/`finally` block precisely to guarantee that the lock is guaranteed to be unlocked no matter what (the `finally` is executed just before `return`). This avoids later code hanging when trying to acquire the same lock if e.g. `_homogeneous_product_table` throws.

src/parameters.jl
PerezHz: Since `output_offsets` and `output_pairs` are loaded lazily, having this struct mutable helps so that it can be re-used without having to construct a new `HomogeneousProductTable` object (which would be how things run with an immutable struct). Having said that, it's not strictly necessary, and we could certainly revisit this if we find we prefer immutable.

src/parameters.jl
PerezHz: Thanks for this suggestion and I completely agree that this should be better documented. I'm adding comments and extended docstrings for `HomogeneousProductTable` and its fields.

PerezHz: About the many entries with trivial entries, these are all related to the fact that each `HomogeneousProductTable` is loaded lazily, meaning they are initialized empty and are filled only when they are required (via `_product_table`, `_init_output_major_product_table!`), which helps reduce a lot the initialization overhead of these fields. I realized this contribution was important when a saw a huge regression in the runtime of the Fateman tests.

src/parameters.jl
lbenet: Thanks for the details. I thought about an immutable struct because all contents are mutable, except for `order` and `numvars`. But, let's leave it as it is now.

PerezHz:
Thanks a lot for the feedback and for having a look @lbenet! I'm still going through your comments and I agree this should all be better documented, so I'm adding docstrings and comments for all of the new infrastructure, if there's more details you think are worth adding please let me know and I can add it.

> I would think that they should be created at the  TaylorNSpace construction step, which I think it is not quite yet done at the moment (some "trivial" table is constructed).

About this, I tried this at first, but Fateman tests were becoming really slow, so I opted for a lazy construction of the `output_pairs` and `output_offsets` fields in `HomogeneousProductTable`, and that seems to give the best of both worlds: minimal initialization overhead, while optimizing speed on heavy-loads of mathematical operations between `TaylorN`s and nested types.

src/parameters.jl
lbenet: Thanks for the explanation. I see that constructing the mul_tables all at once may be very costly for Fatemann tests.

lbenet: One more suggestion: I see that `mul_table` is used as `mul_table[index_a][index_b]`; wouldn't it help to define `mul_table` as a `Matrix`? (I even think the that matrix would be symmstric.)

src/arithmetic.jl
```diff
@@ -16,6 +16,9 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
                 ($f)(promote(a, b)...)
 
             function ($f)(a::$T{T}, b::$T{T}) where {T<:Number}
+                if $T == TaylorN
+                    _check_same_space(a, b)
+                end
```
lbenet: Perhaps, we should add methods for `_check_same_space` which involve `Taylor1` and that trivially return `nothing`; we would avoid this `if` (following #362).

src/arithmetic.jl
```diff
@@ -34,15 +37,44 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
...
+                if $T == TaylorN
+                    R = promote_type(T, S)
+                    aa = convert(TaylorN{R}, a)
+                    bb = convert(R, b)
+                    c = TaylorN(aa.coeffs[:], get_order(aa))
+                    if $(QuoteNode(f)) == :+
+                        c[0][1] = aa[0][1] + bb
+                    else
+                        c[0][1] = aa[0][1] - bb
+                    end
```
lbenet:
```suggestion
                    c[0][1] = ($f)(aa[0][1], bb)
```

src/arithmetic.jl
```diff
@@ -34,15 +37,44 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
...
+                    R = promote_type(T, S)
+                    aa = convert(TaylorN{R}, a)
+                    bb = convert(R, b)
+                    if $(QuoteNode(f)) == :+
+                        c = TaylorN(aa.coeffs[:], get_order(aa))
+                        c[0][1] = bb + aa[0][1]
+                    else
+                        c = TaylorN((-aa).coeffs[:], get_order(aa))
+                        c[0][1] = bb - aa[0][1]
+                    end
```
lbenet:
```suggestion
                    c = TaylorN(($f)(aa.coeffs[:]), get_order(aa))
                    c[0][1] = bb + aa[0][1]
```

src/arithmetic.jl
```diff
@@ -417,12 +457,19 @@ end
 
 for (T, W) in ((:Taylor1, :Number), (:TaylorN, :NumberNotSeriesN))
     @eval function *(a::$T{T}, b::$T{T}) where {T<:$W}
+        if $T == TaylorN
+            _check_same_space(a, b)
+        end
```
lbenet: See [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/410/changes#r3267938157)

src/arithmetic.jl
```diff
@@ -417,12 +457,19 @@ end
...
             a, b = fixorder(a, b)
         end
         c = zero(a)
         for ord in eachindex(c)
-            mul!(c, a, b, ord) # updates c[ord]
+            if $T == TaylorN
+                _muladd_unchecked!(c, a, b, ord) # updates c[ord]
+            else
+                mul!(c, a, b, ord) # updates c[ord]
+            end
```
lbenet: What about defining `_muladd_unchecked!` for `Taylor1`s, which returns `mul!(c, a, b, ord)`?

lbenet: Question: Would this internal change (the existence of `_muladd_unchecked!`, not necessarily the suggestion above) impose changes related to `@taylorize`?

src/arithmetic.jl
```diff
@@ -766,39 +830,222 @@ c_k = \sum_{j=0}^k a_j b_{k-j}.
...
-    mul!(c, a, b) --> nothing
+@inline function _check_homogeneous_product_order(c::HomogeneousPolynomial,
+        a::HomogeneousPolynomial, b::HomogeneousPolynomial)
+    get_order(c) == get_order(a) + get_order(b) ||
+        throw(DimensionMismatch("result homogeneous degree must equal the sum of input degrees"))
+    return nothing
+end
 
-Accumulates in `c` the result of `a*b` with minimum allocation. Arguments
-c, a and b are `HomogeneousPolynomial`.
```
lbenet, line 773: Can we keep the docstrings?

src/arithmetic.jl
```diff
@@ -766,39 +830,222 @@ c_k = \sum_{j=0}^k a_j b_{k-j}.
...
-    order_c = get_order(c)+1
-    @inbounds num_coeffs_a = size_table[order_a]
-    @inbounds num_coeffs_b = size_table[order_b]
-    @inbounds posTb = pos_table[order_c]
-    @inbounds indTa = index_table[order_a]
-    @inbounds indTb = index_table[order_b]
+    degree_a = get_order(a)
+    degree_b = get_order(b)
+    degree_a == 0 && return _muladd_scalar_unchecked!(c, a[1], b)
+    degree_b == 0 && return _muladd_scalar_unchecked!(c, b[1], a)
```
lbenet, line 860: I think these lines *the facto* make that the first entry of `mul_table` is *never* updated nor used. If this is the case, there is no need to store that entry (in `mul_table`. I think that the code would be simplified, since the order would coincide then with the vector entry (no need for an entry corresponding to order 0).

src/arithmetic.jl
```diff
@@ -766,39 +830,222 @@ c_k = \sum_{j=0}^k a_j b_{k-j}.
...
+            na = pair ÷ num_right + 1
+            nb = pair - (na-1) * num_right + 1
+            acc += a_coeffs[na] * b_coeffs[nb]
+        end
+        c_coeffs[pos] = acc
+    end
+    return nothing
+end
+
+@inline function _mul_unchecked!(c::HomogeneousPolynomial{Float64},
```
lbenet: I think this function is coded above (line 851); we should eliminate this one.

src/arithmetic.jl
```diff
@@ -766,39 +830,222 @@ c_k = \sum_{j=0}^k a_j b_{k-j}.
...
+                pos = input_positions[pair]
+                c[pos] += sca * cb
+            end
+            pair += 1
         end
     end
     return nothing
 end
 
+@inline function _mul_scalar_unchecked!(c::HomogeneousPolynomial{Float64},
```
lbenet: I think this function is coded in line 945; we should remove this one.

src/arithmetic.jl
```diff
@@ -848,6 +1077,9 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
         @inbounds aux = a.coeffs[1] / b
         v = FixedSizeVectorDefault{typeof(aux)}(undef, length(a.coeffs))
         v .= a.coeffs ./ b
+        if $T == HomogeneousPolynomial
+            return HomogeneousPolynomial(a.space, v, get_order(a))
+        end
```
lbenet: I think this `if` is not needed (since the next line would work). Yet, if the need is related to some specificity of the `TaylorNSpace`, then I think a similar line is needed for `TaylorN`. In that case, I suggest:
```suggestion
        if $T == HomogeneousPolynomial || $T == TaylorN
            return $T(a.space, v, get_order(a))
        end
```

src/arithmetic.jl
```diff
@@ -868,6 +1100,9 @@ for T in (:HomogeneousPolynomial, :TaylorN)
         R = typeof(aux)
         coeffs = FixedSizeVectorDefault{R}(undef, length(b.coeffs))
         coeffs .= b.coeffs ./ a
+        if $T == HomogeneousPolynomial
+            return HomogeneousPolynomial(b.space, coeffs, get_order(b))
+        end
         return $T(coeffs, get_order(b))
```
lbenet:
```suggestion
        return $T(b.space, coeffs, get_order(b))
```

src/arithmetic.jl
```diff
@@ -876,6 +1111,9 @@ for T in (:HomogeneousPolynomial, :TaylorN)
...
```
lbenet:
```suggestion
        return $T(b.space, coeffs, get_order(b))
```

lbenet (commented):
I left various comments for src/arithmetic.jl

src/other_functions.jl
```diff
@@ -10,7 +10,12 @@ for T in (:Taylor1, :HomogeneousPolynomial, :TaylorN)
 
     ## real, imag, conj and ctranspose ##
     for f in (:real, :imag, :conj)
-        @eval ($f)(a::$T) = $T($f(a.coeffs), get_order(a))
+        @eval function ($f)(a::$T)
+            if $T == HomogeneousPolynomial
+                return HomogeneousPolynomial(a.space, ($f)(a.coeffs), get_order(a))
+            end
```
lbenet: Same as [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/410/changes#r3268295594)

src/parameters.jl
PerezHz: Thanks for both comments! I had added `variables` as a way to refer to the new behavior using `TaylorNSpace`s, while leaving `get_variables` untouched so that this PR is non-breaking. I now changed this so that all methods use `get_variables`, and we can think at some point in the future if we want to change the name to `getvariables` or something else. My style preference (which I sometimes see in other Julia packages, but I don't think is anything close to a standard) on this would be for "getters" to just use the name of the field, so `variables(space)` is alternative syntax for `space.variables`, and same for other fields.

About broadcasting yes, thanks, I brought broadcasting back!

src/hash_tables.jl
PerezHz: I added comments around this function and simplified a bit the workflow here, in the end the `else` clause below was not needed. The idea was that at module initialization the `else` branch would be chosen to initialize `default_space`, but that's already being done with the default values for the global lookup tables, so it's not needed.

src/hash_tables.jl
PerezHz: See [reply above](https://github.com/JuliaDiff/TaylorSeries.jl/pull/410#discussion_r3270070221)

src/hash_tables.jl
lbenet: My comment is about the last line: `dst` exposes the modifications, but does nothing more than that. Or does it?

src/parameters.jl
lbenet: If you prefer `variables` is fine with me 😄

src/zero_one.jl
```diff
@@ -37,13 +37,13 @@ zeros(::Type{HomogeneousPolynomial{T}}, order::Int) where {T<:Number} =
...
-    HomogeneousPolynomial(one.(a.coeffs), get_order(a))
+    HomogeneousPolynomial(a.space, one.(a.coeffs), get_order(a))
 
 function ones(a::HomogeneousPolynomial{T}, order::Int) where {T<:Number}
-    order == 0 && return [HomogeneousPolynomial([one(a[1])], 0)]
+    order == 0 && return [HomogeneousPolynomial(a.space, [one(a[1])], 0)]
     v = FixedSizeVectorDefault{HomogeneousPolynomial{T}}(undef, order+1)
     for ord in eachindex(v)
-        v[ord] = HomogeneousPolynomial(ones(T, size_table[ord]), ord-1)
+        v[ord] = HomogeneousPolynomial(a.space, ones(T, a.space.size_table[ord]), ord-1)
```
lbenet, line 46: In line 31 above, you use `Ref(a.space)` but here you use `a.space`. Is there any difference?

src/constructors.jl
PerezHz: Yes, I agree in principle we could remove the `space` field here from `TaylorN` and just read it off from one of the coefficients of the homogeneous polynomials, though the saving would be one pointer per `TaylorN`, not a copy of the space object, so the memory improvement is small (typically 8 bytes per `TaylorN` object). I kept it initially for direct access, but I agree it is cleaner to avoid redundant state. If we do this, we should update internal code to use the `space(a)` accessor rather than direct `a.space`.

src/auxiliary.jl
PerezHz: Yes, this is expected, this just feels safer because user code for two spaces like `space1` and `space3` in your example above can have very different purposes and meanings semantically, and having them this way helps them work independently. But I agree in principle we could design this so that two spaces of the same shape are regarded equivalent to each other.

src/auxiliary.jl
PerezHz: These lines are precisely the method body of  `_check_same_space(a, b, c)`, so I think as it is now it is correct? (otherwise a never-ending recursion would happen)

src/arithmetic.jl
PerezHz: Thanks, yes, I should probably listen to my own advice 😄

PerezHz: I removed the occurrences I could find of these `if`-patterns in the spirit of #362

PerezHz: And also added the `_check_same_space` methods you suggested for `Taylor1`

src/arithmetic.jl
PerezHz: Thanks, done!

src/arithmetic.jl
PerezHz: Done 👍

src/arithmetic.jl
PerezHz: thanks, removed

src/arithmetic.jl
PerezHz: These docstrings are attached to the `mul!` methods for `HomogeneousPolynomial`, around line 1040, should I move them here?

src/arithmetic.jl
PerezHz: Yes, you're right, thanks!

src/arithmetic.jl
PerezHz: I added `_muladd_unchecked!` for `Taylor1` as suggested, thanks! About @taylorize, these `_muladd_unchecked!` methods are meant as TS internals, since jet transport uses `mul!` over `TaylorN` arguments I think in principle this does not impose a change there? Or are you suggesting that @taylorize can use fused `muladd!` for arithmetic operations like `a = a + b`?

src/arithmetic.jl
PerezHz: I did some rework of the code using this suggestion, many thanks!

src/auxiliary.jl
lbenet: Let's leave this as it is now; indeed, if the user defines two space, even they are identical, it is better to keep them apart.

src/auxiliary.jl
lbenet: Yes, you are right!

src/arithmetic.jl
lbenet: No, I was not suggesting that we could use `muladd!`. I was wondering simply if something in src/dictmutfunct.jl would need a fix so `@taylorize` still works, and from your answer we expect that everything works fine.

src/arithmetic.jl
lbenet: Fine with me as it is now!

src/arithmetic.jl
PerezHz: thanks, removed

src/constructors.jl
lbenet: I agree that having `space` is better. The 8 bytes saving per TaylorN object are worth, if you think in some of the jet transport integrations with high order in `t`. As a side remark, I noticed that the performance of `fateman.jl` is greatly improved!

src/arithmetic.jl
PerezHz: thanks, removed

src/TaylorSeries.jl
```diff
@@ -45,7 +45,8 @@ import Base: zero, one, zeros, ones, isinf, isnan, iszero, isless,
 
 import Base.float
 
-export Taylor1, TaylorN, HomogeneousPolynomial, AbstractSeries, TS
+export Taylor1, TaylorN, HomogeneousPolynomial, AbstractSeries, TS,
+    TaylorNSpace, JetSpace
```
lbenet: I think we should opt for one, either `JetSpace` or `TaylorNSpace` (or if you prefer `JetNSpace`); I am inclined for either *Jet* form since it also applies to `HomogeneousPolynomial`s (and not exclusively to `TaylorN`s).

src/functions.jl
```diff
@@ -13,7 +13,7 @@ for T in (:Taylor1, :TaylorN)
         function exp(a::$T)
             aux = exp(constant_term(a))
             aa = convert($T{typeof(aux)}, a)
-            c = $T(aux, get_order(a))
+            c = _constant_series_like(a, aux, get_order(a))
```
lbenet: I see the advantage of `_constant_series_like`. Yet, I think this may impose some changes in `@taylorize` internals, because otherwise we would use the `default_space`. A note to myself too...

src/parameters.jl
lbenet: Resolving this, because I think my last suggestion (having `Matrix` for `mul_table`) may break the lazy construction of the tables; feel free to unresolve it if you think the suggestion has any future

src/arithmetic.jl
PerezHz: I fixed this by taking `Taylor1` out of this code generation loop, and since the syntax is shared by `HomogeneousPolynomial` and `TaylorN` then the `if` can be removed from the method body.

PerezHz: (there was another case of this; also fixed it)

src/arithmetic.jl
PerezHz: Thanks, removed

src/other_functions.jl
PerezHz: Thanks, removed

src/power.jl
```diff
@@ -626,23 +626,27 @@ Returns `c += a*a` with no allocation; all parameters are `HomogeneousPolynomial
...
+    _check_same_space(c, a)
     iszero(a) && return nothing
 
-    @inbounds num_coeffs_a = size_table[get_order(a)+1]
-    @inbounds posTb = pos_table[get_order(c)+1]
-    @inbounds idxTb = index_table[get_order(a)+1]
+    sp = c.space
+    degree_a = get_order(a)
+    get_order(c) == 2 * degree_a ||
+        throw(DimensionMismatch("result homogeneous degree must be twice the input degree"))
```
lbenet: I don't think this is needed (seemingly it was not in the old version), but perhaps I oversee something. Can you explain?

src/hash_tables.jl
PerezHz: the line with `dst` was there so that it would be assigned to `active_space` (now the `if`/`else` block has been removed, so the variable `dst`  is not there anymore)

PerezHz: Now we only have the branch which was within the `if` before, and the assignment is simply `active_space = default_space[]`

src/arithmetic.jl
PerezHz: done

src/power.jl
PerezHz: You're right that this check is not strictly needed, since this method is assumed to be called with arguments which already pass `c` with degree `2*get_order(a)`. I added it because the new product-table path obtains output positions from `_product_table(space, degree_a, degree_a)`, whose positions are for degree `2*degree_a` and no longer depend on `get_order(c)`. In the previous code, `pos_table[get_order(c)+1]` implicitly tied the lookup to the degree of `c`. This explicit check was just to try to prevent silent misuse of `accsqr!` with a result polynomial of the wrong homogeneous degree; a similar check in `mul!` uses `_check_homogeneous_product_order(c, a, b)`. I have changed this for `_check_homogeneous_product_order(c, a, a)`, which is essentially the same as the current check, but we can also change it to something else.

src/TaylorSeries.jl
PerezHz: I went ahead and picked `JetSpace`, which is shorter to type and I thought of leaving out the `N`, since it can give the idea that it's somehow only related to `TaylorN`.

src/power.jl
lbenet: Thanks for the explanation! I think it is ok as it is now!

src/zero_one.jl
PerezHz: Yes, in line 31 `Ref(a.space)` is used because that line uses broadcasting and `Ref(a.space)` broadcasting rules treat `a.space` as a "scalar" during broadcasting. In line 46 there is no broadcasting, so no `Ref` is needed here.

lbenet: I see, thanks for the clarification.

src/functions.jl
PerezHz: Yes, `@taylorize` will use the `default_space` so if we want `@taylorize` to use a non-default space we would eventually need to make that consistent across all constructor call sites in `@taylorize`-generated methods, since those somehow break-up what these methods in `functions.jl` do in 0-th order and higher orders

src/parameters.jl
PerezHz: About `UInt32`: I chose this mainly as a compact storage choice: it uses 4 bytes per pair instead of 8 bytes per Int on 64-bit Julia, and it still allow us to store `typemax(UInt32)` (~ 4 billion) pairs

src/parameters.jl
PerezHz: Thanks for the suggestion about having `mul_table` as a `Matrix`. I think it could work in principle even with lazy construction of the tables, but at the same time, this field is only used one in homogeneous polynomial product methods, and it uses `mul_table[index_a][index_b]` just to get the corresponding product table before starting the "hot" product loop see e.g. [here](https://github.com/PerezHz/TaylorSeries.jl/blob/68f748a5a3449f677900552bb56d667d853c02ff/src/arithmetic.jl#L867), so I don't think it has considerable performance implications, but could be worth trying

lbenet:
I checked (locally) and there are four tests of `TaylorIntegration.jl` that do not pass (using this branch of TS). I will dig deeper into what's going on in the next few days, but these errors may point out some subtle error not catched by the tests that we currently have here. I don't think this is directly related to [this comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/410#discussion_r3282803962) about some updates required for `@taylorize`, which are not difficult and can be done later.

src/parameters.jl
PerezHz: I went ahead and changed `get_variables` for `variables`, and added deprecation warnings

lbenet:
As for this PR I think there are still three remaining things to be done, which are not really indendent. 
- First, decide if we release a new minor version or only a patch version. 
- Second, decide if we retain the compatibility with the old global tables, or if we simply get rid of them.
- Finally, update the [userguide](https://github.com/JuliaDiff/TaylorSeries.jl/blob/master/docs/src/userguide.md) in the docs, since there is no reference to the new scheme to define `JetSpace`s. 

In my opinion, we should release a minor version, getting rid of the old global tables but retaining a default `JetSpace` as a global constant (which could be modified through e.g. `set_variables` or exploiting the mutability of `JetSpace`). (Other minor addition would be a getter function `space` or `jetspace`.) Proceeding in this way, packages that rely on TS would need to explicitly update their compat entry, and thus be forced to migrate to the use of `JetSpace`s, or otherwise simply don't change the compat entry. The changes in the docs would reflect the version that they apply to.

Let me know what's your opinion.

src/arithmetic.jl
```diff
@@ -136,7 +143,27 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
...
+                    c[0][1] = ($f)(aa[0][1], bb)
+                    return c
+                end
+
+                function ($f)(b::S, a::$T{T}) where {T<:Number, S<:NumberNotSeries}
+                    R = promote_type(T, S)
+                    aa = convert(TaylorN{R}, a)
+                    bb = convert(R, b)
+                    c = TaylorN(($f)(aa.coeffs[:]), get_order(aa))
+                    c[0][1] = bb + aa[0][1]
```
lbenet: Found it! This line causes that four tests fail in TaylorIntegration!

```suggestion
                    c[0][1] = ($f)(bb, aa[0][1])
```

src/functions.jl
```diff
@@ -1309,7 +1309,7 @@ function expm1!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         return nothing
     end
     # The recursion formula
-    tmp = TaylorN( zero(a[k][0][1]), get_order(a[0]))
+    tmp = TaylorN(a[0].space, zero(a[k][0][1]), get_order(a[0]))
```
lbenet: In the same spirit as above:

```suggestion
    tmp = _constant_series_like(a[k], zero(a[k][0][1]), get_order(a[0]))
```

src/functions.jl
```diff
@@ -1345,7 +1345,7 @@ function log!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
...
```
lbenet:
```suggestion
    tmp = _constant_series_like(a[k], zero(a[k][0][1]), get_order(a[0]))
```

src/functions.jl
```diff
@@ -1371,7 +1371,7 @@ function log1p!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         end
         return nothing
     end
-    tmp1 = TaylorN( zero(a[k][0][1]), get_order(a[0]))
+    tmp1 = TaylorN(a[0].space, zero(a[k][0][1]), get_order(a[0]))
```
lbenet:
```suggestion
    tmp1 = _constant_series_like(a[k], zero(a[k][0][1]), get_order(a[0]))
```

src/functions.jl
```diff
@@ -1385,7 +1385,7 @@ function log1p!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         return nothing
     end
     # The recursion formula
-    tmp = TaylorN( zero(a[k][0][1]), get_order(a[0]))
+    tmp = TaylorN(a[0].space, zero(a[k][0][1]), get_order(a[0]))
```
lbenet:
```suggestion
    tmp = _constant_series_like(a[k], zero(a[k][0][1]), get_order(a[0]))
```

src/functions.jl
```diff
@@ -1530,7 +1530,7 @@ end
 
 function atan!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         r::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
-    aux = TaylorN( zero(a[0][0][1]), get_order(a[0]) )
+    aux = TaylorN(a[0].space, zero(a[0][0][1]), get_order(a[0]) )
```
lbenet:
```suggestion
    aux = _constant_series_like(a[0], zero(a[0][0][1]), get_order(a[0]))
```

src/functions.jl
```diff
@@ -1582,7 +1582,7 @@ end
 
 function tanh!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         res2::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
-    aux = TaylorN( zero(a[0][0][1]), get_order(a[0]))
+    aux = TaylorN(a[0].space, zero(a[0][0][1]), get_order(a[0]))
```
lbenet:
```suggestion
    aux = _constant_series_like(a[0], zero(a[0][0][1]), get_order(a[0]))
```

src/functions.jl
```diff
@@ -1609,7 +1609,7 @@ end
 
 function asinh!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         r::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
-    tmp = TaylorN( zero(a[0][0][1]), get_order(a[0]))
+    tmp = TaylorN(a[0].space, zero(a[0][0][1]), get_order(a[0]))
```
lbenet:
```suggestion
    tmp = _constant_series_like(a[0], zero(a[0][0][1]), get_order(a[0]))
```

src/functions.jl
```diff
@@ -1656,7 +1656,7 @@ end
 
 function acosh!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         r::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
-    tmp = TaylorN( zero(a[0][0][1]), get_order(a[0]))
+    tmp = TaylorN(a[0].space, zero(a[0][0][1]), get_order(a[0]))
```
lbenet:
```suggestion
    tmp = _constant_series_like(a[0], zero(a[0][0][1]), get_order(a[0]))
```

src/functions.jl
```diff
@@ -1704,7 +1704,7 @@ end
 
 function atanh!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         r::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
-    tmp = TaylorN( zero(a[0][0][1]), get_order(a[0]) )
+    tmp = TaylorN(a[0].space, zero(a[0][0][1]), get_order(a[0]) )
```
lbenet:
```suggestion
    tmp = _constant_series_like(a[0], zero(a[0][0][1]), get_order(a[0]))
```

lbenet (commented):
I have few more suggestions for `src/functions.jl` (and one important change in `src/arithmetic.jl`)

src/arithmetic.jl
PerezHz: Many thanks for looking into this!

src/functions.jl
PerezHz: Thank you for the suggestions, @lbenet! I see some of them have `_constant_series_like(a[k], ...)`, but should those be `_constant_series_like(a[0], ...)` instead? (with `0` index instead of `k`?)

PerezHz: Ah, I see `a[0]` is only used to get its `space` field, and it should work anyway, I'll commit (and maybe will just set the index k->0 for consistency with other functions)

lbenet: > Thank you for the suggestions, @lbenet! I see some of them have `_constant_series_like(a[k], ...)`, but should those be `_constant_series_like(a[0], ...)` instead? (with `0` index instead of `k`?)

You are right, I noticed it and then edit it; forgot to mention it.

PerezHz:
@lbenet I have updated the user guide to reflect the usage of `JetSpace` to reflect the current state of things, but we can update it again if needed

PerezHz:
@lbenet: about the three points you mentioned above, I agree we should eventually release this as 0.22 to signal the new `JetSpace` machinery, and downstream packages with compat bounds on 0.21 will not pick it up automatically. I was thinking of leaving the "compatibility layer" (i.e., the existing lookup tables, etc.) there so that migration can be easier for users, but that also implies cleaning this up in another release. And since releasing 0.22 would be considered breaking anyway, we might as well do all at once. So if you agree I can try to clean up the code related to the existing lookup tables and switch entirely to the `JetSpace` "backend", and bump a minor version.

Update: we will do this in a two-stage release, as described in [here](https://github.com/JuliaDiff/TaylorSeries.jl/pull/410#issuecomment-4522459125)

PerezHz:
Update: after an offline discussion, we are considering doing this in a two-part release: first, a full backwards-compatible patch release v0.21.10 with this PR essentially as it is now, and then do fully migrate to the new `JetSpace` backend in a minor release v0.22.0.

PerezHz:
I ran tests on CI for TaylorIntegration [here](https://github.com/PerezHz/TaylorIntegration.jl/actions/runs/26308917143/job/77455191733) and for NEOs [here](https://github.com/PerezHz/NEOs.jl/actions/runs/26311262220) and both are passing, though I found for NEOs.jl had to re-run a couple of tests, and then it worked... @LuEdRaMo have you noticed whether this has happened on recent test runs on NEOs.jl?

PerezHz:
Also opened an issue about the JLD2 extension, since it will have to be updated during the minor version release to support non-default `JetSpace`s (not needed at this point for the patch version release).

LuEdRaMo:
> I ran tests on CI for TaylorIntegration [here](https://github.com/PerezHz/TaylorIntegration.jl/actions/runs/26308917143/job/77455191733) and for NEOs [here](https://github.com/PerezHz/NEOs.jl/actions/runs/26311262220) and both are passing, though I found for NEOs.jl had to re-run a couple of tests, and then it worked... @LuEdRaMo have you noticed whether this has happened on recent test runs on NEOs.jl?

If you had to rerun the tests on Mac because of an error with an unknown variable `ind` then that is a recurrent problem that NEOs' tests have on Github that I have not quite worked out. If the error is different, then it might be related to this PR...

PerezHz:
Yeah there was an error in macos, not sure though if I saw the indexing issue. Since tests in NEOs.jl main are failing I'm basing tests off of https://github.com/PerezHz/NEOs.jl/pull/198, so maybe a combined effect?

PerezHz:
Anyway I'll rerun tests several times to see if the error is reproducible

LuEdRaMo:
> Yeah there was an error in macos, not sure though if I saw the indexing issue. Since tests in NEOs.jl main are failing I'm basing tests off of [PerezHz/NEOs.jl#198](https://github.com/PerezHz/NEOs.jl/pull/198), so maybe a combined effect?

Indeed https://github.com/PerezHz/NEOs.jl/pull/198 fixes an error that just came up today regarding a hardcoded observatories tests. Otherwise, any change in the bounds used for the orbit determination tests can be a combined effect of this PR and https://github.com/PerezHz/NEOs.jl/pull/198.

lbenet:
> I ran tests on CI for TaylorIntegration [here](https://github.com/PerezHz/TaylorIntegration.jl/actions/runs/26308917143/job/77455191733) and for NEOs [here](https://github.com/PerezHz/NEOs.jl/actions/runs/26311262220) and both are passing, though I found for NEOs.jl had to re-run a couple of tests, and then it worked... @LuEdRaMo have you noticed whether this has happened on recent test runs on NEOs.jl?

I ran (locally) TaylorIntegration.jl as well as TaylorModels.jl tests using this branch, and everything is working fine. Regarding NEOs.jl, there are some broken tests in `astrometry.jl` (related to Mt Lemmon designation, which may have changed), so unrelated to this PR; there are also some broken tests in the compatibility with JPL section, which is difficult to tell if they are related or not with this PR. (My tests didn't use https://github.com/PerezHz/NEOs.jl/pull/198 so there may be an influence of this PR on those tests.) I'll report on PlanetaryEphemeris.jl later...

Everything was done using one thread; it may be important to test using more threads...

PerezHz:
I also ran TaylorIntegration.jl tests locally on 4 threads and passed fine; I also ran NEOs.jl tests on 4 threads locally and only hit the issue with JPL compatibility tests. I understand NEOs.jl CI config uses 2 threads in all jobs (see [here](https://github.com/PerezHz/NEOs.jl/blob/8eb1d678bcc6b2bd46e924f547670cd6a163e6bd/.github/workflows/CI.yml#L47)); I'm rerunning NEOs.jl tests on CI with a commit in main from ~1 week ago which ran fine then, but let's see...

PerezHz:
Running PlanetaryEphemeris.jl tests on CI using this PR [here](https://github.com/PerezHz/PlanetaryEphemeris.jl/actions/runs/26319211489)

PerezHz:
Update: NEOs.jl show some variation in MacOs due to the issue described by @LuEdRaMo, and also on windows: sometimes tests pass, and sometimes JPL comparisons don't work. I'm rerunning in a single thread to rule out multi-threading issues

ext/TaylorSeriesIAExt.jl
```diff
@@ -41,15 +41,15 @@ for I in (:Interval, :ComplexI)
         TS._isthinzero(x::$I{T}) where {T<:Real} = isthinzero(x)
 
         function ==(a::Taylor1{$I{T}}, b::Taylor1{$I{S}}) where {T<:NumTypes, S<:NumTypes}
-            if get_order(a) != get_order(b)
+            if TS.order(a) != TS.order(b)
```
lbenet: Note that, since `order` is exported, `TS` is not needed. (It's ok if you leave it as it is now.)

lbenet: I see (in `src/arithmetic.jl`) that there are many uses of `order` as variable name; if that is the reason, perhaps `getorder` is better...

ext/TaylorSeriesJLD2Ext.jl
lbenet, line 1: What about adding to `TaylorNSerialization` struct a field `space`, which precisely stores the space associated to what will be serialized?

src/deprecated.jl
```diff
@@ -0,0 +1,10 @@
+# This file is part of the TaylorSeries.jl Julia package, MIT license
+#
+# Luis Benet & David P. Sanders
+# UNAM
+#
+# MIT Expat license
+#
+
+Base.@deprecate set_variables(args...; kwargs...) variables!(args...; kwargs...)
+Base.@deprecate get_variables(args...; kwargs...) variables(args...; kwargs...)
```
lbenet, line 10: Could you also add `get_order`?

PerezHz: I actually had added it but that breaks TaylorIntegration tests since logs are captured in some tests, so removed it to be able to test this PR but I agree we should add those

PerezHz:
PlanetaryEphemeris.jl tests pass using this PR

ext/TaylorSeriesJLD2Ext.jl
PerezHz: I went for adding a new serialization struct `TaylorNSerializationV2` in the JLD2 extension so that `TaylorN` saved from the same julia session will preserve their `JetSpace`-ity (variables belonging to same space will continue to operate between them, and those which are from different spaces or different julia sessions will not). Also the info about the `space` of a `TaylorN` is saved without saving the whole tables, so that the space can be reconstructed. I kept the current `TaylorNSerialization` for backwards compatibility; `TaylorN` saved with the serialization will be able to load as they do now (or at least that's the intention).

PerezHz:
> Also opened an issue about the JLD2 extension, since it will have to be updated during the minor version release to support non-default `JetSpace`s (not needed at this point for the patch version release).

About this, I think I found a way to add this now via a `TaylorNSerializationV2` and a `JetSpaceSerialization`, while keeping the existing `TaylorNSerialization` so this is actually now done.

ext/TaylorSeriesJLD2Ext.jl
lbenet: "JetSpacity" sounds pretty "Space Oddity" 😄

lbenet: > Also the info about the `space` of a `TaylorN` is saved without saving the whole tables, so that the space can be reconstructed. I kept the current `TaylorNSerialization` for backwards compatibility; `TaylorN` saved with the serialization will be able to load as they do now (or at least that's the intention).

The question then is if we can somehow also save somehow the generated tables? As a dummy example, I think in a test like Fateman's: you produce the result, save it (together with the tables), so when you load them back, the spped up produced by knowing the tables is ready.

src/deprecated.jl
lbenet: I suggest to make an appropriate PR in TaylorIntegration; that failure is *mild*

lbenet:
> About this, I think I found a way to add this now via a `TaylorNSerializationV2` and a `JetSpaceSerialization`, while keeping the existing `TaylorNSerialization` so this is actually now done.

Maybe the [my comment](https://github.com/JuliaDiff/TaylorSeries.jl/pull/410#discussion_r3293003066) is alreaady outdated...

PerezHz:
I did a small fix in NEOs.jl current main, v0.26.3, but now CI is passing, both using current TS release and this PR:

[NEOs.jl CI run, TS v0.21.9](https://github.com/PerezHz/NEOs.jl/actions/runs/26336860045)
[NEOs.jl CI run, this PR](https://github.com/PerezHz/NEOs.jl/actions/runs/26337227275)

As a side note, I see intermittent issues sometimes in windows, sometimes in macos, but those appear to be platform-related and also appear using TS v0.21.9, so those issues are not related to this PR. Also, CI times are highly variable, but at least on the last couple of runs I've seen tests on NEOs.jl taking:

- ~14 minutes on julia 1.10, TS v0.21.9
- ~10 minutes on julia 1.10, TS this PR
- ~18 minutes on julia 1.12, TS v0.21.9
- ~14 minutes on julia 1.12, TS this PR

ext/TaylorSeriesJLD2Ext.jl
PerezHz: We could, but it's a tradeoff of speed of initialization vs use of memory: if we save the whole tables serialized TaylorN will take more disk space but are quicker to load, while if we only keep the minimal info needed for reconstructing then they take less space, but will take more time to load.

PerezHz: I would suggest to go with how it is now, and if we find loading speed matters more, then we can switch the approach

PerezHz:
As things stand now, this PR already fixes #412 (putting this comment here so that the issue will be closed when this PR is merged).

src/deprecated.jl
PerezHz: Yeah, I'm looking into this

ext/TaylorSeriesIAExt.jl
PerezHz: Looking into this

ext/TaylorSeriesJLD2Ext.jl
lbenet: Fine with me!

ext/TaylorSeriesIAExt.jl
PerezHz: Indeed, if inside a method a binding `order` exists, that will shadow the `order(...)` method. I did a pass through the code looking for places where the prefix was not needed and removed those. Even so, I think there's value in keeping `order(...)` as a method

src/deprecated.jl
PerezHz: Opened PR https://github.com/PerezHz/TaylorIntegration.jl/pull/232

PerezHz: And also added deprecations for `get_order` here in TS

lbenet: Thanks a lot!

LuEdRaMo:
> I did a small fix in NEOs.jl current main, v0.26.3, but now CI is passing, both using current TS release and this PR:
> 
> [NEOs.jl CI run, TS v0.21.9](https://github.com/PerezHz/NEOs.jl/actions/runs/26336860045) [NEOs.jl CI run, this PR](https://github.com/PerezHz/NEOs.jl/actions/runs/26337227275)
> 
> As a side note, I see intermittent issues sometimes in windows, sometimes in macos, but those appear to be platform-related and also appear using TS v0.21.9, so those issues are not related to this PR. Also, CI times are highly variable, but at least on the last couple of runs I've seen tests on NEOs.jl taking:
> 
> * ~14 minutes on julia 1.10, TS v0.21.9
> * ~10 minutes on julia 1.10, TS this PR
> * ~18 minutes on julia 1.12, TS v0.21.9
> * ~14 minutes on julia 1.12, TS this PR

Thanks for looking into this! Indeed, independently of this PR, NEOs' tests are very unstable on Github. I'll try to fix that sometime in the (not so far) future. 

The above means that, at least on that front, this PR is good to go!

LuEdRaMo:
More encouraging results:

```Julia
# master

BenchmarkTools.Trial: 10 samples with 1 evaluation per sample.
 Range (min … max):  31.897 s … 34.769 s  ┊ GC (min … max): 1.06% … 0.89%
 Time  (median):     32.175 s             ┊ GC (median):    1.01%
 Time  (mean ± σ):   32.630 s ±  1.085 s  ┊ GC (mean ± σ):  1.05% ± 0.12%

  █ █ ▁ ▁ ▁  ▁                                        ▁   ▁  
  █▁█▁█▁█▁█▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁█ ▁
  31.9 s         Histogram: frequency by time        34.8 s <

 Memory estimate: 1.40 GiB, allocs estimate: 22963651.

# This PR

BenchmarkTools.Trial: 10 samples with 1 evaluation per sample.
 Range (min … max):  18.374 s …   19.905 s  ┊ GC (min … max): 1.94% … 1.67%
 Time  (median):     19.410 s               ┊ GC (median):    1.77%
 Time  (mean ± σ):   19.314 s ± 414.306 ms  ┊ GC (mean ± σ):  1.86% ± 0.22%

  █                   █           █     ████ █  █           █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁████▁█▁▁█▁▁▁▁▁▁▁▁▁▁▁█ ▁
  18.4 s          Histogram: frequency by time         19.9 s <

 Memory estimate: 1.53 GiB, allocs estimate: 23830731.
```
The above correspond to a 25-year integration of Apophis' orbit using second-order jet transport and five threads. 

In summary, a 40% decrease in computation time in exchange for an increase of only 9% in memory and 3% in allocations.

PerezHz:
Many thanks for looking into this @LuEdRaMo I agree these results are really encouraging!

ext/TaylorSeriesIAExt.jl
PerezHz: So what I did was removed the `TS.` prefix from calls to `order(...)` where a variable `order` did not exist in the same function body which could shadow the `order(...)` method. Otherwise, I left the prefix. I tried to address all of the instances of this pattern that appeared in the code.

lbenet:
I think it's better to merge https://github.com/PerezHz/TaylorIntegration.jl/pull/231 and then this PR, so things go smoother

lbenet (approved):

lbenet:
Merging! Thanks a lot @PerezHz, this is a great addition!

Issue #411, opened 2026-05-22
New in-place functions


LuEdRaMo:
I noticed there is no method of `TaylorSeries.div!` to divide a `TaylorN` by a `NumberNotScalar`, which would be an equivalent of the following:
https://github.com/JuliaDiff/TaylorSeries.jl/blob/c2698dbaf93616dc25403de05a85f483b61dd853/src/arithmetic.jl#L1049-L1053
but for a `TaylorN`. 

In addition, as discussed online, it would be helpful to have an in-place function to update only the constant term of an `AbstractSeries`, something like:
```Julia
constant_term!(c::Taylor1{T}, a::T) where {T <: NumberNotSeries} = c[0] = a
```
with analogous methods for `Taylor1{Number}`, `TaylorN`, `HomogeneousPolynomial`, etc.

PR #415, opened 2026-05-29
Improve Taylor1 performance

PerezHz:
This PR adds some improvements to `Taylor1{Float64}` performance:

- `src/arithmetic.jl`: add direct coefficient-vector paths for scalar `add!`, `subst!`, `mul!`, `muladd!`, `mul_scalar!`, plus scalar `+` and `-`. Turns out there is a small overhead by accessing `a.coeffs` lots of times in the innermost loops, which can build up. Also these methods now use a local accumulator for each coefficient, then store once; this avoids repeated `c[k]` loads/stores inside the order traversing loop.
- `src/power.jl`: cached scalar `Taylor1` power metadata (`findfirst(a)`, `findlast(a)`, `a[l0]`, and `order(a)`) instead of recomputing per coefficient. Also, switch to direct coefficient access in `pow!`, and faster scalar `sqr!`.
- `src/identity.jl`: clean up a small inference in `identity!`, which was returning an assigned scalar in one method instead of `nothing`.
- `test/mutatingfuncts.jl`: added regression checks for full-series scalar mutating methods.
- Add a `nowarn` kwarg in `variables!`, so that tests are free of warnings.

Some quick, local, preliminary benchmarks on order 30 /100 `Taylor1{Float64}`:

mul!: ~205 ns / 2419 ns before, now ~85 ns / 1105 ns
a*b: ~211 ns / 2481 ns before, now ~103 ns / 1100 ns
pow! coefficient loop: ~438 ns / 3961 ns before, now ~321 ns / 2560 ns
a^1.5: ~474 ns / 3995 ns before, now ~298 ns / 2235 ns
new full-series add!: (there's no before), now ~3.5 ns / 9.3 ns, allocation-free

~TODO: add more systematic benchmarks~
EDIT: See below for systematic benchmarks

### DIFF
--- a/src/TaylorSeries.jl
+++ b/src/TaylorSeries.jl
@@ -56,7 +56,7 @@ export getcoeff, derivative, integrate, differentiate,
     get_variable_names, get_variable_symbols,
     # jacobian, hessian, jacobian!, hessian!,
     ∇, taylor_expand, update!,
-    constant_term, linear_polynomial, nonlinear_polynomial,
+    constant_term, constant_term!, linear_polynomial, nonlinear_polynomial,
     normalize_taylor, norm
 
 const TS = TaylorSeries
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -361,6 +361,50 @@ function subst!(v::Taylor1{Taylor1{T}}, a::Taylor1{Taylor1{T}},
     return nothing
 end
 
+function add!(v::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
+        b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries}
+    v_coeffs = v.coeffs
+    a_coeffs = a.coeffs
+    b_coeffs = b.coeffs
+    _check_same_space(v_coeffs[1], a_coeffs[1], b_coeffs[1])
+    @inbounds for i in eachindex(v_coeffs)
+        v_hps = v_coeffs[i].coeffs
+        a_hps = a_coeffs[i].coeffs
+        b_hps = b_coeffs[i].coeffs
+        for j in eachindex(v_hps)
+            v_hp = v_hps[j].coeffs
+            a_hp = a_hps[j].coeffs
+            b_hp = b_hps[j].coeffs
+            for k in eachindex(v_hp)
+                v_hp[k] = a_hp[k] + b_hp[k]
+            end
+        end
+    end
+    return nothing
+end
+
+function subst!(v::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
+        b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries}
+    v_coeffs = v.coeffs
+    a_coeffs = a.coeffs
+    b_coeffs = b.coeffs
+    _check_same_space(v_coeffs[1], a_coeffs[1], b_coeffs[1])
+    @inbounds for i in eachindex(v_coeffs)
+        v_hps = v_coeffs[i].coeffs
+        a_hps = a_coeffs[i].coeffs
+        b_hps = b_coeffs[i].coeffs
+        for j in eachindex(v_hps)
+            v_hp = v_hps[j].coeffs
+            a_hp = a_hps[j].coeffs
+            b_hp = b_hps[j].coeffs
+            for k in eachindex(v_hp)
+                v_hp[k] = a_hp[k] - b_hp[k]
+            end
+        end
+    end
+    return nothing
+end
+
 function +(a::Taylor1{T}, b::Taylor1{T}) where {T<:NumberNotSeries}
     if order(a) != order(b)
         a, b = fixorder(a, b)
@@ -409,45 +453,70 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
                 a, b = fixorder(a, b)
             end
             c = zero(a)
-            for k in eachindex(a)
-                ($fc)(c, a, b, k)
-            end
+            ($fc)(c, a, b)
             return c
         end
         function ($fc)(v::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
                 b::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
-            @inbounds for i in eachindex(v[k])
-                for j in eachindex(v[k][i])
-                    v[k][i][j] = ($f)(a[k][i][j], b[k][i][j])
+            v_k = v.coeffs[k+1]
+            a_k = a.coeffs[k+1]
+            b_k = b.coeffs[k+1]
+            v_hps = v_k.coeffs
+            a_hps = a_k.coeffs
+            b_hps = b_k.coeffs
+            @inbounds for i in eachindex(v_hps)
+                v_hp = v_hps[i].coeffs
+                a_hp = a_hps[i].coeffs
+                b_hp = b_hps[i].coeffs
+                for j in eachindex(v_hp)
+                    v_hp[j] = ($f)(a_hp[j], b_hp[j])
                 end
             end
             return nothing
         end
         function ($fc)(v::Taylor1{TaylorN{T}}, a::NumberNotSeries,
                 b::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
-            @inbounds for i in eachindex(v[k])
-                aaa = ifelse(k == 0 && i == 0, a, zero(a))
-                for j in eachindex(v[k][i])
-                    v[k][i][j] = ($f)(aaa, b[k][i][j])
+            v_k = v.coeffs[k+1]
+            b_k = b.coeffs[k+1]
+            v_hps = v_k.coeffs
+            b_hps = b_k.coeffs
+            @inbounds for i in eachindex(v_hps)
+                aaa = ifelse(k == 0 && i == 1, a, zero(a))
+                v_hp = v_hps[i].coeffs
+                b_hp = b_hps[i].coeffs
+                for j in eachindex(v_hp)
+                    v_hp[j] = ($f)(aaa, b_hp[j])
                 end
             end
             return nothing
         end
         function ($fc)(v::Taylor1{TaylorN{T}}, b::Taylor1{TaylorN{T}},
                 a::NumberNotSeries, k::Int) where {T<:NumberNotSeries}
-            @inbounds for i in eachindex(v[k])
-                aaa = ifelse(k == 0 && i == 0, a, zero(a))
-                for j in eachindex(v[k][i])
-                    v[k][i][j] = ($f)(b[k][i][j], aaa)
+            v_k = v.coeffs[k+1]
+            b_k = b.coeffs[k+1]
+            v_hps = v_k.coeffs
+            b_hps = b_k.coeffs
+            @inbounds for i in eachindex(v_hps)
+                aaa = ifelse(k == 0 && i == 1, a, zero(a))
+                v_hp = v_hps[i].coeffs
+                b_hp = b_hps[i].coeffs
+                for j in eachindex(v_hp)
+                    v_hp[j] = ($f)(b_hp[j], aaa)
                 end
             end
             return nothing
         end
         function ($fc)(v::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
                 k::Int) where {T<:NumberNotSeries}
-            @inbounds for l in eachindex(v[k])
-                for m in eachindex(v[k][l])
-                    v[k][l][m] = ($f)(a[k][l][m])
+            v_k = v.coeffs[k+1]
+            a_k = a.coeffs[k+1]
+            v_hps = v_k.coeffs
+            a_hps = a_k.coeffs
+            @inbounds for l in eachindex(v_hps)
+                v_hp = v_hps[l].coeffs
+                a_hp = a_hps[l].coeffs
+                for m in eachindex(v_hp)
+                    v_hp[m] = ($f)(a_hp[m])
                 end
             end
             return nothing
@@ -1010,9 +1079,15 @@ end
 
 function mul!(res::Taylor1{TaylorN{T}}, a::NumberNotSeries,
         b::Taylor1{TaylorN{T}}, k::Int) where {T<:NumberNotSeries}
-    for l in eachindex(b[k])
-        for m in eachindex(b[k][l])
-            res[k][l][m] = a*b[k][l][m]
+    res_k = res.coeffs[k+1]
+    b_k = b.coeffs[k+1]
+    res_hps = res_k.coeffs
+    b_hps = b_k.coeffs
+    @inbounds for l in eachindex(res_hps)
+        res_hp = res_hps[l].coeffs
+        b_hp = b_hps[l].coeffs
+        for m in eachindex(res_hp)
+            res_hp[m] = a*b_hp[m]
         end
     end
     return nothing
@@ -1878,9 +1953,15 @@ end
 
 @inline function div!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
         b::NumberNotSeries, k::Int) where {T<:NumberNotSeries}
-    for l in eachindex(a[k])
-        for m in eachindex(a[k][l])
-            res[k][l][m] = a[k][l][m]/b
+    res_k = res.coeffs[k+1]
+    a_k = a.coeffs[k+1]
+    res_hps = res_k.coeffs
+    a_hps = a_k.coeffs
+    @inbounds for l in eachindex(res_hps)
+        res_hp = res_hps[l].coeffs
+        a_hp = a_hps[l].coeffs
+        for m in eachindex(res_hp)
+            res_hp[m] = a_hp[m]/b
         end
     end
     return nothing
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -37,13 +37,13 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
 
             function ($f)(a::$T{T}, b::T) where {T<:Number}
                 c = $T(a.coeffs[:])
-                c[0] = $f(a[0], b)
+                constant_term!(c, $f(constant_term(a), b))
                 return c
             end
 
             function ($f)(b::T, a::$T{T}) where {T<:Number}
                 c = $T($f(a.coeffs[:]))
-                c[0] = $f(b, a[0])
+                constant_term!(c, $f(b, constant_term(a)))
                 return c
             end
 
@@ -199,7 +199,7 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
                     aa = convert(TaylorN{R}, a)
                     bb = convert(R, b)
                     c = TaylorN(aa.coeffs[:], order(aa))
-                    c[0][1] = ($f)(aa[0][1], bb)
+                    constant_term!(c, ($f)(constant_term(aa), bb))
                     return c
                 end
 
@@ -208,7 +208,7 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
                     aa = convert(TaylorN{R}, a)
                     bb = convert(R, b)
                     c = TaylorN(($f)(aa.coeffs[:]), order(aa))
-                    c[0][1] = ($f)(bb, aa[0][1])
+                    constant_term!(c, ($f)(bb, constant_term(aa)))
                     return c
                 end
 
@@ -285,8 +285,9 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
             coeffs = FixedSizeVectorDefault{HomogeneousPolynomial{Taylor1{R}}}(
                     undef, order(a)+1)
             coeffs .= a.coeffs
-            @inbounds coeffs[1] = aux
-            return TaylorN(coeffs, order(a))
+            c = TaylorN(coeffs, order(a))
+            constant_term!(c, aux)
+            return c
         end
 
         function ($f)(b::S, a::TaylorN{Taylor1{T}}) where
@@ -296,8 +297,9 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
             coeffs = FixedSizeVectorDefault{HomogeneousPolynomial{Taylor1{R}}}(
                     undef, order(a)+1)
             coeffs .= $f.(a.coeffs)
-            @inbounds coeffs[1] = aux
-            return TaylorN(coeffs, order(a))
+            c = TaylorN(coeffs, order(a))
+            constant_term!(c, aux)
+            return c
         end
 
         function ($f)(a::TaylorN{Taylor1{T}}, b::Taylor1{S}) where
@@ -306,8 +308,9 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
             R = TS.numtype(aux)
             coeffs = FixedSizeVectorDefault{HomogeneousPolynomial{Taylor1{R}}}(undef, order(a)+1)
             coeffs .= a.coeffs
-            @inbounds coeffs[1] = aux
-            return TaylorN(coeffs, order(a))
+            c = TaylorN(coeffs, order(a))
+            constant_term!(c, aux)
+            return c
         end
 
         function ($f)(b::Taylor1{S}, a::TaylorN{Taylor1{T}}) where
@@ -316,8 +319,9 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
             R = TS.numtype(aux)
             coeffs = FixedSizeVectorDefault{HomogeneousPolynomial{Taylor1{R}}}(undef, order(a)+1)
             coeffs .= $f.(a.coeffs)
-            @inbounds coeffs[1] = aux
-            return TaylorN(coeffs, order(a))
+            c = TaylorN(coeffs, order(a))
+            constant_term!(c, aux)
+            return c
         end
 
         function ($f)(a::Taylor1{TaylorN{T}}, b::S) where
--- a/src/arithmetic.jl
+++ b/src/arithmetic.jl
@@ -1836,6 +1836,16 @@ end
     return nothing
 end
 
+# c[k] <- a[k]/b, where b is a scalar
+@inline function div!(c::TaylorN{T}, a::TaylorN{T}, b::NumberNotSeries,
+        k::Int) where {T<:Number}
+    _check_same_space(c, a)
+    @inbounds for i in eachindex(c[k])
+        c[k][i] = a[k][i] / b
+    end
+    return nothing
+end
+
 # in-place division c <- c/a (assumes equal order among TaylorNs)
 function div!(c::TaylorN, a::TaylorN)
     @inbounds for k in eachindex(c)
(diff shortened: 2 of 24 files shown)


LuEdRaMo:
Nice work @PerezHz! This could be a great improvement for packages upstream.

I'm curious, will this PR also affect the performance of `Taylor1{Taylor1{T}}` operations?

PerezHz:
@LuEdRaMo in principle yes, this should help improve performance for nested `Taylor1` since it touches the `Taylor1{Float64}` kernels which the nested operations use. This is still a proof of concept; I'd like to add some benchmarks to get a better sense of the potential speedups here.

PerezHz:
Some copy-pasteable benchmarks:

```julia
using TaylorSeries, BenchmarkTools
const TS = TaylorSeries

ord = 30
a = Taylor1(rand(ord + 1), ord)
b = Taylor1(rand(ord + 1), ord)
c = zero(a)
aux = zero(a)

function test_add_kloop!(ord, a, b, c)
    @inbounds for k in 0:ord
        TS.add!(c, a, b, k)
    end
    return c
end
function test_muladd_kloop!(ord, a, b, c)
    fill!(c.coeffs, 0.0)
    @inbounds for k in 0:ord
        TS.muladd!(c, a, b, k)
    end
    return c
end
function test_pow_kloop!(ord, a, aux, c, r)
    @inbounds for k in 0:ord
        TS.pow!(c, a, aux, r, k)
    end
    return c
end

@btime test_add_kloop!($ord, $a, $b, $c);
@btime TS.add!($c, $a, $b);
@btime TS.mul!($c, $a, $b);
@btime test_muladd_kloop!($ord, $a, $b, $c);
@btime test_pow_kloop!($ord, $a, $aux, $c, 1.5);
@btime $a + $b;
@btime $a * $b;
@btime $a ^ 1.5;
```

This PR on my laptop:
```julia-repl
julia> @btime test_add_kloop!($ord, $a, $b, $c);
  9.926 ns (0 allocations: 0 bytes)

julia> @btime TS.add!($c, $a, $b); # doesn't exist in master
  4.791 ns (0 allocations: 0 bytes)

julia> @btime TS.mul!($c, $a, $b);
  74.632 ns (0 allocations: 0 bytes)

julia> @btime test_muladd_kloop!($ord, $a, $b, $c);
  80.489 ns (0 allocations: 0 bytes)

julia> @btime test_pow_kloop!($ord, $a, $aux, $c, 1.5);
  278.347 ns (0 allocations: 0 bytes)

julia> @btime $a + $b;
  19.475 ns (1 allocation: 272 bytes)

julia> @btime $a * $b;
  86.589 ns (1 allocation: 272 bytes)

julia> @btime $a ^ 1.5;
  265.975 ns (2 allocations: 544 bytes)
```

`master` on my laptop:
```julia-repl
julia> @btime test_add_kloop!($ord, $a, $b, $c);
  10.135 ns (0 allocations: 0 bytes)

julia> # @btime TS.add!($c, $a, $b); # does not exist on master

julia> @btime TS.mul!($c, $a, $b);
[ Warning: mul! is defined in LinearAlgebra and is not public in TaylorSeries
  185.098 ns (0 allocations: 0 bytes)

julia> @btime test_muladd_kloop!($ord, $a, $b, $c);
  182.570 ns (0 allocations: 0 bytes)

julia> @btime test_pow_kloop!($ord, $a, $aux, $c, 1.5);
  396.557 ns (0 allocations: 0 bytes)

julia> @btime $a + $b;
  22.401 ns (1 allocation: 272 bytes)

julia> @btime $a * $b;
  193.639 ns (1 allocation: 272 bytes)

julia> @btime $a ^ 1.5;
  443.182 ns (3 allocations: 816 bytes)
```

PerezHz:
I took a closer look at nested Taylor1 operations and found a couple of places where the all-orders-at-once versions of add! and mul! for non-nested Taylor1s can benefit the nested case. So I added some changes in the last commit which should help improve performance.

LuEdRaMo:
I tested this branch using NEOs' impact monitoring benchmark (which mostly relies on `Taylor1{Taylor1{T}}` integrations) and these are the results:

```Julia
# master

BenchmarkTools.Trial: 10 samples with 1 evaluation per sample.
 Range (min … max):  5.914 s … 10.634 s  ┊ GC (min … max): 0.28% … 0.59%
 Time  (median):     6.564 s             ┊ GC (median):    0.34%
 Time  (mean ± σ):   7.534 s ±  1.832 s  ┊ GC (mean ± σ):  0.36% ± 0.19%

  █                                                         
  █▁▁▁▁▇▁▇▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇▁▁▁▇ ▁
  5.91 s        Histogram: frequency by time        10.6 s <

 Memory estimate: 143.05 MiB, allocs estimate: 1834386.

# this PR

BenchmarkTools.Trial: 10 samples with 1 evaluation per sample.
 Range (min … max):  3.946 s …    4.228 s  ┊ GC (min … max): 0.41% … 0.41%
 Time  (median):     3.998 s               ┊ GC (median):    0.32%
 Time  (mean ± σ):   4.046 s ± 106.174 ms  ┊ GC (mean ± σ):  0.35% ± 0.06%

  ███   █  █ █               █     █                    █  █  
  ███▁▁▁█▁▁█▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁█ ▁
  3.95 s         Histogram: frequency by time         4.23 s <

 Memory estimate: 139.81 MiB, allocs estimate: 1818534.
```

PerezHz:
@LuEdRaMo many thanks for checking this, sounds great! Do you find differences which are larger than expected numerical noise?

PerezHz:
Fixes #411

PerezHz:
Also, extended this concept to `Taylor1{Float64}` and `Taylor1{Taylor1{Float64}}` methods: elementary functions, division helpers (`div_scalar!`), differentiation and integration, evaluation methods, as well as `zero!`, `one!`, `identity!` and `sqrt!`

PerezHz:
I'm also running downstream tests (TaylorIntegration, NEOs, PlanetaryEphemeris) using this branch

PerezHz:
@lbenet this is now ready for review. Worth noting that downstream tests for TaylorIntegration, PlanetaryEphemeris and NEOs are passing with this branch

lbenet:
> @lbenet this is now ready for review. Worth noting that downstream tests for TaylorIntegration, PlanetaryEphemeris and NEOs are passing with this branch

Thanks @PerezHz, in the next few days I'll be sending some comments (I'm at a conference this week).

Here is a very general comment (from skimming `src/arithmetic.jl`): While I like very much the approach and its benefits (e.g., `add!(a, b, c)`) I find it subtle that `@taylorize` will use a different method (i.e., `add!(a, b, c, k)`). While the example is perhaps not the best, my point is to use the new approach at the level of `add!(a, b, c, k)`, and also reuse that function as much as possible, e.g. in order to compute `a+b`, use `add!(a, b, c)` which loops over k and uses `add!(a, b, c, k)`.

Also, we should avoid code repetition; makes life simpler to maintain...

PerezHz:
@lbenet the idea of the new "full-loop" `add!(a, b, c)` method for `Taylor1{Float64}` is mainly to be used internally by `add!(a, b, c, k)` method for `Taylor1{TaylorN{Float64}}`, which makes addition for the latter type faster, since it does not have lots of calling overhead going back and forth for addition of nested-type variables (same goes for `subst!`, while for `mul!`, `div!` and `pow!` things are different since they do "heavier" computations).

That said, I agree on both points: we should make this consistent with how `@taylorize` is doing things, to avoid divergence between taylorized and non-taylorized computations, and I also agree that we should avoid code duplication where possible (I mainly focused on performance for this PR, but happy to look for places where code can be condensed and simplified). If you agree I can try to take a look into possibly condensing/simplifying code before you take a deeper look into the changes so far, what do you think?

lbenet:
> @lbenet the idea of the new "full-loop" `add!(a, b, c)` method for `Taylor1{Float64}` is mainly to be used internally by `add!(a, b, c, k)` method for `Taylor1{TaylorN{Float64}}`, which makes addition for the latter type faster, since it does not have lots of calling overhead going back and forth for addition of nested-type variables (same goes for `subst!`, while for `mul!`, `div!` and `pow!` things are different since they do "heavier" computations).

I see that the direct method (`add!(a, b, c)`) is faster, by avoiding the allocations that `a[k]` causes in favor of `a.coeffs[k+1]` (including the "views" a_coeffs). But I am not sure that I follow your argument, since it is not clear to me how `add!(a, b, c, k)` will use ``add!(a, b, c)`...

> That said, I agree on both points: we should make this consistent with how `@taylorize` is doing things, to avoid divergence between taylorized and non-taylorized computations, and I also agree that we should avoid code duplication where possible (I mainly focused on performance for this PR, but happy to look for places where code can be condensed and simplified). If you agree I can try to take a look into possibly condensing/simplifying code before you take a deeper look into the changes so far, what do you think?

I am already half-way through a first review, so let me finish it, and then you try condensing/simplifying the code.

PerezHz:
> But I am not sure that I follow your argument, since it is not clear to me how add!(a, b, c, k) will use ``add!(a, b, c)`...

I meant this for `Taylor1{Taylor1{Float64}}`. For example, [here](https://github.com/PerezHz/TaylorSeries.jl/blob/63f90ce5d4c5caa7fe5126ccc946411625d905c7/src/arithmetic.jl#L819) the `add!` method to update the k-th coefficient of a `Taylor1{Taylor1{Float64}}` is using the "full-loop" `add!(a, b, c)` to update the inner (or nested) coefficients in a single call.

> I am already half-way through a first review, so let me finish it, and then you try condensing/simplifying the code.

Alright, looking forward to receiving feedback to improve this PR!

src/arithmetic.jl
```diff
@@ -317,6 +317,132 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
...
 end
 
+function add!(v::Taylor1{T}, a::Taylor1{T}, b::Taylor1{T}) where
+        {T<:NumberNotSeries}
+    v_coeffs = v.coeffs
+    a_coeffs = a.coeffs
+    b_coeffs = b.coeffs
+    @inbounds for i in eachindex(v_coeffs)
+        v_coeffs[i] = a_coeffs[i] + b_coeffs[i]
+    end
```
lbenet: Note: This is **not** a suggestion for replacement (even as it appears as such), it is simply an illustration of my general comment [here](https://github.com/JuliaDiff/TaylorSeries.jl/pull/415#issuecomment-4644648762):

```suggestion
    for k in eachindex(a)
        add!(v, a, b, k)
    end
```

The idea is to use in the inner method `add!(v, a, b, k)` the kind of performance optimisations you are introducing in this PR.

I don't know if this change would provide the same performance improvements, but it is worth checking it. The key point is that the *same* operations would be implemented when operating directly with the optimized (`@taylorized`) methods, or directly with the Taylor1 objects.

src/arithmetic.jl
```diff
@@ -317,6 +317,132 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
...
+            v_hp = v_hps[j].coeffs
+            a_hp = a_hps[j].coeffs
+            b_hp = b_hps[j].coeffs
+            for k in eachindex(v_hp)
+                v_hp[k] = a_hp[k] - b_hp[k]
+            end
+        end
+    end
+    return nothing
+end
```
lbenet: I suggest to remove these methods here (involving `Taylor1{TaylorN{T}}`) and include them (using metaprogramming) below.

src/arithmetic.jl
```diff
@@ -317,6 +317,132 @@ for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
...
+    if order(a) != order(b)
+        a, b = fixorder(a, b)
+    end
+    c = zero(a)
+    subst!(c, a, b)
+    return c
+end
+
 for (f, fc) in ((:+, :(add!)), (:-, :(subst!)))
     @eval begin
```
lbenet, line 373: (Functions removed above are included here.)
```suggestion
    @eval begin
        function ($fc)(v::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
                b::Taylor1{TaylorN{T}}) where {T<:NumberNotSeries}
            v_coeffs = v.coeffs
            a_coeffs = a.coeffs
            b_coeffs = b.coeffs
            _check_same_space(v_coeffs[1], a_coeffs[1], b_coeffs[1])
            @inbounds for i in eachindex(v_coeffs)
                v_hps = v_coeffs[i].coeffs
                a_hps = a_coeffs[i].coeffs
                b_hps = b_coeffs[i].coeffs
                for j in eachindex(v_hps)
                    v_hp = v_hps[j].coeffs
                    a_hp = a_hps[j].coeffs
                    b_hp = b_hps[j].coeffs
                    for k in eachindex(v_hp)
                        v_hp[k] = ($f)(a_hp[k], b_hp[k])
                    end
                end
            end
            return nothing
        end
```

I think this could be written more compactly calling the corresponding functions with the order, i.e., `($fc)(v, a, b, k)`.

src/arithmetic.jl
```diff
@@ -611,6 +816,32 @@ function mul_scalar!(c::TaylorN{T}, scalar::NumberNotSeries, a::TaylorN{T},
...
+        b::Taylor1{Taylor1{T}}, k::Int) where {T<:NumberNotSeries}
+    @inbounds add!(v.coeffs[k+1], a.coeffs[k+1], b.coeffs[k+1])
+    return nothing
+end
+
+function subst!(v::Taylor1{Taylor1{T}}, a::Taylor1{Taylor1{T}},
+        b::Taylor1{Taylor1{T}}, k::Int) where {T<:NumberNotSeries}
+    @inbounds subst!(v.coeffs[k+1], a.coeffs[k+1], b.coeffs[k+1])
+    return nothing
+end
```
lbenet: I suggest to move these methods above, to keep them together with other methods of `add!` and `subst!`

src/arithmetic.jl
```diff
@@ -1514,13 +1899,14 @@ function div!(c::TaylorN, a::NumberNotSeries, b::TaylorN)
...
-#     return nothing
-# end
+# c[k] <- a[k]/b, where b is a scalar
+function div!(c::TaylorN{T}, a::TaylorN{T}, b::NumberNotSeries) where {T<:Number}
+    _check_same_space(c, a)
+    @inbounds for k in eachindex(c)
+        div!(c[k], a[k], b)
+    end
+    return nothing
+end
```
lbenet, line 1853: I think that #411 refers to `div!(c::TaylorN{T}, a::TaylorN{T}, b::NumberNotSeries, k::Int)`, and you may solve it in this PR.

src/evaluate.jl
```diff
@@ -57,11 +59,28 @@ Note that the syntax `a(x)` is equivalent to `evaluate(a, x)`.
...
+    aux_coeffs = aux.coeffs
+    a_coeffs = a.coeffs
+    @inbounds for k in reverse(eachindex(a_coeffs))
+        mul!(aux, suma, x)
+        copyto!(suma_coeffs, aux_coeffs)
+        suma_coeffs[1] += a_coeffs[k]
+    end
+    return suma
+end
+
```
lbenet, line 71: I think that (some of the) methods of `evaluate` should be written in terms of `_horner`. For instance

```suggestion
function evaluate(a::Taylor1{T}, x::Taylor1{T}) where {T<:NumberNotSeries}
    if order(a) != order(x)
        a, x = fixorder(a, x)
    end
    suma = zero(x)
    aux = zero(x)
    _horner!(suma, a, x, aux)
    return suma
end
```

src/functions.jl
```diff
@@ -232,18 +232,33 @@ for T in (:Taylor1, :TaylorN)
...
+    aux = log1p(constant_term(a))
+    aa = convert(Taylor1{typeof(aux)}, a)
+    c = _constant_series_like(a, aux, order(a))
+    denom = _constant_series_like(aux, zero(aux[0][1]), order(aux))
+    _log1p_denominator!(denom, aa.coeffs[1])
+    for k in eachindex(a)
+        log1p!(c, aa, k, denom)
+    end
+    return c
+end
```
lbenet, line 245: I think `log1p` is already coded (above), though not specialized for `Taylor1{TaylorN{T}}`. The question is why do we need this specific method *only* for `log1p`?

src/identity.jl
```diff
@@ -17,13 +17,21 @@ for T in (:Taylor1, :TaylorN)
         identity!(c[k], a[k])
 end
 
-identity!(c::Taylor1{T}, a::Taylor1{T}, k::Int) where {T<:NumberNotSeries} =
-    c[k] = a[k]
+function identity!(c::Taylor1{T}, a::Taylor1{T}) where {T<:NumberNotSeries}
+    copyto!(c.coeffs, a.coeffs)
```
lbenet, line 21: Does `copyto!` here (line 21) preserves the *independence* in memory of the elements of `a` and `c`?

src/power.jl
```diff
@@ -214,44 +231,57 @@ exploits `k_0`, the order of the first non-zero coefficient of `a`.
...
 """ pow!
 
-function pow!(c::Taylor1{T}, a::Taylor1{T}, aux::Taylor1{T},
-                r::S, k::Int) where {T<:NumberNotSeries, S<:Real}
-    (r == 0) && return one!(c, a, k)
-    (r == 1) && return identity!(c, a, k)
-    (r == 2) && return sqr!(c, a, constant_term(aux), k)
-    (r == 0.5) && return sqrt!(c, a, aux, k)
-    # Sanity
+@inline function _pow_taylor1_cached!(c::Taylor1{T}, a::Taylor1{T},
```
lbenet: I suggest to rename this function as `_pow_cached!` (and if there are other methods, let multiple dispatch do its magic).

src/power.jl
```diff
@@ -332,6 +362,61 @@ function pow!(res::Taylor1{TaylorN{T}}, a::Taylor1{TaylorN{T}},
...
+    @inbounds aux_k = aux_coeffs[kk]
+    ilo = max(0, l0+kprime-order_a)
+    ihi = min(k-lnull-1, order_a-lnull)
+    @inbounds for i = ilo:ihi
+        rr = r*(kprime-i) - i
+        mul_scalar!(aux_k, rr, c_coeffs[i+lnull+1], a_coeffs[l0+kprime-i+1])
+        add!(c_k, c_k, aux_k)
+    end
+    # c[k] = c[k] / (kprime * a[l0])
+    @inbounds copyto!(aux_k.coeffs, c_k.coeffs)
```
lbenet: Same comment as above (for `identity!`): does `copyto!` here preserves the independency in memory of `aux_k.coeffs` and `c_k.coeffs`?

src/TaylorSeries.jl
```diff
@@ -56,7 +56,7 @@ export getcoeff, derivative, integrate, differentiate,
     get_variable_names, get_variable_symbols,
     # jacobian, hessian, jacobian!, hessian!,
     ∇, taylor_expand, update!,
-    constant_term, linear_polynomial, nonlinear_polynomial,
+    constant_term, constant_term!, linear_polynomial, nonlinear_polynomial,
```
lbenet, line 59: Is `constant_term!` used? I think I only saw it in the tests...

test/manyvariables.jl
```diff
@@ -13,7 +13,7 @@ using Test
     @test b isa TaylorN{Complex{Float64}}
     @test_throws MethodError TaylorN(-4//7im)
     # Issue #85 is solved!
-    variables!("x", numvars=66, order=1)
+    variables!("x", numvars=66, order=1, nowarn=true)
```
lbenet, line 16: I suggest to keep at least one use of `variables!` (for each different method) with the default value (of `nowarn=false`), and check the actual thrown warning.

lbenet (commented):
First, I think this PR is great, finding a way to avoid some unnecessary allocations. Yet, I think some of the methods introduced define differing (code) paths that could change the behaviour in comparisson with the taylorized functions, which is not advisable. Also, I think some methods are too specialized, and we should specialize the inner methods (such as `add!(a, b, c, k)`). 
I've thus left some comments which are mostly suggestions. Their goal is to avoid repeated code, and make sure that the taylorized functions yield the same.

lbenet:
> I meant this for `Taylor1{Taylor1{Float64}}`. For example, [here](https://github.com/PerezHz/TaylorSeries.jl/blob/63f90ce5d4c5caa7fe5126ccc946411625d905c7/src/arithmetic.jl#L819) the `add!` method to update the k-th coefficient of a `Taylor1{Taylor1{Float64}}` is using the "full-loop" `add!(a, b, c)` to update the inner (or nested) coefficients in a single call.
> 

Thanks for the clarification; it is good point, and I agree in the usufulness of the method. Yet, consider that we may have `Taylor1{Taylor{Taylor1{<:NumberNotSeries}}}` which would not use those methods.

src/arithmetic.jl
PerezHz: Done (and added with metaprogramming below, as suggested)

src/arithmetic.jl
PerezHz: Done

src/arithmetic.jl
PerezHz: Right, thanks, I added that method so now this PR does fix #411

src/TaylorSeries.jl
PerezHz: You're right that `constant_term!` is not used a lot internally; I added it mainly as a public, high-level helper for users, as requested in #411 (and covered by tests). I do not think we should replace all internal constant-term assignments with it, because many of those are in inner-loop methods where direct coefficient assignment is intentional.

That said, I agree these methods can make code more readable; so I can also look for sites in TaylorSeries where using it improves readability without affecting performance-critical inner loops.

test/manyvariables.jl
PerezHz: Thanks; added tests for this

src/power.jl
PerezHz: Thanks for the suggestion; done 👍 (there's currently only one method, but this change does leave the room open in case we want to add similar methods for other types in the future)

src/identity.jl
PerezHz: Yes independence in memory is preserved here: since `NumberNotSeries = Union{Real, Complex}`, the elements of `c.coeffs` and `a.coeffs` are scalar numbers, not mutable series objects. `copyto!` copies the scalar element values from `a.coeffs` into the existing `c.coeffs` vector, element-by-element. It does not make `c.coeffs` point to `a.coeffs`, so the vectors remain independent. So for `Float64`, `ComplexF64`, etc., this is safe; we would only have to be careful in the case that there is a user-defined type `<: Real` which is actually some kind of `mutable struct`. That's also the reason why I have not used `copyto!` for other methods of `identity!`, where nested Taylor polynomials can occur.

src/power.jl
PerezHz: Yes, in this case we are dealing with scalar coefficients, so the reply is the same: in-memory independence is preserved. I actually figured we could precisely use the `identity!` method from your other comment here, and that makes the code more readable, and consistent.

src/functions.jl
PerezHz: The reason for this specialization is only performance. In the `Taylor1{TaylorN}` mutating recurrence, `log1p!` needs the TaylorN denominator `1 + a[0]` for every outer order. The generic path calls `log1p!(c, aa, k)`, whose no-auxiliary method rebuilds that denominator for each `k > 0`. This method precomputes it once and calls the cached-denominator form `log1p!(c, aa, k, denom)`.

That said, I agree the specialization looks kind of asymmetric. I can either remove it, or keep it with a comment/helper making clear that it is only a method added for performance.

lbenet: I'm fine to leave it like this.

src/evaluate.jl
PerezHz: I updated the `evaluate` method above as suggested (thanks!), and I think that covers evaluate methods where the result of an evaluation (and/or the evaluation argument) is a `Taylor1`, which is where current `_horner1!` methods are applicable. That said, I'd like to mention I'm working on a follow-up PR #416 which will help reduce allocations of `evaluate!`, which is important for TaylorIntegration, but left it separate from here to help keep things separated conceptually, and PR diffs easier to read for everyone. I would suggest to leave further changes to `evaluate.jl` for #416.

src/arithmetic.jl
PerezHz: Thank you @lbenet for this suggestion. I double-checked this, and it does look like adding the optimizations at the `add!(v, a, b, k)` level gives essentially the same performance, with a difference ~0.5 ns, which is already around the noise level (this probably means that somehow the Julia compiler manages to inline away the loops and generates optimized code very similar to what the previous approach did). So I modified `add!(v, a, b, k)` and `subst!(v, a, b, k)` to use operate directly on the `.coeffs`.

src/evaluate.jl
lbenet: Thanks for the clarification; fine with me to keep the discussion of these methods in #416. Resolving this...

PerezHz:
Updated benchmarks on my laptop:

```julia-repl
julia> @btime test_add_kloop!($ord, $a, $b, $c); # dropped from 9.926 ns vs last run of this PR, 10.135 ns on `master`
  4.458 ns (0 allocations: 0 bytes)

julia> @btime TS.add!($c, $a, $b);
  4.416 ns (0 allocations: 0 bytes)

julia> @btime TS.mul!($c, $a, $b);
[ Warning: mul! is defined in LinearAlgebra and is not public in TaylorSeries
  74.554 ns (0 allocations: 0 bytes)

julia> @btime test_muladd_kloop!($ord, $a, $b, $c);
  81.006 ns (0 allocations: 0 bytes)

julia> @btime test_pow_kloop!($ord, $a, $aux, $c, 1.5);
  279.394 ns (0 allocations: 0 bytes)

julia> @btime $a + $b;
  19.204 ns (1 allocation: 272 bytes)

julia> @btime $a * $b;
  87.022 ns (1 allocation: 272 bytes)

julia> @btime $a ^ 1.5;
  267.590 ns (2 allocations: 544 bytes)
```

src/TaylorSeries.jl
PerezHz: I've added internal use of `constant_term!`where appropriate

PerezHz:
@lbenet thanks a lot for your suggestions and comments; I think I managed to go through all of them; so this is ready for you to have another look

lbenet:
Thanks a lot! I'll have another look, everything seems promising, but probably will do it tomorrow. Yet, I've just checked (locally) if tests for `TaylorIntegration` pass, and there are two failing tests (in `taylorize.jl`, for `kepler1!` defined in line 2093 or around it). The first test that fails is because taylorizing yields no problems (so line 2107 yields no logs), and then `parse_eqs` is `true` (line 2116). It would be worth checking if the produced series are the expected ones...

PerezHz:
@lbenet thanks for checking this; yes fixing #411 causes an errored test to pass in TI; I have opened https://github.com/PerezHz/TaylorIntegration.jl/pull/243 which tests TI with this TS branch (plus further performance improvements on TI side); will update it once this is merged.

lbenet:
I am ok with this PR as it is, but fear how it affects TaylorIntegration.jl and/or TaylorModels.jl. I see that TaylorIntegration (using https://github.com/PerezHz/TaylorIntegration.jl/pull/243) now has passing tests, but also lots of other things, which seem to me a bit unrelated. I'll check how TaylorModels behaves...

lbenet:
I'll go ahead and merge this. As far as I've checked everything is fine with TaylorIntegration.jl and TaylorModels.jl

lbenet:
Thanks a lot @PerezHz for this!

PR #416, opened 2026-06-11
Performance improvements for evaluation methods

PerezHz:
## Disclaimer

Code in this PR has been assisted by an LLM.

## Summary of this PR and intended design

This PR contains mainly changes to `src/evaluate.jl` in order to help TaylorIntegration, as well as any users of `TaylorSeries.evaluate!` avoid allocating when using such methods. We have identified this can help performance, especially for JT integrations. This PR improves the overall performance of polynomial evaluation, particularly for the nested Taylor types used by TaylorIntegration.jl and NEOs.jl. It also cleans up and documents the evaluation API.

The main design that I tried to follow is the following:

`evaluate!(a, x, dest[, scratch...])` is the lower-level, in-place evaluation API. It evaluates `a(x)` into `dest`. Numeric (non-series) evaluations require only `dest`; series-valued evaluations may require reusable scratch space (caches). Ideally `evaluate!` is allocation-free in all cases, though in practice it is allocation-free only for all uses in TaylorIntegration.jl and NEOs.jl, while some edge cases still need to be covered (outside of the scope of this PR, but added TODOs where appropriate).

`evaluate(a, x)` is the higher-level, out-of-place API. Conceptually, it allocates an appropriate destination, any required workspace if needed, and then performs the same evaluation.

Thus, the intended layering is approximately:
```julia
evaluate(a, x) ≈ begin
    dest = allocate_result(a, x)
    scratch = allocate_workspace_if_needed(a, x, dest)
    evaluate!(a, x, dest, scratch...)
    dest
end
```

This is the conceptual API rather than the literal implementation of every existing method. Some evaluate methods currently use broadcasting to preserve promotion and container behavior. Using `sorting=true` for evaluation of `TaylorN`s, and some other methods, may still allocate. As mentioned before, those are not methods currently used by TI/NEOs’ allocation-sensitive `evaluate!` calls.

Additional housekeeping includes adding internal helpers and documentation/docstrings where appropriate, support for abstract arrays and views, validation of workspace aliasing and `JetSpace` compatibility, and correct handling of mixed-order coefficients and order-zero exact constants.

## Benchmarks

Benchmarks were run with Julia 1.12.6 and BenchmarkTools 1.8.0, comparing`master` at `9a7c39b3` ( TS v0.22.4) against this PR at `2f95f3b2`. Inputs were warmed-up before measurement, and the table reports minimum times.

Representative workloads used 12 state polynomials of outer order 16. Nested `Taylor1` coefficients had order 10; `TaylorN` workloads used three variables of order 4.

| Workload |`master` | This PR | Change | Allocations: `master` → PR |
|---|---:|---:|---:|---:|
| `Taylor1{Taylor1}` at a number | 5.29 μs | 1.54 μs | 3.4× faster | 47,040 B/420 → 0 |
| `Taylor1{Taylor1}` at a `Taylor1` | 5.04 μs | 4.79 μs | 1.05× faster | 4,032 B/36 → 0 |
| `Taylor1{TaylorN}` at a number | 16.08 μs | 3.08 μs | 5.2× faster | 16,512 B/180 → 0 |
| `Taylor1{TaylorN}` at a `TaylorN` | 31.42 μs | 29.63 μs | 1.06× faster | 16,512 B/180 → 0 |
| `TaylorN` array at numbers, `sorting=false` | 37.67 μs | 0.58 μs | 64.6× faster | 49,056 B/1,740 → 0 |
| `TaylorN` array at numbers, `sorting=true` | 38.88 μs | 11.17 μs | 3.5× faster | 49,056 B/1,740 → 11,136 B/624 |
| Scalar `TaylorN` at `TaylorN`s | 4.17 μs | 3.42 μs | 1.22× faster | 6,464 B/74 → 0 |
| `TaylorN` array at `TaylorN`s | 39.08 μs | 40.46 μs | approximately unchanged | 5,632 B/62 → 0 |

For series-valued workloads, the PR measurements use the new explicit-workspace overloads.`master` uses the equivalent existing allocating operation. The numerical results were checked for equivalence.

The strongest improvement is allocation elimination. Some series-valued workloads have similar raw execution times, but now perform zero heap allocations when the caller provides reusable workspace.

### Minimal benchmark

As an example, the following runs unchanged on `master` and this PR:

```julia
using TaylorSeries
using BenchmarkTools

space = JetSpace(order=4, variables=[:x, :y, :z])
x, y, z = variables(space)

ordt = 16 # outer order (time)
ordq = 12 # inner order (state variables)

coeffs = [
    sin((k + 1) * x + y) / (k + 1) + z^2 / (k + 2)
    for k in 0:ordt
]

src = [
    Taylor1([(1 + 0.01 * j) * c for c in coeffs], ordt)
    for j in 1: ordq
]
dest = [zero(x) for _ in src]

evaluate!(src, 0.125, dest) # warm-up
@btime evaluate!($src, 0.125, $dest)
```

This produces the following output on my laptop:

```text
`master`:     15.875 μs (180 allocations: 16.12 KiB)
this PR:   2.889 μs (0 allocations: 0 bytes)
```

## Downstream effects

As far as I can tell, changes introduced by this PR require only minimal changes in TI and NEOs.jl, and corresponding PRs are being prepared for that. In this sense, this PR intends to be as non-breaking as possible.


lbenet:
You may need to rebase this, since #419 touched evaluation.jl (and other things related to intervals)... sorry

PerezHz:
@lbenet this is now ready for review

src/evaluate.jl
```diff
@@ -8,22 +8,32 @@
...
-equivalent to `evaluate(a,dx)`, and `a()` is equivalent to `evaluate(a)`.
+Evaluate `a` at the ordinary numeric scalar `dx` using Horner's rule without
+creating intermediate containers. This is the scalar kernel shared by
+`evaluate` and the corresponding array `evaluate!` method.
 """
-function evaluate(a::Taylor1{T}, dx::S) where {T<:NumberNotSeries, S<:NumberNotSeries}
+@inline function _evaluate_taylor1_scalar(a::Taylor1{T}, dx::S) where
+        {T<:NumberNotSeries, S<:NumberNotSeries}
     a_coeffs = a.coeffs
     @inbounds suma = zero(a_coeffs[end])
```
lbenet, line 20: For type stability, I think the following is requires:
```suggestion
@inline function _evaluate_taylor1_scalar(a::Taylor1{T}, dx::S) where
        {T<:NumberNotSeries, S<:NumberNotSeries}
    a_coeffs = a.coeffs
    @inbounds suma = zero(a_coeffs[end])*dx
```

(This is not something you introduced, it was already there, but still worth correcting.)

lbenet: This is function implements the core of Horner's method. Shall we call it `_horner` and define the needed methods (if there are more) exploiting dispatch?

src/evaluate.jl
```diff
@@ -78,8 +90,29 @@ function evaluate(a::Taylor1{T}, x::Taylor1{T}) where {T<:Number}
...
 end
 
+"""
+    _evaluation_order(a, x)
+
+Return the minimum positive order represented by `x` and the coefficients of
+`a`, or zero if they all have order zero. Order-zero series represent exact
+constants, so they do not truncate a positive-order composition.
+"""
+function _evaluation_order(a::Taylor1{<:AbstractSeries}, x::AbstractSeries)
```
lbenet, line 100: There are two internal functions (`orderH` and `maxorderH`) in `auxiliary.jl` which return some information related to the order in some cases. The question is, should we move this method there? It seems aproapriate to me.

src/evaluate.jl
```diff
@@ -78,8 +90,29 @@ function evaluate(a::Taylor1{T}, x::Taylor1{T}) where {T<:Number}
...
+    ord = order(x)
+    for coeff in a.coeffs
+        coeff_order = order(coeff)
+        # Only positive orders limit the information available in the result.
+        if ord == 0
+            ord = coeff_order
+        elseif coeff_order > 0
+            ord = min(ord, coeff_order)
+        end
```
lbenet, line 109: - Since the condition `ord == 0` does not depend on `coeff_order`, I think the first part of the `if` can be moved out from the `for` (returning `order(a.coeffs[end])`. Am I right?
- Can we somehow avoid the `elseif` branch?

lbenet: I guess the motivation for this function is to avoid havinf problems (dimension mismatches) within `_horner!`. Have you encounter such problems?

src/evaluate.jl
```diff
@@ -191,22 +224,59 @@ function evaluate(a::HomogeneousPolynomial{T}) where {T}
...
-    # @assert length(vals) == get_numvars()
+"""
+    _evaluate_homogeneous_scalar(a, vals)
+
+Evaluate one homogeneous polynomial directly at `vals` without constructing
+vectors of powers or monomial terms. The caller must supply a nonempty
+collection with one value per variable and, for series values, compatible
+series spaces.
+"""
+function _evaluate_homogeneous_scalar(a::HomogeneousPolynomial{T},
```
lbenet, line 235: The proposal above is to call all those methods `_horner`, and distinguish their appicability by dispatch.

src/evaluate.jl
```diff
@@ -191,22 +224,59 @@ function evaluate(a::HomogeneousPolynomial{T}) where {T}
...
+        vals) where {T}
     order(a) == 0 && return a[1]*one(vals[1])
     ct = a.space.coeff_table[order(a)+1]
-    suma = zero(a[1])*vals[1]
-    vv = Base.literal_pow.(^, vals, Val.(ct[1]))
+    suma = zero(a[1]*one(vals[1]))
```
lbenet, line 239: My naive guess is that `suma = zero(a[1])*vals[1]` is *cheaper* than `suma = zero(a[1])*one(vals[1])`; is there any specific use for `one` here?

src/evaluate.jl
```diff
@@ -191,22 +224,59 @@ function evaluate(a::HomogeneousPolynomial{T}) where {T}
...
+            term *= Base.literal_pow(^, vals[j], Val(exponent))
+        end
+        suma += term
+    end
+    return suma
+end
+
+# Ordinary scalar vectors can use type-level promotion. The generic method
+# above derives zeros and ones from instances to retain series-space metadata.
+function _evaluate_homogeneous_scalar(a::HomogeneousPolynomial{T},
```
lbenet, line 255: Same suggestion as above, about the name of this function

src/evaluate.jl
```diff
@@ -333,8 +403,8 @@ evaluate(a::TaylorN{T}, x::Pair{Symbol,S}) where {T, S} =
 
 evaluate(a::TaylorN{T}) where {T<:Number} = constant_term(a)
 
-
-# _evaluate
+# TODO: Avoid the component-vector and sorting allocations produced by
+# `sorting=true`, likely by accepting caller-provided workspace.
```
lbenet, line 407: I'm not sure what this means. The idea of the `sorting`, which indeed costs time, I to add (in increasing order) the terms of the sum, a priory for precision. I think this is independent of the provided workspace, but maybe I'm misunderstanding something.

src/evaluate.jl
```diff
@@ -359,7 +429,24 @@ function _evaluate(a::TaylorN{T}, vals::NTuple{N,<:Number}) where {N,T<:Number}
...
+end
+
+"""
+    _evaluate_taylorN_scalar(a, vals)
+
+Evaluate `a` by summing its homogeneous components in stored order. This is
+the direct unsorted scalar kernel and avoids the temporary component vector
+used by magnitude-sorted evaluation.
+"""
+function _evaluate_taylorN_scalar(a::TaylorN{T},
```
lbenet, line 444: Same suggestion as above, to rename it as `_horner` and use dispatch

src/evaluate.jl
```diff
@@ -504,38 +597,228 @@ evaluate(A::AbstractArray{TaylorN{T}}) where {T<:Number} = evaluate.(A)
...
+    end
+    return nothing
+end
+
+function evaluate!(x::AbstractArray{Taylor1{Taylor1{T}}}, δt::S,
+        dest::AbstractArray{Taylor1{R}}) where
+        {T<:NumberNotSeries,S<:NumberNotSeries,R<:NumberNotSeries}
+    @inbounds for i in eachindex(x, dest)
+        evaluate!(x[i], δt, dest[i])
+    end
```
lbenet, line 650: This could be broadcasted as follows:
```suggestion
    evaluate!.(x, Ref(δt), dest)
```
I'm not sure if it helps or not, but since you are (above) in favor of such thing, I include this remark.

src/evaluate.jl
```diff
@@ -504,38 +597,228 @@ evaluate(A::AbstractArray{TaylorN{T}}) where {T<:Number} = evaluate.(A)
...
 function evaluate!(x::AbstractArray{Taylor1{TaylorN{T}}}, δt::S,
-        dest::AbstractArray{TaylorN{T}}) where {T<:Number, S<:Number}
-    @inbounds for i in eachindex(x)
-        zero!(dest[i])
-        aux = zero(dest[i])
-        _horner!(dest[i], x[i], δt, aux)
+        dest::AbstractArray{TaylorN{T}}) where {T<:Number, S<:NumberNotSeries}
+    @inbounds for i in eachindex(x, dest)
+        evaluate!(x[i], δt, dest[i])
+    end
```
lbenet, line 689: Same comment on perhaps using broadcasting here.

src/evaluate.jl
```diff
@@ -547,21 +830,124 @@ function _evaluate!(a::TaylorN{T}, vals::NTuple{N,TaylorN{T}}, dest::TaylorN{T},
...
 
+"""
+    _check_taylorN_evaluation(a, vals, dest, valscache, aux)
+
+Validate dimensions, `JetSpace` and order compatibility, and non-aliasing
+requirements for `TaylorN` evaluation with explicit workspace. Cache entries
+are destructive scratch and must be distinct from every input and from one
+another. All checks happen before the destination is mutated.
+"""
+function _check_taylorN_evaluation(a::TaylorN{T}, vals::NTuple{N,TaylorN{T}},
```
lbenet, line 841: I'm not yer sure where this function is used, but aren't all the checks (`if`s) too costly for this function?

src/evaluate.jl
```diff
@@ -547,21 +830,124 @@ function _evaluate!(a::TaylorN{T}, vals::NTuple{N,TaylorN{T}}, dest::TaylorN{T},
...
+    if sorting
+        result = evaluate(a, vals; sorting=true)
+        zero!(dest)
+        for ord in eachindex(dest)
+            identity!(dest, result, ord)
+        end
+    else
+        zero!(dest)
+        _evaluate!(a, vals, dest, valscache, aux)
+    end
```
lbenet, line 915: Instead of this `if`, shouldn't we use two methods and dispatch, with a `Val(sorting)` argument?

test/manyvariables.jl
```diff
@@ -114,6 +114,13 @@ end
     @test_throws ArgumentError x + a
     @test_throws ArgumentError x * a
 
+    x32, y32, _ = variables(Float32, sx)
+    vals64 = variables(Float64, sx) .+ [0.1, 0.2, 0.3]
+    explicit_space_src = [x32 + y32^2]
+    explicit_space_dest = [zero(vals64[1])]
+    evaluate!(explicit_space_src, vals64, explicit_space_dest; sorting=false)
+    @test explicit_space_dest[1] == vals64[1] + vals64[2]^2
```
lbenet, line 122: Could we add a test using `evaluate` directly, that would end up calling this method of `evaluate!`?

lbenet: Same comment applies in some of the tests added below.

src/evaluate.jl
lbenet, line 1: The key points I find in this file are:
- Some of the methods essentially do the same, but with specificity of the input arguments, so I think they should be renamed to something more familiar, like `_horner`, and exploit dispatch.
- It seems to me, but I may be wrong, that some nested `Taylor1` evaluations, only cover the case `Taylor1{Taylor1{<:NumberNotSeries}}`, but no more nesting. Is this true, or only a wrong impression?
- It is not too obvious to me when some of the `evaluate!` methods are called.

lbenet: Also, some of the docstrings are really too obscure, so perhaps some rewording would be appreciated.

test/mixtures.jl
```diff
@@ -161,8 +161,26 @@ using Test
...
+    vt1N = [t1N, t1N^2]
+    @test eltype(evaluate(vt1N, 0.5im)) == TaylorN{ComplexF64}
+    @test isnothing(evaluate!(vt1N, 0.0, v))
     @test v == [TaylorN(1), TaylorN(1)^2]
+    evaluate!(vt1N, 0.5, v)
+    @test (@allocated evaluate!(vt1N, 0.5, v)) == 0
+    δtN = 0.5 + t1N[0]
+    auxN = zero(v[1])
+    evaluate!(vt1N, δtN, v)
+    @test isapprox(v, evaluate(vt1N, δtN))
```
lbenet, line 173: Can we make more explicit (i.e. as sharp as possible) this type of use of `isapprox`?

lbenet (commented):
I have left a bunch of comments, specially for `src/evaluate.jl`. While I believe I understand the general approach, I think there are many methods which are not obvious to me when they are called. Let me know your what you think of my comments.

src/evaluate.jl
PerezHz: Thanks for this suggestion, I agree it makes sense to condense them into a single helper and let dispatch do its magic. I see existing methods for `_evaluate` which do similar operations for special cases, so I renamed this helpers to `_evaluate` since it seemed an appropriate place, but I can also separate them into `_horner` if you think that would be better.

src/evaluate.jl
PerezHz: I had added it as a way to handle better cases like `vals[1] = Inf`, but I agree that while safer, it can induce allocations, so I've removed it.

src/evaluate.jl
PerezHz: Agreed; renamed to `_evaluate`

src/evaluate.jl
PerezHz: Yes, I modified this docstring to explain the code better, I agree the text was not the clearest. "workspace" meant some pre-allocated cache, but that was confusing with a variable space (`JetSpace`).

src/evaluate.jl
PerezHz: Renamed to `_evaluate` like the other two helpers

src/evaluate.jl
PerezHz: In this case, broadcasting without a broadcasted assignment ` .= `, eg like (`a .= b .+ c`) would allocate a vector of `nothing`s (broadcasting operates component-wise and each call to `evaluate!` returns `nothing` so the end result is `Vector{Nothing}`). So I'd say in this case the loop is better (though I agree maybe there's a way to write the loop in a more compact form)

src/evaluate.jl
PerezHz: Broadcasting here would also allocate a `Vector{Nothing}`

src/evaluate.jl
PerezHz: Thanks for this suggestion, done 👍

src/evaluate.jl
PerezHz: This function is called in one of the `evaluate!` methods for `TaylorN` which substitutes a `TaylorN` into another `TaylorN`, and the method for an array of `TaylorN` calls it once per polynomial:

```julia
using TaylorSeries

space = JetSpace(order=2, variables=[:x, :y])
x, y = variables(space)

a = x + y^2
vals = (x + 1, y + 2)

dest = zero(x)
auxiliaries = [zero(x), zero(y)]
aux = zero(x)

evaluate!(a, vals, dest, auxiliaries, aux) # _check_taylorN_evaluation is called in here

dest == (x + 1) + (y + 2)^2  # true
```

I would say compared to the evaluation workload, the checks represent minimal overhead, though I agree we could reduce repeated work by running these checks only once in the array method.

src/evaluate.jl
lbenet: Thanks for the explanation; resolving this point (and those related below)

src/evaluate.jl
PerezHz: I changed these checks to run only once per array of `TaylorN`s

src/evaluate.jl
PerezHz: This was just trying to say that using `sorting=true` induces allocations, so the TODO means getting rid of those allocations by adding some pre-allocated cache which can be used to avoid those.

src/evaluate.jl
PerezHz: Thanks, done 👍

src/evaluate.jl
PerezHz: > Since the condition ord == 0 does not depend on coeff_order, I think the first part of the if can be moved out from the for (returning order(a.coeffs[end]). Am I right?

`ord` is updated precisely in the first part of the if `ord = coeff_order`, so it changes throughout the loop, and it keeps track of the smallest positive order in the coefficients of `a`.

> I guess the motivation for this function is to avoid havinf problems (dimension mismatches) within _horner!. Have you encounter such problems?

The main motivation of this helper here is to compute the minimum order that we need to accommodate the result of evaluating a `Taylor1{Taylor1{T}}` in a `Taylor1{T}`, this is used here:
```julia
function evaluate(a::Taylor1{Taylor1{T}}, x::Taylor1{T}) where {T<:NumberNotSeriesN}
    suma = Taylor1(zero(x[0]), _evaluation_order(a, x))
    _horner!(suma, a, x, zero(suma))
    return suma
end
```
That said, I agree maybe that's not what we want, do you think we should keep instead the maximum order?

src/evaluate.jl
PerezHz: > It seems to me, but I may be wrong, that some nested Taylor1 evaluations, only cover the case Taylor1{Taylor1{<:NumberNotSeries}}, but no more nesting. Is this true, or only a wrong impression?

Thanks for this comment, you are right that nested `Taylor` evaluation was only covering that case, while `main` has a generic fallback for deeper nesting (which does allocate). I have brought that generic fallback.

PerezHz: > Some of the methods essentially do the same, but with specificity of the input arguments, so I think they should be renamed to something more familiar, like _horner, and exploit dispatch.

Done

> It is not too obvious to me when some of the evaluate! methods are called.

Your'e right that we have many `evaluate!` methods and it might not be clear which one is applicable to which situation. I tried improving the docstrings for `evaluate!` and added a couple of examples, but let me know if you think we should add more clarity to that.

test/manyvariables.jl
PerezHz: Done in both cases

test/mixtures.jl
PerezHz: Thanks, replaced the two approximate comparisons with exact equality.

PerezHz:
@lbenet thank you for all your comments and suggestions! I think I went through all of them and have addressed them accordingly. So this is ready for you to take another look.

src/evaluate.jl
```diff
@@ -345,6 +406,7 @@ function _evaluate(a::TaylorN{T}, vals::NTuple{N,<:TaylorN}, ::Val{false}) where
         {N,T<:Number}
     R = promote_type(T, TS.numtype(vals[1]))
     _check_same_space(a, vals[1])
+    a = convert(TaylorN{R}, a)
```
lbenet, line 409: We could do here something like
```suggestion
    a = T == R ? a : convert(TaylorN{R}, a)
```
Two things I'm not sure about this suggestion: (1) the produced `a` (after this suggestion) changes/mutates or not the `a` that was used in the input argument of the function; ideally, it should mutate internally in the scope of this function, but not the one provided by the user. (2) Do we need `R == T` or `R === T`?

lbenet (commented):
I left a comment which is requires a simple check (and perhaps will be ignored). Besides that, everything LGTM. If possible, I suggest to `git mv` the file `src/evaluate.jl` to `src/evaluate/evaluate.jl`, to restructure slowly the package (and eventually split this file since it is very long).

src/evaluate.jl
PerezHz: Thanks for this suggestion! In the conversion methods (`src/conversion.jl`) we have `convert(::Type{TaylorN{T}}, a::TaylorN{T}) where {T<:Number} = a` so when `T == R` conversion already returns the same type, so I understand this shouldn't be needed? Also, AFAIU assignment to `a` only changes the binding inside the method body, so it shouldn't affect `a` outside that scope (it would be different if we changed the fields, i.e., the coeffs of `a`), so I think I would be inclined to keep the line as it is now.

lbenet: I think you are right, so there is no need at all of the `if`. So let's keep this as it is.

PR #418, opened 2026-06-18
Define `JetSpace` equality

### LINKED ISSUE #417: Define `==` for `JetSpace`
https://github.com/JuliaDiff/TaylorSeries.jl/blob/7a8ccbd1a59d462476235907d9554cf4448fa962/src/total_order.jl#L23-L29

The code above compares the spaces. But there is no definition of `==`, so it uses the fallback `===`. This should change, as you can create a new space that is identical and then the comparison should return `true` (in my opinion).

PerezHz:
This PR is an attempt to fix #417:

- Adds structural `JetSpace` `==` and matching hash. By structural I mean comparing order, number of variables, variable names, and symbols, not caches/locks.
- Updates `TaylorN` and `HomogeneousPolynomial` equality to use `space(a) == space(b)`.
- Leaves arithmetic checks identity-based via `_check_same_space`, so mixed-space operations still throw.
- Updates `set_default_space!` to keep and return the existing default object when the incoming space is structurally equal.
- Updates `variables!` to reuse the existing default space instead of constructing/replacing an equivalent one.
- Adds tests for equality, hashing, no-warning reuse, and continued arithmetic rejection across distinct space identities.

cc: @lbenet @schillic

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
-version = "0.22.1"
+version = "0.22.2"
 repo = "https://github.com/JuliaDiff/TaylorSeries.jl.git"
 
 [deps]
--- a/src/hash_tables.jl
+++ b/src/hash_tables.jl
@@ -243,12 +243,16 @@ such as `variables!` replace `default_space[]` with `space`. Existing `TaylorN`
 and `HomogeneousPolynomial` objects keep their original spaces, while future
 default-space constructors use the new default algebra.
 
+If `space` is structurally equal to the current default space, the existing
+default space object is kept and returned.
+
 Use `nowarn=true` to suppress the warning emitted when replacing the default
 space.
 """
 function set_default_space!(space::JetSpace; nowarn::Bool=false)
     old_space = default_space[]
     old_space === space && return space
+    old_space == space && return old_space
 
     msg = "Updating TaylorSeries.default_space[]; existing TaylorN and " *
         "HomogeneousPolynomial objects keep their original JetSpace, while " *
--- a/src/parameters.jl
+++ b/src/parameters.jl
@@ -196,6 +196,38 @@ order(space::JetSpace) = space.order
 get_numvars(space::JetSpace) = space.num_vars
 get_variable_names(space::JetSpace) = space.variable_names
 get_variable_symbols(space::JetSpace) = space.variable_symbols
+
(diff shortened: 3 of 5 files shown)


src/hash_tables.jl
```diff
@@ -243,12 +243,16 @@ such as `variables!` replace `default_space[]` with `space`. Existing `TaylorN`
...
 
+If `space` is structurally equal to the current default space, the existing
+default space object is kept and returned.
+
 Use `nowarn=true` to suppress the warning emitted when replacing the default
 space.
 """
 function set_default_space!(space::JetSpace; nowarn::Bool=false)
     old_space = default_space[]
     old_space === space && return space
```
schillic, line 254:
```suggestion
```
This line is now redundant.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "TaylorSeries"
 uuid = "6aa5eb33-94cf-58f4-a9d0-e4b2c4fc25ea"
-version = "0.22.1"
+version = "0.22.2"
```
schillic, line 3: Not sure this qualifies as a patch version. It changes the behavior.

schillic (approved):
Thanks, in my later comments of #417 I did not mean to change the definition of `==` anymore. But if you think it makes sense, I will also not complain 🙂

schillic:
Sorry, while this is helpful, it is not solving all of my issues.
Is there a reason why `HomogeneousPolynomial`s need to have the same space? In particular, it could be sufficient that the _relevant_ variables are the same. Consider the extreme case below where both polynomials are constants but created for different spaces.

```julia
julia> cp
 33.0

julia> cq
 33.0

julia> cp.space
JetSpace
    Expansion order:        3
    Number of variables:    3
    Variable names:         ["x₁", "x₂", "x₃"]
    Variable symbols:       [:x₁, :x₂, :x₃]

julia> cq.space
JetSpace
    Expansion order:        3
    Number of variables:    5
    Variable names:         ["x₁", "x₂", "x₃", "x₄", "x₅"]
    Variable symbols:       [:x₁, :x₂, :x₃, :x₄, :x₅]

julia> cp == cq
false
```

I understand that this is much more intricate to implement. So would it be possible to provide a function (again, it does not have to be `==`) to check equality of polynomials while ignoring the space?

lbenet:
> Is there a reason why HomogeneousPolynomials need to have the same space?

My naive guess to answer this question is that `HomogeneousPolynomial`s are the basic consitituents of `TaylorN`, and in that sense they need to be related to a `JetSpace`.

PR #419, opened 2026-06-23
Improvements when coeffs are Intervals

lbenet:


lbenet:
This PR removes some NGs (`isguaranteed == false`) produced by evaluation, and which could be avoided. Implications of this PR are reflected in TaylorModels.

Issue #420, opened 2026-06-29
Definition of `round(Int, ::AbstractSeries)`


LuEdRaMo:
In NEOs.jl, in order to update to the latest version of SatelliteToolboxTransformations.jl (v1.12), we need to define `round(Int, ::AbstractSeries)`. Therefore, I am opening this issue to discuss what the appropriate definition would be. Intuitively, the above method should simply fall back to rounding the constant term of the corresponding series. What do you think @lbenet @PerezHz?

lbenet:
While I naively agree with your expectation, that is, apply round to the constant term and leave the rest as is, I think it would be worth to have a bit more context about its use.

LuEdRaMo:
In short, SatelliteToolboxTransformations.jl creates an interpolation object representing the time evolution of the Earth Orientation Parameters (EOP) which NEOs.jl then evaluates in a time of type `::AbstractSeries` for various applications. Such evaluation is handled by DataInterpolations.jl which uses a function `get_idx` to know which element of the vector of coefficients to use for the evaluation. It is the `get_idx` function that ultimately requires to "round" the `AbstractSeries`. More precisely, since `get_idx` is expected to return an integer, `round(Int, ::AbstractSeries)` should also return an Int.

lbenet:
If I understood correctly, if you look for a specific index of a vector, it seems you should use something like `findfirst` or a related function, rather than `round`.

LuEdRaMo:
It seems that DataInterpolations.jl falls back to FindFirstFunctions.jl which uses `round` to "guess" the best index to start the search...

lbenet:
Closed by #421

PR #421, opened 2026-06-30
Define round(::Type, ::AbstractSeries, ::RoundingMode)

LuEdRaMo:
I've opened this PR to test different definitions for the rounding of an `AbstractSeries` and their implications for NEOs.jl. In https://github.com/JuliaDiff/TaylorSeries.jl/issues/420 we are still debating which is the correction definition.


LuEdRaMo:
I've checked that the definition implemented here, and explained in https://github.com/JuliaDiff/TaylorSeries.jl/issues/420, for the rounding of an `AbstractSeries` is compatible with NEOs and the latest version of SatelliteToolboxTransformations. Therefore, this is ready...

@lbenet Any comments?

lbenet (approved):
Thanks for this addition @LuEdRaMo. LGTM, so I'll merge and edit it.

PR #414, opened 2026-05-26
Migrate fully to `JetSpace` backend (remove legacy TaylorN lookup tables)

PerezHz:
After #410 introduced `JetSpace`, v0.21.10 was released as a compatibility release. Here, we migrate fully to the `JetSpace` backend, removing the legacy lookup tables. The intention is for this PR to become the next minor release (0.22).

Issue #424, opened 2026-09-10
Residual allocations in in-place nested Taylor1s methods


LuEdRaMo:
I have identified the following in-place methods, specialized on nested Taylor1s, that still allocate some memory:
- Because there is no custom method to substract a nested Taylor1 from a scalar, TS falls back to the following method which constructs intermediate objects:
https://github.com/JuliaDiff/TaylorSeries.jl/blob/9a7c39b3bb5b7be48bf7362161f149720fa667c0/src/arithmetic.jl#L90-L97
- The call to `divfactorization` in the code below divides a Taylor1 by another Taylor1 thus allocating an intermediate object:
https://github.com/JuliaDiff/TaylorSeries.jl/blob/9a7c39b3bb5b7be48bf7362161f149720fa667c0/src/arithmetic.jl#L1609-L1614
- The call to `sqrt` in the if block below is not taylorized, hence introducing an intermediate allocation. I think that the `aux` already present in the arguments could be used to replace `sqrt` by its in-place method:
https://github.com/JuliaDiff/TaylorSeries.jl/blob/9a7c39b3bb5b7be48bf7362161f149720fa667c0/src/power.jl#L962-L968

lbenet:
Thanks for reporting; I'll try to get to this as soon as possible, but have some patience...

PR #426, opened 2026-09-13
Fixes for #414 and reorganize files

lbenet:
This PR aims to fix #414, also reorganizing (partially) the file structure.


lbenet:
@LuEdRaMo Can you check that the first part of #414 has been fixed (no promotion with addition/substraction of a nested Taylor1 and a NumberNotSeries)? And also that there is no performance regresion wrt NEOs.jl?

LuEdRaMo:
> @LuEdRaMo Can you check that the first part of #414 has been fixed (no promotion with addition/substraction of a nested Taylor1 and a NumberNotSeries)? And also that there is no performance regresion wrt NEOs.jl?

Using this branch to run the same benchmarks I used to discover the residual allocations reported in https://github.com/JuliaDiff/TaylorSeries.jl/issues/424, the number of allocations is reduced by ~20% while the computation time remains unchanged. Yet, the allocations profiler still flags memory usage within the functions discussed in the issue as I indicate below.

src/arithmetic/division.jl
```diff
@@ -0,0 +1,711 @@
...
+    # c[k] = c[imin] * b[k+ordfact-imin]
+    mul!(c_coeffs[kk], c_coeffs[imin+1], b_coeffs[kk+ordfact-imin])
+    for i = imin+1:k-1
+        # c[k] += c[i] * b[k+ordfact-i]
+        for ord in eachindex(
+                minlength(c_coeffs[kk], c_coeffs[i+1], b_coeffs[kk+ordfact-i]))
+            muladd!(c_coeffs[kk], c_coeffs[i+1], b_coeffs[kk+ordfact-i], ord)
+        end
+    end
+    aux = zero(c_coeffs[kk])
```
LuEdRaMo, line 390: This line creates a `Taylor1{T}` at every call.

src/arithmetic/sqrt.jl
```diff
@@ -0,0 +1,264 @@
...
+        if l0nz < 0
+            return Taylor1(_aaux, order(a_coeffs[dosk01]))
+        elseif isodd(l0nz) # l0nz must be pair
+            throw(DomainError(a_coeffs[dosk01],
+                """First non-vanishing Taylor1 coefficient must correspond
+                to an **even power** in order to expand `sqrt` around 0."""))
+        end
+        # The last l0nz coefficients are dropped.
+        lnull = l0nz >> 1 # integer division by 2
+        aux.coeffs[1] = zero(a_coeffs[dosk01])
```
LuEdRaMo: This line creates a `Taylor1{T}` at every call.

src/arithmetic/add_subst.jl
```diff
@@ -0,0 +1,522 @@
...
+
+                function ($fc)(v::$T, a::$T, b::Number, k::Int)
+                    bb = k==0 ? b : zero(b)
+                    v.coeffs[k+1] = ($f)(a.coeffs[k+1], bb)
+                    return nothing
+                end
+
+                function ($fc)(v::$T, a::Number, b::$T, k::Int)
+                    aa = k==0 ? a : zero(a)
+                    v.coeffs[k+1] = ($f)(aa, b.coeffs[k+1])
```
LuEdRaMo, line 93: For nested Taylor1's, this line falls back to the method in line 51, which creates a `Taylor1{T}`.

src/arithmetic/division.jl
lbenet: I am aware of that, thanks for making it explicit. I'm testing adding a new parameter in the `div!` function, but I have encountered some subtleties...

src/arithmetic/sqrt.jl
lbenet: I think this is solved by [0d676e9](https://github.com/JuliaDiff/TaylorSeries.jl/pull/426/commits/0d676e93712797a04a951c29456864f26bdaa66f)

src/arithmetic/add_subst.jl
lbenet: Solved in [31e0710](https://github.com/JuliaDiff/TaylorSeries.jl/pull/426/commits/31e0710202f4412e91880bb06e216648ee19b6b0)

src/arithmetic/division.jl
lbenet: This one, and another similar around line 390 (using the last uploaded commit), I don't see how to get rid of them...

PR #425, opened 2026-09-11
Define `round(::AbstractSeries, ::RoundingMode)`

LuEdRaMo:
A new SatelliteToolboxTransformations minor version (v1.3) broke the tests in NEOs (see https://github.com/PerezHz/NEOs.jl/actions/runs/34602448371) by using yet another method for rounding `Number`s (`AbstractSeries` for NEOs). So, this PR defines the missing method (see https://github.com/JuliaDiff/TaylorSeries.jl/pull/421 for the original PR on rounding methods for `AbstractSeries`).

PR #427, opened 2026-09-14
Define `ceil` and `floor` for `AbstractSeries`

LuEdRaMo:
In addition to https://github.com/JuliaDiff/TaylorSeries.jl/pull/425, the new minor version of SatelliteToolboxTransformations.jl (v1.3) used in NEOs.jl's tests also requires a definition of `ceil` and `floor` for AbstractSeries; this PR introduces the necessary methods.


lbenet (approved):
Tests pass, and this is essentially the same as some previous PRs, so I'll go ahead and merge. Thanks for the contribution.

